PR #5, opened 2020-10-22
meltAssay

### LINKED ISSUE #3: additional parameters for `agglomerateByRank` or a new supporting function
During testing for visualization, I realize the need for some functionality for making the selection of taxa to plot. This feature for aggregating is important to reduce the high number of variables (taxa in this case) for any visualization to a practical number.   

In `microbiome R package`, functions below help in improving visualization by reducing number of taxa to plot. 
see [aggregate_taxa.R](https://github.com/microbiome/microbiome/blob/master/R/aggregate_taxa.R)
see [aggregate_rare.R](https://github.com/microbiome/microbiome/blob/master/R/aggregate_rare.R)
Now deprecated [aggregate_top_taxa](https://github.com/microbiome/microbiome/blob/master/R/aggregate_top_taxa.R) but I think it can be modified to include an option for users to choose a method for selecting top taxa by a custom function like below: 

```
#' @param x \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
#' @param top Numeric, how many top taxa to return.
#' @param method Specify the method to determine top taxa. Either sum, mean or median.
#'
#' @examples
#' se <- GlobalPatterns
#' top_taxa <- topTaxa(se, method="median", top=5)
topTaxa <- function(x, method=c("sum","mean","median"), top=5){
    taxs <- NULL
    if(method=="sum"){
        taxs <- sort(rowSums(assay(x)),decreasing = TRUE)[1:top]
    } else if (method=="mean"){
        taxs <- sort(rowMeans(assay(x)),decreasing = TRUE)[1:top]
    } else if (method=="median") {
        taxs <- rowMedians(assay(x))
        names(taxs) <- rownames(assay(x))
        taxs <- sort(taxs,decreasing = TRUE)[1:top]
    }
    return(names(taxs))
}
```
Below is a quick and dirty attempt at barplot.

```
library(mia)
library(dplyr)
library(tibble)
library(tidyr)
library(ggplot2)
library(RColorBrewer)
# edit
data(GlobalPatterns, package="MicrobiomeExperiment")
#
se <- GlobalPatterns
se.sgg <- agglomerateByRank(se, rank = "Phylum")
se.sgg <- relAbundanceCounts(se.sgg)
##### melt SE to long data format

# taxonomy data
me_tax <- rowData(se.sgg) %>% 
    data.frame() %>%
    rownames_to_column("TaxaID") # may need to enfore in case of user supplied duplicate

# Metadata
me_sam <- colData(se.sgg) %>% 
    data.frame() %>%
    rownames_to_column("SamId") # may need to enfore in case of user supplied duplicate

dfSE <- assays(se.sgg)$relabundance %>% # this can be optional (relative==TRUE) or something similar
    data.frame() %>%
    rownames_to_column("TaxaID") %>% 
    # may need to enforce in case of user supplied duplicate names
    pivot_longer(!TaxaID, values_to = 'Abundance', names_to = "SamId") %>%  
    left_join(me_tax, by = "TaxaID") %>% 
    left_join(me_sam, by = "SamId") %>% 
    mutate_if(is.factor,as.character)

top_taxa <- topTaxa(se.sgg, method="median", top=10)

dfSE <- dfSE %>% 
    mutate(TaxVar= ifelse(TaxaID %in% names(top_taxa), TaxaID, "Other"))

p <- ggplot(dfSE, aes_string(x="SamId", y="Abundance", fill="TaxVar")) +
    geom_bar(position="stack", stat="identity") +
    #scale_x_discrete(labels=dfm$xlabel, breaks=dfm$Sample)
    scale_x_discrete() + 
    labs(y = "Abundance",x="") + 
    theme_bw() +
    theme(axis.text.x=element_text(angle=90, 
                                   hjust=0.95,
                                   vjust=0.2),
          panel.grid = element_blank(),
          legend.text = element_text(face="italic")) +
    scale_fill_brewer("Phylum",palette = "Paired") # this can be argument (fill.colors= "Paired") or something similar
p + scale_y_continuous(labels = scales::percent) # this can be optional (relative==TRUE) or something similar

```
barplotSE

```
#Option for facetting by user provided variable. Should match `colnames(colData(se.sgg))`
p + facet_wrap(~SampleType, scales = "free")
```

Heatmap option for `SE` found this package called [SEtools](https://github.com/plger/SEtools).

microsud:
Tested on windows 10, R 4.0.3 with `devtools::check()`. No errors and warnings.


R/meltAssay.R
```diff
@@ -0,0 +1,135 @@
...
+#' @aliases meltAssay
+#'
+#' @importFrom SummarizedExperiment rowData rowData<- colData colData<-
+#'
+#' @export
+setMethod("meltAssay",
+  signature = c(x = "MicrobiomeExperiment"),
+  function(x, colData_choice = "all", abund_values = "counts") {
+    # input check
+    if (isFALSE(any(abund_values == c("counts", "relabundance", "both")))) {
```
FelixErnst: You can checkout the function `.check_abund_values` in `MicrobiomeExperiment`. I added that as the universal check for a certain assay to be present. I also don't like to hard code assay names. How the user names their assay and maybe wants to compare two assays of the same type, is up to them.

R/meltAssay.R
```diff
@@ -0,0 +1,135 @@
...
+        data.frame() %>%
+        rownames_to_column("uTaxaID") %>%
+        pivot_longer(!uTaxaID,
+          values_to = "RelAbundance",
+          names_to = "uSamId"
+        )
+
+      dfSE <- me_count_ab %>%
+        left_join(me_rel_ab, by = c("uTaxaID", "uSamId"))
+    }
```
FelixErnst, line 101: This whole section has to be split up into different functions. One function should to one thing.

`meltAssay` should melt only one assay. The used can then merge how many assays as the like. Currently only one or two are allowed, which is too specific.

If you really need the this function, construct a second method which wraps this functionality, eg.  `meltAssays`.

R/meltAssay.R
```diff
@@ -0,0 +1,135 @@
...
+      me_sam <- sub.colData %>%
+        data.frame() %>%
+        rownames_to_column("uSamId")
+
+      dfSE <- dfSE %>%
+        left_join(me_sam, by = "uSamId") %>%
+        mutate_if(is.factor, as.character)
+
+      return(dfSE)
+    }
```
FelixErnst, line 184: I don't see the necessity for distinguishing `all` and `some`. technically is doesn't make a difference, which is shown by the fact that lines 115-117 and 128-130 are duplicates.

I would suggest, selecting the colData to use based on the chosen colums and then `left_join` them independent on how many there are.

In addition, one usage case is missing, which is `none`. I would argue, this should also be the default defined by the value `NULL`

Furthermore, `colData_choice` is mixing snake and camel case. I would strongly advise to revised this. Maybe `add_sample_data` or `add_col_data` would be a better name. Then the whole function could be formulated as a sentence, which is sometimes a good sanity check for naming arguments. The sentence `melt the assay with the abund_values "counts" and add the sample data "x,y and z" ` works. With colData_choice the sentence gets kind of clunky.

FelixErnst (changes requested):
Hi Sudarshan @microsud ,

thanks for the PR. I have a few comments on specific points in the code. See below.

In addition tests are missing. tests might benefit from splitting the separate steps into internal functions. For example I could see this split into `.melt_assay`, `.add_taxonomic_data_to_molten_assay` and `.add_col_data_to_molten_assay`. You can then write a little unit test for each internal function and one for the exported function at the end. This way future debugging gets much easier.

Thanks for considering making the changes.
Felix

microsud:
Attempt at addressing the comments on code and unit tests

FelixErnst:
Did you add tests in `tests/testthat`? I don't see any files added in that directory

microsud:
not sure why it was not added. Hopefully should be there now

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#' long data.frame which can be used for \code{\link[tidyverse]{tidyverse}}-tools.
+#'
+#' @param x a numeric matrix or a \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'          object containing a tree.
+#' @param abund_values Must be one of the values returned by assayNames.
+#' @param add_col_data Choice of colData columns to add. Default="none".
+#' @param ... optional arguments not used.
+#' @return A data.frame tbl_df.
+#' @examples
+#' \dontrun{
```
FelixErnst: Please set examples to be run. If we submit it to bioconductor they have to be run anyway

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#' )
+#' head(molten_se)
+#' }
+#' @importFrom SummarizedExperiment assayNames rowData rowData<-
+#' @importFrom tibble rownames_to_column
+#' @importFrom tidyr pivot_longer
+#' @importFrom dplyr left_join mutate_if %>%
+#' @keywords Utilities
+#' @export
+setGeneric("meltAssay",
```
FelixErnst, line 60: Have a look at some other files regarding the structuring of roxygen blocks. Sometimes it is helpful to structure the blocks with spaces and let the end in `NULL`. This way the man page is set up independent of any function. This is probably just a bit of personal flavour. Maybe you want to have look at other packages, like `scater` or `TreeSummarizedExperiment`.

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+setMethod("meltAssay",
+  signature = c(x = "TreeSummarizedExperiment"),
+  function(x, add_col_data = "none", abund_values = "counts") {
+
+    # input check
+    .check_abund_values(abund_values, x)
+
+    uTaxaID <- uSamId <- Abundance <- NULL
+
+    dfSE <- .melt_assay(x, abund_values)
```
FelixErnst: Alsways creating a new object is inefficient, since it increase RAM consumption. You are not reusing any of the objects, so just overwrite them

In addition if `add_col_data` would be `NULL` the function `add_col_data_to_molten_assay` could be skipped. This is what it could look like:

```
if(!is.null(add_col_data)){
    if(is.logical(add_col_data) && length(add_col_data) == 1L && add_col_data){
        add_col_data <- colnames(colData(x))
    } else {
        # check that all values are valid colnames. Otherwise through error
    }
    dfSE <- .add_col_data_to_molten_assay(dfSE, x, add_col_data)
}
```

The same principle would apply to `add_row_data`/`add_row_data_to_molten_assay`

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#'
+#' @export
+setMethod("meltAssay",
+  signature = c(x = "TreeSummarizedExperiment"),
+  function(x, add_col_data = "none", abund_values = "counts") {
+
+    # input check
+    .check_abund_values(abund_values, x)
+
+    uTaxaID <- uSamId <- Abundance <- NULL
```
FelixErnst: Not used.

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#'        object to long data format.
+#' @param x a numeric matrix or a \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'          object containing a tree.
+#' @param abund_values Must be one of the values returned by assayNames.
+#'
+.melt_assay <- function(x, abund_values) {
+  uTaxaID <- NULL
+
+  # input check
+  .check_abund_values(abund_values, x)
```
FelixErnst: You already checked above. So this is not necessary.

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+
+
+#' @description Melts assay in \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'        object to long data format.
+#' @param x a numeric matrix or a \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'          object containing a tree.
+#' @param abund_values Must be one of the values returned by assayNames.
+#'
+.melt_assay <- function(x, abund_values) {
+  uTaxaID <- NULL
```
FelixErnst: I am not sure, but you are not using this either. If using `!uTaxaID` below gets you an WARNING during R CMD check, try to solve it via `.data`. For details have a look at `?.data`

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+}
+
+
+#' @description Combines molten assay with rowData i.e. taxonomy table.
+#' @param dfSE Molten SE.
+#' @param x a \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'          object.
+#'
+.add_taxonomic_data_to_molten_assay <- function(x, dfSE) {
+  dfSE2 <- uTaxaID <- NULL
```
FelixErnst: See above

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+  if (is.null(rowData(x))) {
+    message("Missing taxonomic information")
+    # message("Missing taxonomic information \n returning only molten assay")
+
+    dfSE2 <- dfSE
```
FelixErnst: The object does not need to be copied

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+    return(dfSE2)
+  } else {
```
FelixErnst: The else branch is not necessary

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+
+    # Get taxonomic data stored in rowData
+    me_tax <- rowData(x) %>%
+      data.frame() %>%
+      # uTaxaID is unique TaxaID
+      rownames_to_column("uTaxaID")
+    # may need to enforce in case of user supplied duplicate
+
+    dfSE2 <- dfSE %>%
```
FelixErnst: duplicated objects not needed

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+
+#' @description Combines molten assay and rowData i.e. taxonomy table with
+#'              ColData i.e. sample information.
+#' @param dfSE2 Molten SE.
+#' @param x a \code{\link[=TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'          object.
+#' @param add_col_data Choice of colData columns to add. Default="none".
+#'
+.add_col_data_to_molten_assay <- function(dfSE2, x, add_col_data) {
+  dfSE3 <- uSamId <- NULL
```
FelixErnst: See above

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+  #if(add_col_data[1] == "none") {
+   # add_col_data <- NULL
+  #}
+
+  # since it can be a vector consisting of more than one colnames(colData(x))
+  # values, choose only 1
+  if (is.null(add_col_data[1]) | is.na(add_col_data[1]) | add_col_data[1] == "none") {
+    message("Returning molten assay without colData")
+
+    dfSE3 <- dfSE2
```
FelixErnst: See above

R/utils.R
```diff
@@ -53,3 +53,18 @@
 .safe_deparse <- function (expr, ...) {
   paste0(deparse(expr, width.cutoff = 500L, ...), collapse = "")
 }
+
+################################################################################
+# checks
+
+#' @importFrom SummarizedExperiment assays
+.check_abund_values <- function(abund_values, x,
```
FelixErnst: Just reimport it like this:

```
.check_abund_values <- MicrobiomeExperiment:::.check_abund_values 
```

Then the function from `MicrobiomeExperiment` becomes a single point of thruth

tests/testthat/test-meltAssay.R
```diff
@@ -0,0 +1,41 @@
...
+test_that("meltAssay", {
+
+    data(GlobalPatterns, package="MicrobiomeExperiment")
+    se <- GlobalPatterns
+    molten_assay <- meltAssay(se,
+                              add_col_data = c("X.SampleID", "Primer"),
+                              abund_values = "counts")
+    expect_equal(class(molten_assay)[1], "tbl_df")
+    expect_equal(class(molten_assay)[2], "tbl")
+    expect_equal(class(molten_assay)[3], "data.frame")
```
FelixErnst: try to use `expect_s3_class`

```
expect_s3_class(tbl, c("tbl_df","tbl","data.frame"))
```

tests/testthat/test-meltAssay.R
```diff
@@ -0,0 +1,41 @@
...
+    expect_equal(is.numeric(molten_assay$Abundance), TRUE)
+
+    only_assay <- .melt_assay(se, abund_values = "counts")
+    expect_equal(colnames(only_assay)[1], "uTaxaID")
+    expect_equal(colnames(only_assay)[2], "uSamId")
+    expect_equal(colnames(only_assay)[3], "Abundance")
+    expect_equal(is.numeric(only_assay$Abundance), TRUE)
+
+    assay_taxa <- .add_taxonomic_data_to_molten_assay(se, only_assay)
+    expect_equal(colnames(assay_taxa)[1], "uTaxaID")
```
FelixErnst: `expect_equal(colnames(assay_taxa),c("uTaxaID","uSamId","Abundance","Kingdom"))` is shorter

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#' }
+#' @importFrom SummarizedExperiment assayNames rowData rowData<-
+#' @importFrom tibble rownames_to_column
+#' @importFrom tidyr pivot_longer
+#' @importFrom dplyr left_join mutate_if %>%
+#' @keywords Utilities
+#' @export
+setGeneric("meltAssay",
+  signature = "x",
+  function(x, add_col_data = "none", abund_values = "counts", ...) {
```
FelixErnst: I think it would be good, if `add_col_data` would accept the following values:

- NULL so no data gets added.
- character values to select certain columns
- TRUE to select all columns.

It could be the case, that a column is named `none` or `all`, which then couldn't be added to the molten data. The same is also true for taxonomic data. I think we need a argument for that as well. Maybe it is also good to use a general term like `add_row_data`. What do you think?

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+#'
+.melt_assay <- function(x, abund_values) {
+  uTaxaID <- NULL
+
+  # input check
+  .check_abund_values(abund_values, x)
+
+  molten_assay <- assays(x)[[abund_values]] %>%
+    data.frame() %>%
+    rownames_to_column("uTaxaID") %>%
```
FelixErnst: Does the `rownames` need to contain taxonomic information? I think a more general term like `uFeatureID` would be more appropriate and would be consistent with `uSamID` used below.

For what does the `u` stand for in `uTaxaID`? unique?

Maybe `SampleID` and `FeatureID` would do as well, since there are now checks for uniqueness in the code. I think the uniqueness is only derived from `rownames` and `colnames`. In addition the values are not unique in the final molten data.

R/meltAssay.R
```diff
@@ -0,0 +1,152 @@
...
+
+
+#' @rdname meltAssay
+#' @aliases meltAssay
+#'
+#' @importFrom SummarizedExperiment rowData rowData<- colData colData<-
+#'
+#' @export
+setMethod("meltAssay",
+  signature = c(x = "TreeSummarizedExperiment"),
```
FelixErnst: It does not need to be a `TreeSummarizedExperiment`. Just a `SummarizedExperiment` would work as well, since `SCE` or `TSE` specific features are not used.

FelixErnst (commented):
Hi Sudarshan @microsud,

The changes are good and certainly in the right direction. Have look through my comments and let me know, what you think. Thanks!

Felix

microsud:
Thanks. All good points and will consider them in one more attempt 🙂

R/meltAssay.R
microsud: I made changes to `add_col_data` options. Regarding `add_row_data`, I am not sure if this option will be of use. This is mostly based on my experience with microbiome data. The `rowData` is one of the most useful information for interpretation and data visualization. `phyloseq` also returns `tax_table` by default. Additionally, the hierarchical nature of information with often *missing* classification at various levels hampers interpretation and users would always keep all information at all taxonomic levels.

FelixErnst: I don't see this as a major problem. Maybe this gets us back to a conceptual discussion. `meltAssay` is for me the main function, the workhorse so to say. However, I don't see that this as a strictly user facing functions.

Think about it this way: What task would a user have towards the object and its data? 
1. Give me the abundance data with all/some/these taxonomic information.
2. give me the abundance data with with certain covariate information.

In addition, `melting` is a really technical term, which isn't even used in the tidyverse anymore (they now use `pivot_longer`) and is certainly not a term associated by default with tasks outlined above.

Therefore, at least in my mind, we will have a function `getAbundanceData` or something, which will have different default and will get the task done and use `meltAssay` internally.

So this is why I am strongly arguing, that we keep our options open.

`meltAssay` must be able to add data, but only, when required. If we want to reuse `meltAssay` for some plotting or other internal functions, I don't want to right the function again, just because `meltAssay` gives me data, I didn't ask for. That will blow the memory consumption at the most unfortunate moment, aka. the moment where you are running low on memory. And then this discussion we are having matters.

FelixErnst (approved):
It is good to go. The messages are a bit unneccessary, since they will make using meltAssay as an internal function harder

PR #13, opened 2020-11-05
data handling update

microsud:
`topTaxa`
`getAbundanceSample`
`getAbundanceFeature`


R/getAbundanceFeature.R
```diff
@@ -0,0 +1,78 @@
+#' Returns abundance values for user specified \dQuote{FeatureID}
+#' in all \dQuote{SampleID}
+#'
+#' Basic extractor function for extracting abundances for a user
```
FelixErnst: These have to be complete sentences

R/getAbundanceSample.R
```diff
@@ -0,0 +1,76 @@
+#' Returns abundance values for all \dQuote{FeatureID} in a \dQuote{SampleID}
+#'
+#' Basic extractor function for extracting \dQuote{FeatureID} abundances
```
FelixErnst: These files should be merged into one file: `getAbundance`. Redundancies are minimised that way. Also make clear that these are just convenience wrappers around `assay(x,name)[row,col]`.

R/getAbundanceFeature.R
```diff
@@ -0,0 +1,78 @@
...
+#' specified \dQuote{FeatureID} from assay in a
+#' \code{\link[=SummarizedExperiment-class]{SummarizedExperiment}}
+#' object.
+#'
+#' @param x A
+#'  \code{\link[=SummarizedExperiment-class]{SummarizedExperiment}} object.
+#'
+#' @param feature_id A \dQuote{FeatureID} for which user wants to extract
```
FelixErnst: This is renamed row name. Make sure to add this in the description. See comments below

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+#' @export
+
+setGeneric("topTaxa", signature = "x",
+           function(x, top, method, abund_values)
+               standardGeneric("topTaxa"))
+
+.check_max_taxa <- function(x, top, abund_values){
+
+    if(top > nrow(assay(x,abund_values))){
+        stop("top value is greater than maximum number of taxa in assay", call. = FALSE)
```
FelixErnst: the assumption that rows equals taxa is true for our purposes, but I think the error message should make clear, what the underlying data is expected to be. I would rephrase the error to `'top' must be <= nrow(x)`

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+#'
+#' @param top Numeric value, how many top taxa to return. Default return top five taxa.
+#'
+#' @param method Specify the method to determine top taxa. Either sum, mean or median.
+#'               Default is 'mean'
+#'
+#' @param abund_values a \code{character} value to select an
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#'
+#' @importFrom DelayedMatrixStats rowSums2 rowMeans2 rowMedians
```
FelixErnst: This should be stated next to the code block, which actually uses it

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+          function(x, top=5, method="mean", abund_values = "counts"){
+
+              # check max taxa
+              .check_max_taxa(x, top, abund_values)
+
+              # check assay
+              .check_abund_values(abund_values, x)
+              #assay(x, abund_values)[,sample_id]
+
+              if(method=="sum"){
```
FelixErnst: How about using `switch`?

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+#'                     top=5,
+#'                     abund_values="counts")
+#'
+NULL
+
+#' @rdname topTaxa
+#'
+#' @export
+
+setGeneric("topTaxa", signature = "x",
```
FelixErnst: The name of this function does not follow convention. It does not contain a verb, which is required for a function not accessing a slot of the main input object. More applicable would be `getTopTaxa`.

Please also note that this is basically the same functionality as `getTopHVGs` from the `scran` package. It would make much more sense, to write a wrapper around `getTopHVGs`, since that is a tried and tested function.

The additional benefit you could provide is a reference to the taxonomic rank. For example you could implement the base wrapper around `getTopHVGs` for `SummarizedExperiment` and implement the function for `MicrobiomeExperiment` as well. In the latter function you could then write something like this:

```
top_taxa <- callNextMethod() # This defers to the normal SummarizedExperiment method
# select top_taxa based on a selected rank
# e.g. translated the returned row names to taxonomic values of a selected rank.
# This way you answer the question of "What are the top Phyla for the most 10 abundant ASV?"
```

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+           function(x, top, method, abund_values)
+               standardGeneric("topTaxa"))
+
+.check_max_taxa <- function(x, top, abund_values){
```
FelixErnst: a check that top is actually an integer value is missing. If `top = "whatever"` the function does not work.

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+}
+
+#' @rdname topTaxa
+#'
+#' @export
+#'
+setMethod("topTaxa", signature = c(x = "SummarizedExperiment"),
+          function(x, top=5, method="mean", abund_values = "counts"){
```
FelixErnst: the value of mean is not sanitised. Setting `mean = "whatever"` will crash this whole function.

Best practice would be to add the generic like this:
```
setGeneric("topTaxa", signature = "x",
            function(x, ...)
                standardGeneric("topTaxa"))
```

and implement the code like this:

```
setMethod("topTaxa", signature = c(x = "SummarizedExperiment"),
           function(x, top = 5L, method=c("mean","sum","median"), abund_values = "counts"){
                method <- match.arg(method)
...
})
```

That way, the generic for `topTaxa` can be reused, if other definition of `topTaxa` need to be implemented. (See comment above wrapping `getTopHVGs`)

However, if you would like to stick to the current implementation more closely you could do this

```
setGeneric("topTaxa", signature = "x",
            function(x, top=5L, method=c("mean","sum","median"), abund_values = "counts")
                standardGeneric("topTaxa"))
```

and implement the code like this:

```
setMethod("topTaxa", signature = c(x = "SummarizedExperiment"),
           function(x, top = 5L, method=c("mean","sum","median"), abund_values = "counts"){
                method <- match.arg(method, c("mean","sum","median"))
...
})
```
Please note that default values have to defined in the generic, if the argument is defined there. In the first example this is not the case and therefore the additional arguments are defined by `setMethod` with their defaults

For more details on `match.arg`, see `?match.arg`

R/topTaxa.R
```diff
@@ -0,0 +1,90 @@
+#' Returns top N \dQuote{FeatureID}
+#'
+#' Basic extractor function for extracting top N \dQuote{FeatureID}
```
FelixErnst: see above about whole sentences.

FelixErnst (changes requested):
Hi Sudarshan @microsud,

thanks for the PR. There are some redundancies, but I think the addition of there wrappers is a nice addition to make transition to `MicrobiomeExperiment` easier.

Have a look at the individual comments and let me know what you think.

We can discuss the details, when you are back.

Felix

R/topTaxa.R
microsud: The `getTopHVGs` functions has a different purpose than `getTopTaxa`. Since the input for variable taxa may have multiple options in microbiome like the output from `DESeq2`, `ALDEx2`, `ANCOM` etc. it may be better to have a separate function that includes `getTopHVGs` for a generally applicable utility. `getTopHVGs` ar similar will be useful for instance when plotting dendrogram-heatmaps, fold-change plots. While `getTopTaxa` is useful for barplots, boxplots, and summarising overall data. 
Do you think the taxonomy option can be more general like `getTaxonomy` that will take input as the output of `getTopTaxa`  or any user-specified list and return classification at a specific or all taxonomic levels? This can come in handy when wanting to analyze/visualize at ASVs level. I currently do use [`format_to_besthit`()](https://microsud.github.io/microbiomeutilities/articles/microbiomeutilities.html#formatting-the-phyloseq-object) when working with ASV/OTU level data. Something we can think about and which in my experience are common questions "Which are the `n` most dominant `ASVs` in each of the phyla/family/...?". Let me know your thoughts.

R/getAbundance.R
```diff
@@ -0,0 +1,158 @@
...
+#'
+#' @param abund_values a \code{character} value to select an
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#'
+#' @return An integer vector of the abundance values for
+#' all \dQuote{FeatureID} in
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#' for specified \dQuote{SampleID}
+#'
+#' @name getAbundanceSample
```
FelixErnst: please rename the man page to `getAbundance` and include both `getAbundanceSample` and `getAbundanceFeature` in the same man page

R/getAbundance.R
```diff
@@ -0,0 +1,158 @@
...
+        stop("Please provide a valid 'sample_id'", call. = FALSE)
+    }
+
+}
+
+
+#' Return abundance values for user specified \dQuote{FeatureID}
+#' in all \dQuote{SampleID}
+#'
+#' This is a basic function for extracting abundances of user specified \dQuote{FeatureID}
```
FelixErnst: This text as well as all the param descriptions are redundant. See comment above

R/getAbundance.R
```diff
@@ -0,0 +1,158 @@
...
+
+
+    if(is.null(feature_id)) {
+
+        stop("'feature_id' must be a single non-empty character value",
+             call. = FALSE)
+
+    } else if (isFALSE(any(feature_id %in% rownames(assay(x, abund_values))))) {
+
+        stop("Please provide a valid 'feature_id'", call. = FALSE)
```
FelixErnst: Please describe what a valid `feature_id` is with the error message. Something like `'feature_id' must be in rownames(x)` or similar

R/getTopTaxa.R
```diff
@@ -0,0 +1,88 @@
...
+                             sum = rowSums2(assay(x, abund_values)),
+                             median = rowMedians(assay(x, abund_values)))
+
+
+              names(taxs) <- rownames(assay(x))
+              taxs <- sort(taxs,decreasing = TRUE)[1:top]
+              return(names(taxs))
+
+          })
+
```
FelixErnst: Please deleted all empty lines, which you don't need. However, each file should end with a single empty line.

R/getTopTaxa.R
```diff
@@ -0,0 +1,88 @@
...
+#'
+#' @export
+setGeneric("getTopTaxa", signature = "x",
+           function(x, top=5L, method=c("mean","sum","median"), abund_values = "counts")
+               standardGeneric("getTopTaxa"))
+
+
+.check_max_taxa <- function(x, top, abund_values){
+
+    if (!round(top) == top) {
```
FelixErnst: Nope. this is not a good test. If top is a `character`, what will happen then?

```
> round("a") == "a"
Error in round("a") : non-numeric argument to mathematical function
```

This is the test we are aiming for:
```
if(!is.numeric(top) || as.integer(top) != top){
    top("'top' must be integer value", call. = FALSE)
}
```

R/getTopTaxa.R
```diff
@@ -0,0 +1,88 @@
...
+#' Sudarshan A. Shetty
+#'
+#' @examples
+#' data(GlobalPatterns, package = "MicrobiomeExperiment")
+#'
+#' top_taxa <- getTopTaxa(GlobalPatterns,
+#'                        method="mean",
+#'                        top=5,
+#'                        abund_values="counts")
+#'
```
FelixErnst: if you add add a print command for top_taxa, this will make the pkgdown much more readable. Like this

```
#' top_taxa <- getTopTaxa(GlobalPatterns,
#'                        method="mean",
#'                        top=5,
#'                        abund_values="counts")
#' top_taxa
```

R/topTaxa.R
FelixErnst: Ok my bad. `getTopHVGs` returns the row names of the most variable rows and uses recalculate statistics in a `DataFrame`, whereas `getTopTaxa` focuses on the overall abundance in assay data.

Underlying this mixup is a different definition of the word `top`. top in most interesting because of most changes or top as in must abundant. Am I right?

Regarding the taxonomic option: Please have a look at the implementation of `agglomerateByRank`. There the `rank` argument directs to the taxonomic level to be used. Translated to `getTopTaxa` this could result in the following:

```
setMethod("getTopTaxa", signature = c(x = "MicrobiomeExperiment"),
    function(x, rank = NULL, ...){
     ans <- callNextMethod(...)
     if(!is.null(rank)){
         if(length(rank) != 1L || !(rank %in% taxonomyRanks(x))){
            stop("If set, 'rank' must be a valid taxonomic rank as returned by taxonomyRanks(x)", call. = FALSE)
         }
         tax_data <- rowData(x)[,taxonomyRanks(x)]
         ans <- unique(tax_data[ans])
     }
     ans
  }
)
```

So if `rank` is not set, it returns a row name of the `SE` object and if it is set a look in the `rowData` starts to return the top taxonomic rank values for the given rank. Would that make sense to you?

As a side note since you brought it up:
Can you explain to me the function of of `format_to_besthit()`? As far as I can tell from the manual page it reformats the `phyloseq` row names, to lowest populated taxonomic level? If that is the case, this function is available as `getTaxonomyLabel` in `mia`

FelixErnst (changes requested):
Good changes. Can you reindent all the code so that it is always steps of 4 spaces?

microsud:
For `getTopTaxa` the taxonomic ranks it is challenging because of the high number of missing information in the hierarchical `rowData`, especially as we go to lower taxonomic ranks. It may not be the ideal output especially if we use `unique()`. I think we can address this in tutorials. 
E.g.
```r
library(mia)
library(dplyr)
data(GlobalPatterns, package = "MicrobiomeExperiment")
me <- GlobalPatterns
top_features <- getTopTaxa(me, method="median", top=10)
tax_data <- rowData(me)[top_features,taxonomyRanks(me)]
tax_data %>% as.data.frame()
```
The `format_to_besthit()` is a utility to change rownames in both `otu_table` and `tax_table` to include taxonomic identity. 
In the example below, I put `getTaxonomyLabels` to use in a different way using `ME` and a `assay`, as it does not take a list of Ids as input to return their best taxonomy. Fetching of best taxonomy from a list of IDs can be a useful utility.
E.g. 
```
library(mia)
library(pheatmap)
data(GlobalPatterns, package = "MicrobiomeExperiment")
me <- GlobalPatterns

## Example heatmap-1
pheatmap(assay(me, "counts")[26:50,1:20])
# Here, we see only IDs. Interpretation is limited with taxonomic information.

# Example heatmap-2
# Some processing 
feature_tab <- assay(me, "counts")
# change IDs to best available taxonomic classification
rownames(feature_tab) <- getTaxonomyLabels(me)
feature_tab[1:6,1:6]
pheatmap(feature_tab[26:50,1:20])
# Here, we see taxonomy and higher interpretation is possible.

# Example heatmap-3
feature_tab2 <- assay(me, "counts")
rownames(feature_tab2) <- paste0("ASV:",rownames(feature_tab2), "-", getTaxonomyLabels(me))
pheatmap(feature_tab2[26:50,1:20])
# Here, we see IDs + taxonomy making this information richer and interpretation can be done specific to IDs and taxonomy.
    
```
pheatmaps


If possible, How can we store this ? or Change rownames in rowData and assays to use the best taxonomic classification instead of some less-informative IDs e.g. "549656","331820" or "ASV001", "ASV002" ?  
Let me know what you think of this.

Best wishes,
Sudarshan

FelixErnst:
I think you are looking for this:

```
# this changes rownames of assays as well as rowData(). Basically the equivalent to format_to_besthit()
rownames(me) <- getTaxonomyLabels(me) 
feature_tab <- assay(me, "counts")
feature_tab[1:6,1:6]
pheatmap(feature_tab[26:50,1:20])
```

microsud:
Right! :smile: `getTaxonomyLabels` works fine. Small edits/tweaks can be shown in tutorials.

FelixErnst:
Hi Sudarshan @microsud,

I cleaned up and simplified the code a bit. Please have a look the changes and let me now, if you agree with the changes. I will merge this once I get the confirmation.

If you have specific questions about certain changes, let me know as well. Happy to discuss this further.

Felix

FelixErnst (approved):

microsud:
These changes are okay with me!

FelixErnst:
Please have a look at the changes. Next time, the indentation must be OK, as well as other formating stuff. Therefore, have a look at changes for reference.

microsud:
yes. I checked it. I had used `formatR::tidy_app` for formatting the code with 4 spaces. Next time will manually format as per your specification. For the `@ ` in Roxygen labels will also refer to this code.

PR #14, opened 2020-11-07
Added MovingPictures data set

antagomir:


antagomir:
I made the PR to microbiome/mia - closing this now.

FelixErnst:
Are you sure about this? I think this will end up in a mess, if this is the git strategy you want to pursue.

.gitignore
```diff
@@ -2,3 +2,5 @@
 .Rhistory
 .RData
 .Ruserdata
+inst/
+*~
```
FelixErnst, line 6: Which files get removed by this?

R/makeMicrobiomeExperimentFromphyloseq.R
```diff
@@ -17,6 +17,12 @@
 #'     makeMicrobiomeExperimentFromphyloseq(enterotype)
 #'     data(esophagus, package="phyloseq")
 #'     makeMicrobiomeExperimentFromphyloseq(esophagus)
+#'
```
FelixErnst, line 20: Examples should be run. What is the intent here? The fix should be applied inside `makeMicrobiomeExperimentFromphyloseq`. Please suggest a general fix for this function

FelixErnst (commented):
Hi Leo @antagomir,  

I have not seen the MovingPictures dataset previously. Can you briefly summarize the added value?

Thanks,
Felix

PR #15, opened 2020-11-08
Prevalence functions.

antagomir:
Prevalence functions added.


.gitignore
```diff
@@ -1,3 +1,5 @@
+inst/
+*~
```
FelixErnst, line 1: What is intent here? I have never seen this type if entry/files. OS specific?

DESCRIPTION
```diff
@@ -1,12 +1,13 @@
 Package: mia
 Type: Package
-Version: 0.0.0.9001
-Date: 2020-11-03
+Version: 0.0.0.9003
```
FelixErnst, line 3: Version gets bumped usually at the merge step. Just fyi

NAMESPACE
```diff
@@ -24,6 +24,8 @@ export(makeMicrobiomeExperimentFromphyloseq)
 export(meltAssay)
 export(mergeCols)
 export(mergeRows)
+export(prevalence)
```
FelixErnst: The functions names do not follow conventions. How about `getPrevalence`/`getPrevalenceMembers`?

R/prevalence.R
```diff
@@ -0,0 +1,94 @@
...
+#' 18(S4):16 20, 2012. 
+#' To cite the R package, see citation('mia') 
+#' @author Leo Lahti
+#' @keywords utilities
+#' @export
+#' @examples
+#' data(enterotype)     
+#' pr <- prevalence(enterotype, detection=0, sort=TRUE, count=TRUE)
+#' pr <- prevalence(enterotype, detection=0, sort=TRUE, count=FALSE)
+prevalence <- function(x, detection=0, sort=FALSE, count=FALSE,
```
FelixErnst: if x is supposed to be a `SummarizedExperiment` or derivative object, the functions, especially these important ones, must be implemented as S4 methods. If you need input on how to proceed best, please let me know.

R/prevalence.R
```diff
@@ -0,0 +1,94 @@
+#' @title Taxa Prevalence 
+#' @description Simple prevalence measure.
+#' @param x A vector, data matrix or \code{\link{MicrobiomeExperiment}} object
+#' @param sort Sort the groups by prevalence
+#' @param count Logical. Indicate prevalence as fraction of samples
```
FelixErnst: Name changes suggestion: `as_relative`

R/prevalence.R
```diff
@@ -0,0 +1,94 @@
...
+        detection <- (-Inf)
+    }
+    
+    if (is.null(x)) {
+        warning("x is NULL - returning NULL")
+        return(NULL)
+    }
+
+    # Add relative abundances if not yet available
+    if (!"relabundance" %in% names(x@assays)) {
```
FelixErnst: Have a look at https://github.com/FelixErnst/mia/blob/645865c1ef0c515d3a75ca752594ede04c28da96/R/utils.R#L68 and its usage here https://github.com/FelixErnst/mia/blob/645865c1ef0c515d3a75ca752594ede04c28da96/R/meltAssay.R#L164

It is good practice to let the user choose the name of assays freely. Otherwise this hinders development, since we cannot anticipate every edge case every user might encounter.

R/prevalence.R
```diff
@@ -0,0 +1,94 @@
+#' @title Taxa Prevalence 
+#' @description Simple prevalence measure.
+#' @param x A vector, data matrix or \code{\link{MicrobiomeExperiment}} object
```
FelixErnst: use the S4 method dispatch system. Include a method for `SummarizedExperiment` and `ANY`. If you need suggestions on how to do this, let me know.

R/prevalence.R
```diff
@@ -0,0 +1,94 @@
...
+        return(NULL)
+    }
+
+    # Add relative abundances if not yet available
+    if (!"relabundance" %in% names(x@assays)) {
+        x <- relAbundanceCounts(x)
+    }
+
+    # Convert to matrix
+    x <- assays(x)$relabundance
```
FelixErnst: The next 17 lines should be handled by the S4 methods.

R/prevalentMembers.R
```diff
@@ -0,0 +1,50 @@
+#' @title Prevalent Taxa
+#' @description Determine members of the prevalent microbiota with a given abundance
```
FelixErnst: see above

R/prevalentMembers.R
```diff
@@ -0,0 +1,50 @@
...
+#' A Salonen et al. The adult intestinal core microbiota is determined by 
+#' analysis depth and health status. Clinical Microbiology and Infection 
+#' 18(S4):16 20, 2012. 
+#' To cite the R package, see citation('mia') 
+#' @author Leo Lahti 
+#' @keywords utilities
+prevalentMembers <- function(x, detection=1/100, prevalence=50/100,
+    include.lowest=FALSE, ...) {
+
+    if ((prevalence < 0) | (prevalence > 1)) {
```
FelixErnst: What about the case, if `prevalence = "meep"`?

R/prevalentMembers.R
```diff
@@ -0,0 +1,50 @@
...
+#' a <- prevalentMembers(enterotype, 1/100, 50/100)
+#' @export
+#' @references 
+#' A Salonen et al. The adult intestinal core microbiota is determined by 
+#' analysis depth and health status. Clinical Microbiology and Infection 
+#' 18(S4):16 20, 2012. 
+#' To cite the R package, see citation('mia') 
+#' @author Leo Lahti 
+#' @keywords utilities
+prevalentMembers <- function(x, detection=1/100, prevalence=50/100,
```
FelixErnst: see above about S4 methods

FelixErnst (changes requested):
Hi Leo @antagomir,

Thanks for the PR. These are important functions.

Before they can be merged, lets discuss improvements and the conversion to S4 methods. See comments above.

Felix

.gitignore
antagomir: The first line prevents adding inst/ folder which is often used only for local purposes. But not always. We can decide to share this, I can remove it from .gitignore or just keep it locally with no reference in .gitignore

I realize that the second line is OS specific. Can be removed.

DESCRIPTION
antagomir: Ok.

NAMESPACE
antagomir: Ok.

.gitignore
FelixErnst: `inst` is valid R package folder, so it can end up in the repo.
I was wondering just about the second line

R/prevalence.R
antagomir: That's possible and I have implemented S4 functions earlier, should be fine. But how do we draw the line between important and less important ones?

R/prevalence.R
antagomir: Can you clarify. Do you refer to an argument name?

NAMESPACE
microsud: `Features` or `Taxa` instead of `members` to keep it in sync with `getTopTaxa` or `getAbundanceFeature`?

R/prevalence.R
antagomir: Just to check - you propose that I will make a PR to `SE` directly? Should be fine.

R/prevalence.R
antagomir: Ok. I just quickly converted the earlier function here to get into a more concrete discussion on the guidelines.

R/prevalentMembers.R
antagomir: Ok

R/prevalentMembers.R
antagomir: Good point. Work in progress.

R/prevalentMembers.R
antagomir: Good

R/prevalence.R
FelixErnst: Thats usually a fine line. If it only makes sense to dispatch on a single type, it is a tied. However, to future proof it a S4 method is usually less trouble long term.

Here you have 2-3 dispatches, which you catch with an if a few lines downstream. So it makes definitely sence here.

R/prevalence.R
FelixErnst: Yes. If I see a variable named `count`, I am thinking numeric, not logical. `as_relative` is immediately recognizable as `logical`

R/prevalence.R
FelixErnst: I am not sure how you mean this. Not to `SummarizedExperiment` package, but to `mia`

I have this in mind:
```
setGeneric("getPrevalence", signature = "x",
           function(x, ...)
               standardGeneric("getPrevalence"))

setMethod("getPrevalence", signature = c(x = "ANY"),
    function(x, ...){
        # do vector or matrix stuff here       
    }
)

setMethod("getPrevalence", signature = c(x = "SummarizedExperiment"),
    function(x, ...){
        # do SE stuff here and subest to matrix or vector
        mat <- some_stuff()
        getPrevalence(mat, ...)
    }
)
```

NAMESPACE
FelixErnst: might make sense. @antagomir ?

.gitignore
antagomir: Ok. The second line should not cause harm so I propose we keep it. And will remove inst/

NAMESPACE
antagomir: makes sense; naming is tricky - 

`Features` is more generic and can be sometimes more accurate than `Taxa`. But it is also longer and less descriptive. 

Moreover, we might like to use the same functions later for other (more refined) metagenomic units or other multiomics that are available through `ME` (or `TSE`/`MAE` or derivatives). Of course there could be aliases for different cases as well if that's easier.

This naming decision may ultimately have implications for a large number of functions.

R/prevalence.R
antagomir: Ah yes I see. True. 

Should we use the term `compositional` or `relative`? This is also a generic decision point.

FelixErnst: I like `as_relative`

NAMESPACE
FelixErnst: prevalence is quite specific for microbiome analysis. Thats why I would go for `Taxa`. You can implement `Feature` as a wrapper and be done with it.

antagomir: We might also like check prevalence of some more refined metagenomic features. But this is indeed possible to sort out later shall needs arise.

R/prevalence.R
antagomir: Actually here relative can refer to either compositional abundances in prevalence calculation; or to prevalence measure itself (relative frequency or absolute count). For clarity, this function now always returns prevalence as a relative frequency. The prevalences can be calculated from absolute or relative data (as_relative option).

antagomir: Another point: this function had `include.lowest` and `as_relative` as arguments. I have now changed to `include_lowest` for consistency. This should be unified across the package/s.

R/getPrevalence.R
```diff
@@ -0,0 +1,94 @@
...
+           function(x, ...)
+               standardGeneric("getPrevalence"))
+
+#' @rdname getPrevalence
+#' @export
+setMethod("getPrevalence", signature = c(x = "ANY"),
+    function(x, detection, include_lowest, sort, as_relative, ...){
+    
+        # do vector or matrix stuff here
+        if (is.null(detection)) {
```
FelixErnst: What about `detection = "some"`?

R/getPrevalence.R
```diff
@@ -0,0 +1,94 @@
...
+            detection <- (-Inf)
+        }
+    
+        if (is.null(x)) {
+            warning("x is NULL - returning NULL")
+            return(NULL)
+        } 
+
+        if (include_lowest) {
```
FelixErnst, line 125: What about `x = "some"`? For rowSums to work it muast be a matrix. `x` needs to be santitized?

R/getPrevalence.R
```diff
@@ -0,0 +1,94 @@
...
+setMethod("getPrevalence", signature = c(x = "ANY"),
+    function(x, detection, include_lowest, sort, as_relative, ...){
+    
+        # do vector or matrix stuff here
+        if (is.null(detection)) {
+            detection <- (-Inf)
+        }
+    
+        if (is.null(x)) {
+            warning("x is NULL - returning NULL")
```
FelixErnst: No point in a warning. This is clearly an error.

R/getPrevalence.R
```diff
@@ -0,0 +1,94 @@
...
+
+    }
+)
+
+#' @rdname getPrevalence
+#' @export
+setMethod("getPrevalence", signature = c(x = "SummarizedExperiment"),
+    function(x, detection=0, include_lowest=FALSE, sort=FALSE, as_relative=FALSE, ...){
+
+        if (as_relative) {
```
FelixErnst: This is a decission for which `as_relative` shouldn`t be used, at least not this way. In this version the user doesn't have a choice, which assay to use. You are forcing a user to have a populated assay named `counts`. This is to restrictive. 

See other function in `mia` as an example. Include an argument `abund_values` for the user to select the assay.

R/getPrevalence.R
```diff
@@ -0,0 +1,94 @@
...
+#' 18(S4):16 20, 2012. 
+#' To cite the R package, see citation('mia') 
+#' @author Leo Lahti
+#' @keywords utilities
+#' @export
+#' @examples
+#' data(GlobalPatterns)     
+#' pr <- getPrevalence(GlobalPatterns, detection=0, sort=TRUE, as_relative=TRUE)
+#' pr <- getPrevalence(GlobalPatterns, detection=0, sort=TRUE, as_relative=FALSE)
+
```
FelixErnst, line 100: Please include a print outline, so that the pkgdown website contains an example output. In this case 
```
#' pr
```

or

```
#' head(pr)
```

If the output is too long

FelixErnst (changes requested):
Hi Leo @antagomir ,

Better but not there yet. You need to sanitize every input, which is user facing. Look at other functions in `mia` for examples and have a look at the function in `utils.R` for help. Also have a look how the code for other functions is formated and follow these examples.

Regarding warnings/errors: If a calculation can continue, it can be a warning, if there is a chance, that the output might make sense. Otherwise aka. if calculation is stopped, it is an error.

Regarding tests: At least one test for every error case, the user might easily trip, should be included. This means that for the main input types a wrong input should be caught.

See earlier review about full sentences for description and details.

Thanks for making the changes. Let me know, if I should give you an example for this code.

Felix

R/getPrevalence.R
antagomir: How about adding  

    ```
    if (is.null(detection)) {
            detection <- (-Inf)
        } else if (!is.numeric(detection)) {
            stop("The detection argument in getPrevalence function should be numeric.")
	}
    
```

R/getPrevalence.R
antagomir: `ME` steps actually take care of this already. Removing this part.

R/getPrevalence.R
antagomir: Removed, see above.

R/getPrevalence.R
antagomir: Changed to just confirming it is numeric and stop() otherwise

R/getPrevalence.R
antagomir: Ok done. This is good. I hope it is sufficient now but if not then kindly specify where further changes necessary.

R/getPrevalence.R
antagomir: done

antagomir:
Above-mentioned changes have been completed but I overlooked something you can point me more specifically to the right direction.

FelixErnst (approved):
Hi @antagomir 

please have a look at the changes. It is mostly cleanup and some optimization for the relative abundance calculation. Let me know, what your take on this is.

antagomir (commented):
Seems fine. All ok.

PR #18, opened 2020-11-27
Agglomerate by rank

antagomir:
Without this PR fix, the following fails. The reason is that the enterotype data set has only rank (Genus) and we are here trying to aggregate to that same rank. This should not lead to failure but return the original object imo.

```
data(enterotype)
x <- enterotype
agglomerateByRank(x, rank = "Genus")

```


antagomir:
The prevailing problems are OS specific.

Windows: 
- problems in installing breakaway from Gitlab

Mac: 
- Problems with external dependencies (imagemagic?); could be potentially fixed by adding some GHA instructions with an actions.yml file

Any comments on how we should proceed to resolve these issues - do not seem to be related to this specific PR but more general unless I'm mistaken.

FelixErnst:
I will have a look at it later today

FelixErnst:
The problems in the GHA on Windows and mac are related to using R-devel. Once R 4.1 is released, we can have a look at it. If it work on Linux it is fine for now.

FelixErnst:
Actually this was to early to merge. If the data in the only available rank is not unique, than this behavior is now wrong

antagomir:
Oh yes. Then we can let it through the calculations. However, when there is a single rank, the earlier version does not go through but fails.

FelixErnst:
Yes, but that was due to the fact that some of the values were `NA` and the calculation or rownames was stuck. The `SCE` object does not need to have unique rownames, but generally it is good way to enforce this.

If you want to have a look, where the function fails, you can due that via `debugonce(agglomerateByRank, signature="SummarizedExperiment")` and then run the function. This way you can go through it step by step.

PR #19, opened 2020-11-29
Agglomerate by rank

antagomir:
Removed the earlier solution as non-unique entries should be handled differently. This led to problems with objects that have a single taxonomic class (example below).

Now the case with a single column in taxonomic table is solved, and the function proceeds to an informative error message.

```
data(enterotype)
x <- agglomerateByRank(enterotype, rank = "Genus")
```

_Error: Only empty taxonomic information detected. Some rows contain only entries selected by 'empty.fields'. Cannot generated labels._

This error arises bc the enterotype data set has one taxonomic unit (row) where all "Genus" are NA. 

> head(rowData(enterotype))
DataFrame with 6 rows and 1 column
                            Genus
                      <character>
-1                             NA
Bacteria                       NA
Prosthecochloris Prosthecochloris
Chloroflexus         Chloroflexus
Dehalococcoides   Dehalococcoides
Thermus                   Thermus

My suggestion is to add an argument that lets the user choose between the following:

 1. Combine such cases into a single class ("Unknown" or "Missing" or name should be explicitly set by the user to avoid mistakes)
 2. Left out from agglomeration
 3. Leading to the error message (only option now)

The default could be the error message (option 3) but augmented by instructions on how to ignore the NA in agglomeration.


FelixErnst:
Hi Leo. 

The code works with the current version of `mia`. The following doesn't error for me.

```
> data(enterotype)
> x <- agglomerateByRank(enterotype, rank = "Genus")
> x
class: TreeSummarizedExperiment 
dim: 552 280 
metadata(0):
assays(1): counts
rownames(552): -1 Prosthecochloris ... Syntrophococcus Mogibacterium
rowData names(1): Genus
colnames(280): AM.AD.1 AM.AD.2 ... TS98_V2 TS99.2_V2
colData names(9): Enterotype Sample_ID ... Age ClinicalStatus
reducedDimNames(0):
altExpNames(0):
rowLinks: NULL
rowTree: NULL
colLinks: NULL
colTree: NULL
```

Option 2. is supported by adding `na.rm = TRUE`.

https://github.com/FelixErnst/mia/blob/0a607893fc60efffca698000882be66991692e26/R/agglomerate.R#L126

```
> data(enterotype)
> x <- agglomerateByRank(enterotype, rank = "Genus", na.rm = TRUE)
> x
class: TreeSummarizedExperiment 
dim: 551 280 
metadata(0):
assays(1): counts
rownames(551): Prosthecochloris Chloroflexus ... Syntrophococcus Mogibacterium
rowData names(1): Genus
colnames(280): AM.AD.1 AM.AD.2 ... TS98_V2 TS99.2_V2
colData names(9): Enterotype Sample_ID ... Age ClinicalStatus
reducedDimNames(0):
altExpNames(0):
rowLinks: NULL
rowTree: NULL
colLinks: NULL
colTree: NULL
```

Option 1. works as well, since the value is `NA`. This means missing in R and special character value can be added by the user like this after agglomeration:

```
rowData(x)$Genus[is.na(rowData(x)$Genus)] <- "missing"
```


PS: can you rebase to the current version of `mia`? You have a lot of commits in this PR, which doen't have anything to to with the issue.
PPS: Most of the changes for getPrevalance were merged yesterday. See #18 which is a PR of the same name

antagomir:
It works currently bc we still have the non-unique agglomeration accepted. I am now rebasing and making another PR.

PR #20, opened 2020-11-29
taxonomy fix

antagomir:
This fix provides now a more informative error in cases with a single taxonomic rank including NAs-


FelixErnst (approved):
Hi Leo,

I am not sure, what you did regarding the rebase. There should only be one commit difference to the current master of `mia` and not the commits 96df471 to 2b4f689.

Regarding the changes you made I see one improvement, since the charlist ist not create twice and the message includes now the na.rm option.

Were the other issues solved?

Felix

FelixErnst:
btw: `rd[,tax_cols,drop=FALSE]` is the same as `DataFrame(rd[,tax_cols])`, so the additional call to DataFrame is not necessary

antagomir:
Yes. something odd was going on with rebase but it seemed to be otherwise ok so I did not spend more time on it now. 

All seems fine now.

I change the DataFrame thing now in a PR.

PR #23, opened 2020-12-04
dominance indices

antagomir:
With @tvborm we have added dominance indices in mia. 

This is conceptually close to alpha diversity indices. 

However, we chose (for now) to implement dominance indices as their own `estimateDominance` function; this makes the distinction between dominance and diversity more clear.

In a sense, dominance indices are form of alpha diversity, and therefore they could be made available also through `estimateDiversity` function (which would then internally call estimateDominance for those indices).

Opening the PR to initiate discussion. The unit tests are on the way.


NAMESPACE
```diff
@@ -30,6 +31,7 @@ export(makeMicrobiomeExperimentFromphyloseq)
 export(meltAssay)
 export(mergeCols)
 export(mergeRows)
+export(prevalentAbundance)
```
FelixErnst: This is missing the verb in the function name

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+
+
+          }
+)
+
+
+
+
+#---------------------------Help functions----------------------------------------------------------------
+checkIndices <- function(index){
```
FelixErnst: please use the `.` prefix for internal functions, and snake_case if you like to keep it consistent

inst/extras/build.sh
```diff
@@ -0,0 +1,8 @@
+#~/bin/R-patched/bin/R CMD BATCH document.R
```
FelixErnst: This is not really necessary, since the package does not need to "know" how to check itself. And the version number needs to be tracked, which is more work than necessary

inst/extras/document.R
```diff
@@ -0,0 +1,31 @@
+# Package release instructions: http://r-pkgs.had.co.nz/release.html
```
FelixErnst, line 1: See above. Not needed, is it?

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
+#' Estimate Dominance
+#'
+#' This function calculates the community dominance index.
+#'
+#' This includes the \sQuote{DBP}, \sQuote{DMN}, \sQuote{Absolute},
```
FelixErnst, line 5: This sentence is not complete. Maybe a bit on the background of these values and some references, if you want to include them, might round it out.

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+#' Dominance indices should be stored in.
+#'
+#' @param BPPARAM A
+#'   \code{\link[BiocParallel:BiocParallelParam-class]{BiocParallelParam}}
+#'   object specifying whether calculation of estimates should be parallelized.
+#'
+#' @details
+#' \code{estimateDominance} calculates the community dominance indices.
+#'
+#' @return Returns ME object, that includes dominance indices in coldata.
```
FelixErnst: Nope. It should return the same class a `x`, so that it doesn't matter whether you use a `SE`, `TSE` or `ME`

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+#'
+#' @seealso
+#' \itemize{
+#'   \item{\code{\link[=estimateDiversity]{estimateDiversity}}}
+#' }
+#'
+#' @name estimateDominance
+#' @export
```
FelixErnst, line 51: two export tags in this block

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+#'
+#' #Calculates all indices
+#' esophagus <- estimateDominance(esophagus)
+#' #Or you can calculate all indices by specify "all"
+#' esophagus <- estimateDominance(esophagus, index="all")
+#' #Shows all indices
+#' colData(esophagus)
+#'
+#' @export
+#'
```
FelixErnst: Just as tip for clarity:

I like to separate the man page docu from the actual function, since the export tag above is only applied to the generic mentioned below. By ending this block like this

```
#' @examples
#' ... some stuff
NULL

#' @rdname estimateDominance
#' @export
setGeneric("estimateDominance",signature = c("x"),
...
```
the association becomes more clear, at least in my mind.

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+#' @param aggregate (Optional, default = TRUE) Aggregate the top members or not.
+#' If aggregate=TRUE, then the sum of relative abundances is returned.
+#' Otherwise the relative abundance is returned for the single taxa with
+#' the indicated rank.
+#'
+#' @param name A name for the column of the colData where the calculated
+#' Dominance indices should be stored in.
+#'
+#' @param BPPARAM A
```
FelixErnst, line 34: Not used in the code

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+           function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE, aggregate=TRUE,
+                    name = index, ...)
+               standardGeneric("estimateDominance"))
+
+
+#' @rdname estimateDominance
+#' @export
+setMethod("estimateDominance", signature = c(x = "MicrobiomeExperiment"),
+          function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE, aggregate=TRUE,
+                   name = index, BPPARAM = SerialParam(), ...){
```
FelixErnst: Regarding `BPPARAM` see above

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+setGeneric("estimateDominance",signature = c("x"),
+           function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE, aggregate=TRUE,
+                    name = index, ...)
+               standardGeneric("estimateDominance"))
+
+
+#' @rdname estimateDominance
+#' @export
+setMethod("estimateDominance", signature = c(x = "MicrobiomeExperiment"),
+          function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE, aggregate=TRUE,
```
FelixErnst: the types of `abund_values` `rank` `as_relative` `aggregate` are not checked. Please apply the "fail early, fail often" maxime

by the way `rank`, if you need to pass it on to the `agglomerateByRank` function needs to be a character. Use `taxonomyRanks(x)[1L]` as the default value

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+
+    # Return all indices
+    #If the user has written in the "index" argument "all"
+    if (length(index) == 1 && index == "all") {
+        index <- accepted
+    }
+
+    #If there is indices, but none of them are accepted, gives warning and returns null
+    if (!is.null(index) && !any(index %in% accepted)) {
+        warning(paste("None of the dominance indices are not recognized:", paste(index, sep = ",")))
```
FelixErnst: this should be an error

R/estimateDominance.R
FelixErnst: This whole function is to much fluff. The user is required to know, that he/she/they are doing, this includes typing correctly. Replace the function with

```
index <- match.arg(index)
```

and be done with it.

R/estimateDominance.R
FelixErnst: If you define `index` likes this

```
index = c("DBP", "DMN", "absolute", "relative", "simpson", "core_abundance", "gini")
```

in the function and the generic definition, you can use `match.arg` very easily and all the options are included in the man pages, which increases readability. See comment below

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+        index <- intersect(index, accepted)
+    }
+
+    return(index)
+}
+
+dominance_help <- function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE,
+                           aggregate=TRUE) {
+    #Stores the absolute abundances to "otu" variable
+    otu <- assays(x)[[abund_values]]
```
FelixErnst: Use `assay(x,abund_values)`

edit: This goes for all instances in the PR

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+
+    #if index does not have any values
+    if (is.null(index)) {
```
FelixErnst: This should never be the case and is defined by the next line of code, which just non sensical

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+        rank <- rank
+    } else if (index == "absolute") {
+        #Rank=1 by default but can be tuned
+        as_relative <- FALSE
+    } else if (index %in% c("relative")) {
```
FelixErnst: index should always be of `length == 1L`. Therefore it should be `index == "relative"` as it is the case in the first if part

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+            warning(paste("The following dominance indices are not recognized:", paste(discarded, sep = ",")))
+        }
+        #Stores indices that are accepted
+        index <- intersect(index, accepted)
+    }
+
+    return(index)
+}
+
+dominance_help <- function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE,
```
FelixErnst: I don't get the function name. For me `.norm_rank` makes more sense.

However, since more than one step of the analysis is done in this function, this should be broken up into seperate functions

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+    # More advanced Simpson dominance (Simpson 1949) However let us not use
+    # this as it is not in [0,1] and it is very highly correlated with the
+    # simpler variant lambda Species richness (number of species)
+    # S <- length(x) sum(p * (p - 1)) / (S * (S - 1))
+
+    lambda
+
+}
+
+gini <- function(x, abund_values="counts") {
```
FelixErnst: Does not match its usage above. `abund_values` not passed on.

`assay(x,abund_values)` should be done externally and a matrix used as only input.

edit: like you did for `simpson_dominance`

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+    })
+    names(do) <- colnames(x)
+
+    do
+
+}
+
+
+
+gini_help <- function(x, w=rep(1, length(x))) {
```
FelixErnst: Rename to `.do_gini` or `.calculate_gini`

R/estimateDominance.R
FelixErnst: Rename to `.get_gini`

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+              #Check if index/indices that user wants to calculate are accepted. Those that are, are stored in "index" variable.
+              index <- checkIndices(index)
+
+              #Initialize table that is used to store indices
+              tab <- NULL
+
+              #If the indices list is not null
+              if (!is.null(index)) {
+                  #loop through index list
+                  for (idx in index) {
```
FelixErnst: Vectorize with `lapply`

edit: then you can also parallelize with `BiocParallel`

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+                   name = index, BPPARAM = SerialParam(), ...){
+
+              #Check if index/indices that user wants to calculate are accepted. Those that are, are stored in "index" variable.
+              index <- checkIndices(index)
+
+              #Initialize table that is used to store indices
+              tab <- NULL
+
+              #If the indices list is not null
+              if (!is.null(index)) {
```
FelixErnst: If `index` is `NULL`, something has gone wrong. The user should not get away with it. "fail early, fail often"

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+}
+
+#Gets ME object and index data frame as parameters
+add_indices_to_coldata <- function(x, indicesDF, names){
+    #Stores sample names
+    sampleNames <- rownames(indicesDF)
+
+    #Checks if the length of names and indices match. If not, returns ME object without new indices.
+    if(length(names) != length(colnames(indicesDF))){
+        warning("Number of names and indices do not match.")
```
FelixErnst: This should be an error

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+    w <- w[o]/sum(w)
+    p <- cumsum(w)
+    nu <- cumsum(w * x)
+    n <- length(nu)
+    nu <- nu/nu[[n]]
+    sum(nu[-1] * p[-n]) - sum(nu[-n] * p[-1])
+}
+
+#Gets ME object and index data frame as parameters
+add_indices_to_coldata <- function(x, indicesDF, names){
```
FelixErnst: This whole function can be replace by doing `colData(x) <- cbind(colData(x), indicesDF)`

Presort `indicesDF` if you need to, but this should not be required

R/prevalentAbundance.R
```diff
@@ -0,0 +1,76 @@
...
+
+              #Saves the relative abundances (or counts, if wanted) to the variable "values"
+              values <- assays(x)[[abund_values]]
+
+              # Core members
+              cm <- getPrevalentTaxa(x, detection=detection, prevalence=prevalence, include_lowest=include_lowest,
+                                     as_relative=as_relative, rank=NULL)
+
+              if (length(cm) == 0) {
+                  warning("With the given abundance and prevalence
```
FelixErnst: This should be an error

R/prevalentAbundance.R
```diff
@@ -0,0 +1,76 @@
...
+#' @export
+#'
+setGeneric("prevalentAbundance", signature = "x",
+           function(x, abund_values = "relabundance", ...)
+               standardGeneric("prevalentAbundance"))
+
+#' @rdname getPrevalence
+#' @export
+setMethod("prevalentAbundance", signature = c(x = "MicrobiomeExperiment"),
+          function(x, abund_values = "relabundance", detection = 0.1/100, prevalence = 50/100,
```
FelixErnst: All the arguments passed onto `getPrevalentTaxa` should be passed on by using the dots `...`

R/prevalentAbundance.R
```diff
@@ -0,0 +1,76 @@
...
+                   include_lowest = FALSE, as_relative=FALSE, ...){
+
+              #Adds relative abundance table to the data
+              x <- relAbundanceCounts(x)
```
FelixErnst: No point in adding it, since you don't return `x`. Use the internal function `.calc_rel_abund` to get the matrix and add a dispatch for `ANY` on `getPrevalentTaxa`. See `getPrevalence` for an example of implementation

R/prevalentAbundance.R
```diff
@@ -0,0 +1,76 @@
...
+
+              if (length(cm) == 0) {
+                  warning("With the given abundance and prevalence
+            thresholds, no taxa were found. Returning NA for prevalentAbundance. Try to
+            change detection and prevalence parameters.");
+                  ret <- NA
+              }
+
+              # Pick the core and calculate abundance
+              if (is.vector(values)) {
```
FelixErnst: `values` is and always will be a `matrix`. This whole if-part doesn't make sense

R/prevalentAbundance.R
```diff
@@ -0,0 +1,76 @@
...
+                  ret <- values # 1 OTU x N samples
+              } else if (length(cm) == 1) {
+                  ret <- values[cm,]
+              } else if (ncol(values) > 1 && length(cm) > 1) {
```
FelixErnst: Since `values` is always a `matrix`, `colSums` will always work. Use `colSums(values[cm, ,drop=FALSE])` If you need to. 

Hint - this code is the solution to this whole code block from line 63-74:
```
colSums(values[cm, ,drop=FALSE])
```

Replace these lines with this and it will always work.

R/estimateDominance.R
```diff
@@ -0,0 +1,366 @@
...
+          function(x, abund_values = "counts", index="all", rank=1, as_relative=TRUE, aggregate=TRUE,
+                   name = index, BPPARAM = SerialParam(), ...){
+
+              #Check if index/indices that user wants to calculate are accepted. Those that are, are stored in "index" variable.
+              index <- checkIndices(index)
+
+              #Initialize table that is used to store indices
+              tab <- NULL
+
+              #If the indices list is not null
```
FelixErnst: I think it is not a list, but a vector

FelixErnst (changes requested):
Hi Tuomas, Hi Leo,

thanks for the PR.

This is a nice additon, but it needs some refactoring. Please see the comments for more details and let me know if some things need to be clarified.

edit: also test for the functions, also the internal functions, are required.

Felix

NAMESPACE
antagomir: Yes. Should be `getPrevalentAbundance`

inst/extras/build.sh
antagomir: Yes this was my local stuff, can be removed.

inst/extras/document.R
antagomir: Same.

NAMESPACE
TuomasBorman: prevalentAbundance changed to getPrevalentAbundance

R/estimateDominance.R
TuomasBorman: Changed the names of internal functions. E..g checkIndices --> .check_indices and gini --> .gini

R/estimateDominance.R
TuomasBorman: You're right --> corrected the comment.

R/prevalentAbundance.R
TuomasBorman: All arguments that are used as arguments of getPrevalentTaxa --> ...

R/estimateDominance.R
TuomasBorman: Changed the text after return to that what is below. I took it from the estimateDiversity, and I am pretty sure (not 100 %) that it does the job.


@return \code{x} with additional \code{\link{colData}} named \code{*name*}

FelixErnst: If you want make it a sentence. Always better for readability.

R/estimateDominance.R
TuomasBorman: Removed BPPARAM argument because it is not used in the code

R/estimateDominance.R
TuomasBorman: Removed BPPARAM argument because it is not used in the code

R/estimateDominance.R
TuomasBorman: .gini_help to .calculate_gini

R/estimateDominance.R
TuomasBorman: .gini to .get_gini

R/estimateDominance.R
antagomir: yes this is good.

R/prevalentAbundance.R
TuomasBorman: All getPrevalentTaxa's arguments --> ...

R/prevalentAbundance.R
TuomasBorman: Deleted the if structure. Now it is only

# Pick the core and calculate abundance
              ret <- colSums(values[cm, ])

R/estimateDominance.R
TuomasBorman: Changed to error

R/estimateDominance.R
TuomasBorman: Changed to error

R/estimateDominance.R
TuomasBorman: I think it is not working exactly same way. At least, something must change if that easier solution is used. 
In that point, indices have their own names i.e. "gini", "simpson"... In the ".add_indices_to_coldata" function, indices are stored in the object with the name that user has specified. It can be "gini", but it can also be "index1".

 colData(x) <- cbind(colData(x), indicesDF) does not take into account that. It just adds indices to the object. If user want that "gini" is "index1" it does not happen. 

I can change the functionality that way that user can not specify the name, so gini is always "gini". Or I can change it that way that the indices name, that user has specified, are stored as a name of indices earlier.

R/prevalentAbundance.R
TuomasBorman: Changed to 

ret <- colSums(values[cm, ,drop=FALSE])

R/estimateDominance.R
TuomasBorman: Also, I tested it both ways, and I found out thay if user calculates same indices multiple times, same indices are stored in the object multiple times, if cbind() is used. 

In the add_indices_to_coldata, one index is saved only one time. Of course, if user wants it to be stored multiple times, user can give them different names. E.g. there can be "gini" and "gini2", but not "gini" and "gini".

FelixErnst: Then try something like this:

```
colData(x)[,names] <- indicesDF
```

If the names exist they will get overwritten, if not they will get attached as long `length(names) == ncol(indicesDF)`

R/estimateDominance.R
TuomasBorman: Changed to that format

R/estimateDominance.R
TuomasBorman: I agree, that it would be beneficial to use that simple solution. Although with this, I think it is not possible to calculate multiple indices at the same time. With the .check_indices, it was. 

Is it ok, that it is possible to calculate indices only one by one?

(Well, e.g. EstimateDiversity's indices are calculated also one by one, because its function are like estimateRichness. estimateShannon...)

R/estimateDominance.R
TuomasBorman: Changed all the assays() in the PR to assay()

R/estimateDominance.R
antagomir: Hmm.. but `estimateDiversity(x)` does calculate all indices by default (see 3978bbf). Isn't this what you are after as well?

FelixErnst: just do a `match.arg(index, several.ok = TRUE)`. Then you get a vector back, which can only contain values as defined in the function definition. If you then do a lapply on those you should be able to get all the values you need. The lapply can also be switched to `BiocParallel::bplapply` so that the values for the different indices are calculated in parallel

FelixErnst: You can have a look at `estimateDiversity` including the latest changes, if you want to implement parallelization. However, parallelization makes only sense, if there are more then 1000 samples and mostly on non-windows systems.

antagomir: i.m.o would make sense to implement it since large collections are increasingly common now

TuomasBorman: Ok, thanks for clarifying. I replaced the check_indices function with this more simple solution

R/estimateDominance.R
TuomasBorman: Changed the functionality that it is similar to .simpson_dominance / matrix is only input

R/estimateDominance.R
TuomasBorman: Vectorized and parallized

R/estimateDominance.R
TuomasBorman: Deleted function and used the command above

R/estimateDominance.R
TuomasBorman: Added testing. taxonomyRanks(x)[1L] to default value of rank

R/estimateDominance.R
TuomasBorman: Added it back, because it is now used in the code

R/estimateDominance.R
TuomasBorman: Changed rank back to numerical values 1 or 2, because it is not a phylogenetic rank. It is used when calculating DBP and DMN. It stores the value 1 or 2 which represents if the most abundant taxon is taken into account (1) or if two most abundant taxa is taken into account (2).

Because rank is used represent phylogenetic rank in many other functions, it might be good idea to consider other name for that to avoid misunderstanding. 

Now, I changed the name of 'rank' to 'ntaxa'.

TuomasBorman: Implemented parallelization.

R/estimateDominance.R
TuomasBorman: Corrected

R/estimateDominance.R
TuomasBorman: I implemented .add_dvrsty_values_to_colData from esimateDiversity, because I thought that there is no need to do it differently.

R/prevalentAbundance.R
TuomasBorman: --> error

R/estimateDominance.R
TuomasBorman: Corrected

R/estimateDominance.R
TuomasBorman: The default value of index in .dominance_help (I haven't yet changed the name) is now "all". But there is no "all" index, so it does not really do anything. I changed the default value of index to NULL, because I think it makes more sense, because then the index is calculated based on other values that are provided by user (or from default).

FelixErnst: change the default to all the indeces and remove the `all` value. Then all the indices will be calculated

FelixErnst: Like this `index = c("DBP", "DMN", "absolute", "relative", "simpson", "core_abundance", "gini")`. add this in the generic or add all the values in `setMethod` and the `match.arg` call

TuomasBorman: Sorry, if I did not understood correctly, but now the .dominance_help is just calculating one index at a time, so it is not possible to add all indices to .dominance_help's default argument.  It is like .run_dvrsty in estimateDiversity.

TuomasBorman: All in all, estimateDominance works like estimateDiversity. All indices are as a default value of "index" in estimateDominance.

Like this 

#' @rdname estimateDominance
#' @export
setMethod("estimateDominance", signature = c(x = "MicrobiomeExperiment"),
          function(x, abund_values = "counts", index = c("DBP", "DMN", "absolute", "relative", "simpson", "core_abundance", "gini"),
                   ntaxa=1, as_relative=TRUE, aggregate=TRUE, name = index, ..., BPPARAM = SerialParam()){

TuomasBorman: And as a default, you can calculate all indices, like with estimtaeDiversity.

#with...
esophagus <- estimateDominance(esophagus)
esophagus <- estimateDiversity(esophagus)
#...you can calculate all dominance and diversity indices which can be accessed with 
colData(esophagus)

TuomasBorman: I implemented major part of estimateDiversity to estimateDominance, so only bigger difference right now is, how .dominance_help/.run_dvrsty work in detail even though they work same way from the point of user. 

E.g. there is now...

#Calculates dominance indices
              dominances <- BiocParallel::bplapply(X=index,
                            FUN = .dominance_help, x=x,
                            abund_values=abund_values, ntaxa=ntaxa,
                            as_relative=as_relative,
                            aggregate=aggregate,
                            BPPARAM = BPPARAM)

              #Add dominance indices to colData
              .add_dominances_values_to_colData(x, dominances, name)


...which is "directly" from estimateDiversity.

TuomasBorman: So, do you think that also .dominance_help should get all indices as default? That would require some kind of loop/apply structure

FelixErnst: Lets take a step back.

1. Is the .domincance_help function still necessary?
2. If yes what is the function of `.domincance_help`?
3. Why does `.domincance_help` need to know the value of `index`?
4. If it needs to know, than it needs to be able to handle indeces with length of `>= 1`.

So I think somewhere in this line of inquiry lies the solution. It might help, if you would push the current state into the branch, so it gets updated here and I can have a look at the current state.

Otherwise keep up the good work! 👍

TuomasBorman: Oh, I did not realize that it is not updating here. I was updating only my own branch... sorry. Now, all changes should be here also

FelixErnst:
Hi @tvborm 

I had a look at the changes. The structure is better, but since we want to be able to calculate all index from the same call it requires some additional changes.

Therefore, I changed some things around, which you should have a look at and comment, whether you think, it makes sense.

In addition, I noticed the follwing things:
-  the argument `as_relative` is always defined by a value of index (dropped for gini, simpson and prevalence and set by the other ones) Therefore, I removed it.
- the dispatch to `MicrobiomeExperiment` is unnecessary, since we dont use any sequence of tree data in the functions. Therefore, the `SummarizedExperiment` dispatch is enough for now
- the `getPrevalentTaxa` needed a tweak to work with a simple matrix. See changes there

If you are done, please add test for all the functions. I added already a file in tests.

Felix

R/estimateDominance.R
antagomir: I just think that we should follow the identical conventions with estimateDominance and estimateDiversity wherever possible.

FelixErnst: I agree. I am quite sorry, that I had to change things around myself, but the dominance values are getting computed in several different ways, so that it was very complex to explain. I hope you both don't mind the changes

TuomasBorman: On my behalf, it is more than fine that you did those changes. Actually, I was thinking if I should implement that "last part" of estimateDiversity also because of the consistency, and there was no reason to do it differently in estimateDominance. So yeah, everything is fine

tests/testthat/test-5dominance.R
```diff
@@ -0,0 +1,29 @@
+context("estimateDominance")
+
+test_that("estimateDominance", {
```
FelixErnst, line 3: Unit test should be structured slighlty differently. The main focus is to make sure it work as intended, not that it works.

Two way exists:
- Either you compute it by two different ways
- you compare the results to fixed value, you know to be true. (Use pen and paper if you have too)

The examples below show that it works but not necessarily as intended. In the follwoing code chunk is an example, which you can use as blue print for adding all the tests.
```
# .simpson_dominance
x <- 1
expect_equal(mia:::.simpson_dominance(x),1)
x <- 1:1000
expect_equal(mia:::.simpson_dominance(x), 0.001332667)
```

So in this style you need to make sure that the follwing functions work correctly:
- `.gini_dominance`
- `.get_core_dominance`
- `.get_dominance` for all four cases

then you check if all indeces get calculated for the call `estimateDominance(esophagus)` by checking the colnames of `colData()` after making sure, that the return value of `estimateDominance` is actually the same type as the input.

edit: also don't use the slots directly. If, for whatever reason the slots change, our code will still work since it uses `colData()<-` to write the data. The tests will not since the slots get accessed directly. You can use `colData(estimateDominance(esophagus))` if you want

R/estimateDominance.R
FelixErnst: Cool 👍 

btw: the `.add_dvrsty_values_to_colData` function is a bit more complex, because it needs to handle nested `DataFrame`s. Since the dominance calculation always return a vector, that is not necessary

tests/testthat/test-5dominance.R
TuomasBorman: Ok, thanks for help. I changed tests little bit. All, but 3 are working. It seems that .simpson_dominance should return value 1-a,  not a.

FelixErnst: Then you have uncovered and error, which you can fix in the code and the tests.

FelixErnst: And I really like that you test against vegan functions. You can fix the error in the code of `.simpson_dominance`, if you are sure about it. @antagomir

tests/testthat/test-5dominance.R
```diff
@@ -2,28 +2,72 @@ context("estimateDominance")
...
-    #Calculates DMN in two different ways, should get the same result.
-    expect_true(all(estimateDominance(esophagus, index = NULL, relative = TRUE, rank = 2)@colData$DMN ==
-                        estimateDominance(esophagus, index = "DMN")@colData$DMN))
+    #estimateDominance
+    #Calculates all indices.
+    tse_idx <- mia:::estimateDominance(tse)
+    #Checks that the type of output is the same as the type of input.
+    expect_true(typeof(tse_idx) == typeof(tse))
+    #Checks that every index is calculated by checking the column names from colData.
+    expect_true(all(
```
FelixErnst: the function `expect_named` is available as well. `expect_named(colData(tse_idx),c("relative", ....... ))`. Be aware, that correct order is tested as well.

tests/testthat/test-5dominance.R
antagomir: Simpson is confusing bc there are [different indices](https://en.wikipedia.org/wiki/Diversity_index#Simpson_index).

The _Simpson index_ is defined as $\lambda = \sum_i p^2_i$. 

_Simpson dominance index_ is defined as $\lambda$. It is the same as Simpson index (one source [here](https://www.sciencedirect.com/science/article/pii/S2405844019362668)

`Inverse Simpson` diversity index is defined as $\frac{1}{\lambda}$.

Then we also have `Gini-Simpson` index $1 - \lambda$.

Would be nice to have all of these in the package but the definitions must be clear in order to avoid misuse. Just check that the function corresponds to these official definitions (inverse Simpson and Gini-Simpson should go to `estimateDiversity` whereas Simpson should go to `estimateDominance`).

Another point to check that all dominance indices (and all diversity indices) have the same sign. In some cases the interpretation of alpha diversity/dominance indices can be inverse compared to the other indices. Where this is the case, we may like to consider how to treat. It is causing some risks for interpretation if the indices with the same purpose have opposite interpretations for increasing index values (higher values meaning either decreasing or increasing diversity or dominance). Could you check on the same go if there are such cases so that we can see what to do. That discussion can go in a new issue.

tests/testthat/test-5dominance.R
TuomasBorman: I changed it to 

expect_named(colData(tse_idx), c("DBP", "DMN", "absolute", "relative", "simpson", "core_abundance", "gini"))

estimateDominance calculates the indices in the same order as they are in the "Index" vector. I assumed that it is the right order.

FelixErnst:
I think the last open topic is the correct value for the simpson domince. Can you comment/check on this?

The PR is ready to be merged.

FelixErnst:
One additional thing, the package fails the test. You have to include the `reldist` package in the suggest and add `skip_if_not_installed` call at the beginning of the dominance test, which rely on the comparison to the `reldist` package

tests/testthat/test-5dominance.R
TuomasBorman: Modified simpson tests. I calculated simpson index with pen and paper and tested .simpson_dominance with that

FelixErnst:
I bumped the version, in order to force an dependency update in GHA.

FelixErnst (approved):

PR #26, opened 2020-12-09
added DMN wrappers

FelixErnst:


R/calculateDMM.R
```diff
@@ -257,9 +257,11 @@ setMethod("plotDMNFit", signature = c(x = "SummarizedExperiment"),
...
         fit <- vapply(dmn, fit_FUN, numeric(1))
         ggplot(data.frame(k = k, fit = fit), aes_string(x = k, y = fit)) +
-          geom_point() +
-          geom_line() +
-          theme_bw()
+            geom_point() +
+            geom_line() +
+            theme_bw() +
+            xlab("Number of Dirichlet Components") +
+            ylab(paste0("Model Fit (",type,")"))
```
antagomir: I suggest using the `labs` function i.e. `labs(x = "Number of Dirichlet Components", y = paste0("Model Fit (",type,")"))` instead of `xlab` and `ylab`. "The labs function is preferable to xlab/ylab/ggtitle, since it generalizes to other aesthetics and labels (e.g. subtitle and caption), which don't have their own functions."

R/calculateDMM.R
```diff
@@ -261,7 +261,7 @@ setMethod("plotDMNFit", signature = c(x = "SummarizedExperiment"),
             geom_line() +
             theme_bw() +
             xlab("Number of Dirichlet Components") +
-            ylab(paste0("Model Fit (",type,")"))
+            xlab(paste0("Model Fit (",type,")"))
```
antagomir: Same comments here, let's use the `labs` function.

FelixErnst: done

antagomir (commented):
ok

FelixErnst:
thx for having a look. I totally missed the lab issue

PR #35, opened 2021-01-03
Get dominant taxa

TuomasBorman:
Hi!

Here is the getDominantTaxa.


FelixErnst:
Hi Tuomas,

I will have a look at it, but for the next addition please create branch of the then current master to avoid all those commits, which were already merged. Thanks.

Felix

TuomasBorman:
Ok, that's clear, sorry. This was my first PR that I made by myself, so I wasn't 100% sure about these things.

FelixErnst:
No problem. Thats part of the learning curve 🔢

R/estimateDominance.R
```diff
@@ -38,6 +38,57 @@
...
+#' large portion of total abundance of microbes, the inequality is large and
+#' Gini index closer to 1. If all species has equally large abundances, the equality
+#' is perfect and Gini index equals 0. (Sitthiyot & Holasut 2020.)}
+#' }
+#'
+#' @references
+#' Maiti SK (2012) Ecorestoration of the coalmine degraded lands.
+#'
+#' Sitthiyot T & Holasut K (2020) A simple method for measuring inequality.
+#' Palgrave Commun 6(112) doi: 10.1057/s41599-020-0484-6.
```
antagomir: We should use here a more original reference of the Gini index.

R/estimateDominance.R
```diff
@@ -38,6 +38,57 @@
...
+#'
+#' Thukral AK, Bhardwaj R, Kumar V & Sharma A (2019)
+#' New indices regarding the dominance and diversity of communities,
+#' derived from sample variance and standard deviation.
+#' Heliyon 5(10). doi: 10.1016/j.heliyon.2019.e02606.
+#'
```
antagomir: Also here, if any more original/classical references can be identified that would be preferable.

R/estimateDominance.R
```diff
@@ -38,6 +38,57 @@
...
+#' In economics, it is used to measure, for example, how unevenly wealth
+#' is distributed among population. Here, Gini index is used similarly, but wealth
+#' is replaced with abundance. If there is small group of species that represent
+#' large portion of total abundance of microbes, the inequality is large and
+#' Gini index closer to 1. If all species has equally large abundances, the equality
+#' is perfect and Gini index equals 0. (Sitthiyot & Holasut 2020.)}
+#' }
+#'
+#' @references
+#' Maiti SK (2012) Ecorestoration of the coalmine degraded lands.
```
antagomir: It this the original reference of this index?

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
+#' Get Dominant Taxa
+#'
+#' This function return the most dominant taxa.
```
antagomir: return -> returns

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
...
+#' @name getDominantTaxa
+#' @export
+#'
+#' @author Leo Lahti and Tuomas Borman. Contact: \url{microbiome.github.io}
+#'
+#' @examples
+#' data(GlobalPatterns)
+#' x <- GlobalPatterns
+#'
+#' #Finds the dominant taxa.
```
antagomir: I would prefer commenting where you have a space after # i.e. "# Finds the dominant taxa". Can this be changed here and elsewhere?

inst/extras/build.sh
```diff
@@ -0,0 +1,7 @@
+~/bin/R-4.0.0/bin/R CMD build ../../ --resave-data #--no-examples  --no-build-vignettes
+~/bin/R-4.0.0/bin/R CMD check mia_0.0.0.9010.tar.gz #--no-build-vignettes --no-examples
+~/bin/R-4.0.0/bin/R CMD BiocCheck mia_0.0.0.9010.tar.gz
+~/bin/R-4.0.0/bin/R CMD INSTALL mia_0.0.0.9010.tar.gz
+
+
+
```
antagomir: This build script we do not necessarily need to have in github; I am not sure if it matters for others.

inst/extras/document.R
```diff
@@ -0,0 +1,31 @@
...
+# revdep_check("../../")
+# add_rstudio_project("../../")
+# use_build_ignore("../NEWS.md", pkg = "../../") # NEWS.md not supported by CRAN
+# use_package("dplyr") # add package to imports
+# load_all(".") # Reload the package
+# test() # Run tests
+# run_examples()
+
+# Vignettes:
+#
```
antagomir: This also not necessarily needed in github

antagomir (commented):
Seems good to me, with the few comments.

R/estimateDominance.R
```diff
@@ -38,6 +38,57 @@
...
+#' ($\frac{N_{core}}{N_{tot}}$, where $N_{core}$ is the sum of absolute abundance of the core species and
+#' $N_{tot}$ is the sum of absolute abundances of all species.) Index gives values in interval [0,1],
+#' where bigger value represent greater dominance.}
+#' \item{'gini' }{Gini index is probably best-known from socio-economic contexts.
+#' In economics, it is used to measure, for example, how unevenly wealth
+#' is distributed among population. Here, Gini index is used similarly, but wealth
+#' is replaced with abundance. If there is small group of species that represent
+#' large portion of total abundance of microbes, the inequality is large and
+#' Gini index closer to 1. If all species has equally large abundances, the equality
+#' is perfect and Gini index equals 0. (Sitthiyot & Holasut 2020.)}
```
FelixErnst: To increase readability of the roxygen block you can ident each item and again the text within it. This will not change the output

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
+#' Get Dominant Taxa
+#'
+#' This function return the most dominant taxa.
+#'
+#' @param x
+#' A \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object
+#'
+#' @param rank A single character defining a taxonomic rank. Must be a value of
+#'   \code{taxonomicRanks()} function.
```
FelixErnst: ... of the output of \code{taxonomicRanks()}

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
...
+#'
+#' @param group With group, it is possible group observations in overview.
```
FelixErnst: the type of input needs to described.

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
...
+              if(!.is_non_empty_string(name)){
+                  stop("'name' must be a non-empty single character value.",
+                       call. = FALSE)
+              }
+
+              #If "rank" is not NULL, species are aggregated according to the taxonomic rank
+              #that is specified by user.
+              if (!is.null(rank)) {
+                  x <- agglomerateByRank(x, rank = rank)
+              }
```
FelixErnst: you change the input, but return it. I think this not the function of this function. The user can choose to do it themself by using `getDominantTaxa(agglomerateByRank(x, rank = rank), rank = rank)`

An alternative would be:

```
tmp <- agglomerateByRank(x, rank = rank)
```

Better would be to restructure `agglomerateByRank` and add a method for the signature `ANY` assuming `x` is in this case a `matrix`-like object. Then you could do something like this

```
mat <- agglomerateByRank(assay(x, abund_values), rank_data = rank_data)
```

However, I think this is to complicated for now. Instead reuse some code and try this 

```
col <- which( taxonomyRanks(x) %in% rank )
tax_factors <- .get_tax_groups(x, col = col, onRankOnly = FALSE)
mat <- mergeRows(assay(x, abund_values), f = tax_factors)
```

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
...
+
+              #assays()$counts returns counts of the taxa in the samples
+              #apply() function finds the indices of taxa's that has the highest
+              #amount of counts.
+              #names() returns the names of taxa that are the most abundant.
+              taxas <- names(x)[apply(assays(x)$counts, 2, which.max)]
```
FelixErnst: hard coded name of the assay. Please make it the users choice.

R/getDominantTaxa.R
```diff
@@ -0,0 +1,152 @@
...
+
+              #Adds taxas to colData
+              x <- .add_dominant_taxas_to_colData(x, taxas, name)
+
+              #Gets an overview
+              overview <- .get_overview(x, group, name)
+
+              #Prints the overview
+              print(overview)
```
FelixErnst: Never use `print`, if the user doesn't specifically asks for it, like in `show`.

Use `message` instead, since this can be turned off by using `suppressMessages` at the users conveniance

FelixErnst (changes requested):
HI Tuomas,

Nice addition! It needs some small adjustment.

I think you are trying to wrap two things into one. `getDominantTaxa` should return the tibble you print as the overview and an additonal function like `addDominantTaxa` or just `dominantTaxa` should add the data to the `colData`

Let me know what you think.

antagomir:
I agree with all those suggestions by @FelixErnst

R/getDominantTaxa.R
TuomasBorman: --> returns

R/getDominantTaxa.R
TuomasBorman: Added space after # in estimateDominance and getDominantTaxa

R/estimateDominance.R
TuomasBorman: Increased readability

R/getDominantTaxa.R
TuomasBorman: Changed

R/getDominantTaxa.R
TuomasBorman: Important information... added description for group and name.

R/getDominantTaxa.R
TuomasBorman: counts --> "abund_values" parameter

R/getDominantTaxa.R
TuomasBorman: Changed to 

                  col <- which( taxonomyRanks(x) %in% rank )

                  # Function from taxonomy.R. Divides taxas to groups where they belong.
                  tax_factors <- .get_tax_groups(x, col = col, onRankOnly = FALSE)

                  # Merges abundances within the groups
                  mat <- mergeRows(x, f = tax_factors)

R/getDominantTaxa.R
TuomasBorman: Changed to message()

TuomasBorman:
Yep, it might be wise to divide them into 2. Now, I have divided the function into 2 functions. dominantTaxa() returns dominant taxa in colData, getDominantTaxa() returns overview in tibble format.

R/estimateDominance.R
TuomasBorman: Changed the reference to 

Gini C (1921) Measurement of Inequality of Incomes. The Economic Journal 31(121): 124-126. doi: 10.2307/2223319

R/estimateDominance.R
TuomasBorman: Changed to 

Simpson EH (1949) Measurement of Diversity. Nature 163(688). doi: 10.1038/163688a0

R/estimateDominance.R
TuomasBorman: Changed to 

Berger WH & Parker FL (1970) Diversity of Planktonic Foraminifera in Deep-Sea Sediments. Science 168(3937): 1345-1347. doi: 10.1126/science.168.3937.1345

inst/extras/build.sh
TuomasBorman: Deleted

inst/extras/document.R
TuomasBorman: Deleted

R/dominantTaxa.R
```diff
@@ -0,0 +1,210 @@
...
+
+                  # Selects the level
+                  col <- which( taxonomyRanks(x) %in% rank )
+
+                  # Function from taxonomy.R. Divides taxas to groups where they belong.
+                  tax_factors <- .get_tax_groups(x, col = col, onRankOnly = FALSE)
+
+                  # Merges abundances within the groups
+                  mat <- mergeRows(x, f = tax_factors)
+              }
```
FelixErnst: `mat` is unset if `rank` is NULL.

I think the following would probably solve it:

```
if (!is.null(rank)) {
    # Selects the level
    col <- which( taxonomyRanks(x) %in% rank )

    # Function from taxonomy.R. Divides taxas to groups where they belong.
    tax_factors <- .get_tax_groups(x, col = col, onRankOnly = FALSE)

    # Merges abundances within the groups
    tmp <- mergeRows(x, f = tax_factors)
    mat <- assay(tmp, abund_values)
} else {
    mat <- assay(x, abund_values)
}

```

R/dominantTaxa.R
```diff
@@ -0,0 +1,210 @@
+#' Dominant Taxa
+#'
+#' This function returns the most dominant taxa.
```
FelixErnst: A more verbose description might be beneficial. Esspecially since you have to different return types

R/dominantTaxa.R
```diff
@@ -0,0 +1,210 @@
...
+#' be a one of the column names of colData.
+#'
+#' @param name A name for the column of the colData where the dominant taxa
+#' will be stored in. Must be a string.
+#'
+#' @details
+#' \code{dominantTaxa} extracts the most abundant taxa
+#' in a \code{\link[=SummarizedExperiment-class]{SummarizedExperiment}} object.
+#'
+#' @return \code{x} with additional \code{\link{colData}} named
```
FelixErnst: Two different return types now

FelixErnst (changes requested):
Nice improvement. A few things still to go. (Don't forget the tests 😉 )

R/dominantTaxa.R
TuomasBorman: You're correct, mat was missing when rank was NULL. dominantTaxa() also returned object containing modified abundance table. I corrected that. Now the only thing that is modified is the colData where dominant taxa is stored.

R/dominantTaxa.R
TuomasBorman: Changed it to 

#' These functions return information about the most dominant taxa in a
#' \code{\link{SummarizedExperiment-class}} object.

R/dominantTaxa.R
TuomasBorman: Changed it to 

#' @return \code{dominantTaxa} returns \code{x} with additional \code{\link{colData}} named
#'   \code{*name*}. \code{getDominantTaxa} returns a tibble containing dominant taxa in
#'   a column named \code{*name*} and its abundance in the data set.

R/dominantTaxa.R
FelixErnst: "Additionally, the information can be directly stored in the \code{colData}"?

Maybe a details section as well, in which you describe the underlying calls regarding data agglomeration

R/dominantTaxa.R
```diff
@@ -124,6 +138,19 @@ setMethod("dominantTaxa", signature = c(x = "SummarizedExperiment"),
...
+                  # Changes the name of species to the name of corresponding rank
+                  # If name contains upper rank, it is cut off
+                  # Tax_factors[rownames(mat)] finds corresponding name
+                  rownames(mat) <- tax_factors[rownames(mat)]
+
+                  # (strsplit(x, "_"))[[1]] splits string from underscores and
+                  # and gives a list of individual ranks
+                  # tail(x, n=1) gives the last rank from list
+                  # --> upper ranks are removed from the name
+                  rownames(mat) <- lapply(rownames(mat), function(x) tail(strsplit(x, "_")[[1]], n=1))
```
FelixErnst: You are assigning a list to vector. This works probably, but better style would be to use `vapply` like this:

```
vapply(strsplit(rownames(mat), "_"),tail,character(1),n = 1)
```

tests/testthat/test-5dominantTaxa.R
```diff
@@ -0,0 +1,57 @@
...
+
+    # Phyloseq object
+    data("GlobalPatterns", package = "phyloseq")
+    phy <- GlobalPatterns
+
+    # Counts table should not be changed
+    expect_equal(assays(mia::dominantTaxa(tse))$counts, assays(tse)$counts)
+
+    # Should get same result
+    expect_equal(colData(mia::dominantTaxa(tse, name="dominant"))$dominant, microbiome::dominant(phy))
```
FelixErnst: You shouldn't use the microbiome package in the test, because then you need to put in the Depends section of the DESCRIPTION, which then pulls in `phyloseq`. It does not make sense. I think an approach with fixed values is OK or compute it fixed with 
`rownames(mat)[apply(mat, 2, which.max)]`

tests/testthat/test-5dominantTaxa.R
```diff
@@ -0,0 +1,57 @@
...
+test_that("getDominantTaxa", {
+
+    # TSE object
+    tse <- x <- microbiomeDataSets::dietswap()
+
+    # Phyloseq object
+    data("dietswap", package = "microbiome")
+    phy <- dietswap
+
+    expect_equal(mia::getDominantTaxa(tse), microbiomeutilities::dominant_taxa(phy)$dominant_overview)
```
FelixErnst: same here. If you need the data use the `microbiomeDataSet` package

FelixErnst (changes requested):
Some fixes still needed.

Please have a look at the build reports and fix the issues:
https://github.com/FelixErnst/mia/pull/35/checks?check_run_id=1651817329

R/dominantTaxa.R
TuomasBorman: Changed

tests/testthat/test-5dominantTaxa.R
TuomasBorman: Used fixed values that I got with microbiome and microbiomeutilities packages

tests/testthat/test-5dominantTaxa.R
TuomasBorman: Used fixed values that I got with microbiome and microbiomeutilities packages

FelixErnst (commented):
Hi Tuomas,

I think thats it, almost 😉 

Can you add the microbiomeDataSets package under the dependes section in DESCRIPTION?

Since it is only on GitHub, you also have to add the remotes section like this:
```
Remotes:
    microbiome/microbiomeDataSets
```

Could you do this and make sure, that the CI pipelines complete without error? Thanks!

FelixErnst:
Fixed one of the build issues on master. If you can merge master into your branch. Thanks and sorry for the inconvenience

TuomasBorman:
Ok, thanks! I merged the master

yiluheihei:
Thanks for your great package :+1:.  To suppress the bioc check warning and error:

* Packages **microbiome** and **microbiomeDataSets** should be added under the `Suggests` filed in `DESCRIPTION`, since these two packages  fonly used in examples or tests. 
* in the `test-5dominantTaxa.R`, `skip_if_not_installed("microbiomeDataSets")` and `skip_if_not_installed("micribome")` should be added before the code  to attach/load the package. Because suggested packages are not installed defaultly.
* the mathematical formulae in the man page of `estimateDominance()` should be fixed.
  > Mathematical formulae should be set beautifully for printed documentation yet we still want something useful for text and HTML online help. To this end, the two commands \eqn{latex}{ascii} and \deqn{latex}{ascii} are used. Whereas \eqn is used for “inline” formulae (corresponding to TeX’s $…$), \deqn gives “displayed equations” (as in LaTeX’s displaymath environment, or TeX’s $$…$$). Both arguments are treated as ‘verbatim’ text.

And I have [created a pull request to `microbiome/mia`](https://github.com/microbiome/mia/pull/5). 

By the way, I'm very interested in this project.  if I want to create a pull request, which repo should I contribute to, [microbiome/mia](https://github.com/microbiome/mia) or [this repo](https://github.com/FelixErnst/mia).

Thanks.

FelixErnst:
Thanks for the interest in the package. This repo is upstream of microbiome/mia.

The suggestions are certainly well intended, but I think are not 100% correct at this moment. The dependency on `microbiomeDataSets` can be put into the `Depends` field and the skip statements can be omited. The datasets are also used in the examples, which requires an entry in the `Depends`, if I am not mistaken.

The `microbiome` dependency should be avoided, since it pulls in the `phyloseq` stack.

I am sure Tuomas will get the PR to a state, which builds without error and warning. This is a learning curve and this PR clearly shows the progress from the other ones he authored. So I was/am happy to wait for Tuomas to get there on his own.

yiluheihei:
Thanks for your response :+1:.  I have closed the pull request.  

Moreover, maybe put **microbiomeDataSets** under `Suggests` field is more appropriate. As shown in the [write R extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Suggested-packages) and hadley's book [r-pkgs](https://r-pkgs.org/description.html#dependencies).

> The ‘Suggests’ field uses the same syntax as ‘Depends’ and lists packages that are not necessarily needed. This includes packages used only in examples, tests or vignettes.

> Suggests: your package can use these packages, but doesn’t require them. You might use suggested packages for example datasets, to run tests, build vignettes, or maybe there’s only one function that needs the package.

> However, using require for conditioning in package code is not good practice as it alters the search path for the rest of the session and relies on functions in that package not being masked by other require or library calls. It is better practice to use code like

    if (requireNamespace("rgl", quietly = TRUE)) {
       rgl::plot3d(...)
    } else {
       ## do something else not involving rgl.
    }

antagomir:
This repo is upstream of microbiome/mia but for the other packages of the ecosystem the microbiome/* is the upstream. We should probably write concise summary & instructions on how to contribute to the different parts of the ecosystem.

FelixErnst:
> > Suggests: your package can use these packages, but doesn’t require them. You might use suggested packages for example datasets, to run tests, build vignettes, or maybe there’s only one function that needs the package.

Thanks for the suggestion. I think that having representative datasets for tests and examples is quite important. Also this is not the first package we author for release on Bioconductor. So I am quite aware of Hadleys suggestions and his reasoning, but in this is case I don't want to add complexity to code early on.

Also keep in mind, that we are talking about a data package and nothing functional. Lets keep this simple and let Tuomas fix the build errors, which largely have nothing to do with the packages brought up in the discussion.

yiluheihei:
@FelixErnst Sorry to disturb you. Thanks very much for your careful and patient explanation.

FelixErnst:
Thanks for understanding. Your points are all valid considerations. However, lets take one step at a time

TuomasBorman:
Hi, 

I ran R CMD check. There were few warnings, which I fixed. They were related e.g. to roxygen. Also, estimatedDominance's tests used microbiome package so I modified tests to use fixed values, which I got with microbiome package. 

Now, there is only one warning which is related to qpdf "‘qpdf’ is needed for checks on size reduction of PDFs".

microsud (approved):

FelixErnst:
@tvborm @yiluheihei I made a mistake. I meant to suggest that the `microbiomeDataSets` package be added into `Imports` and not `Depends`.  My bad, sory for the confusion.

@tvborm Locally this warning is quite common on certain OSes, but have a look at the GHA pipeline summary below the comment section. It still states: `ERROR: dependency ‘microbiomeDataSets’ is not available for package ‘mia’`.

This was due to the fact, that the prefix `github::` was missing. I forgot to add this in my comments. So again my fault. SHould be fixed now

edit: and apparently the GHA doesn't pick up the new dependency. Here goes another potential fix

TuomasBorman:
@FelixErnst Ok, thanks for clarifying the issue. I actually saw that error, but I could not figure out what caused that. I hoped that by solving those warning it would had disappeared.

PR #38, opened 2021-01-16
Transform abundance

TuomasBorman:
Hi!

Here is the transformAbundance. Is there some functions that should be removed, added or modified? I hope that now all git-related things are correct. 

-Tuomas


R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#' Boolean or numeric value deciding whether pseudocount is added. Numerical
+#' value specifies the value of pseudocount. When selecting pseudocount==TRUE,
+#' pseudocount is 1.
+#'
+#' @param scale
+#' A numeric value for scaling. Values are multiplied with specified value.
+#'
+#' @param target
+#' A single character value specifying, if z transformation is applied
+#' for samples or features.
```
antagomir: Use transformation per sample as the default, and here explicitly that this is the default. 

We could even consider removing this option entirely since it is confusing (typically these transformations are always per-sample basis). But in some cases we like to do it the other way for visualization purposes, so it could be useful to keep this. Just make sure that the default is what it should be, and clearly stated.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' @details
+#' \code{transformAbundance} applies transformation to abundance table.
+#' Transformation methods include
```
antagomir: I would rename this to `transformationMethod` or `transformAbundanceMethod` or something like this. Perhaps @FelixErnst likes to comment.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' \itemize{
+#'
+#'     \item{"relabundance" }{Transforms abundances to relative.
+#'     ($\frac{x}{x_{tot}}$, where $x$ is a single value and $x_{tot}$ is the sum of
+#'     all values.)}
+#'
+#'     \item{"Z" }{Z-transform or Z-standardization can be used for normalizing the data.
+#'     It can be done for samples or features by specifying 'target'. However, Z-transform
+#'     done for features can give misleading results. By default, it is done for samples.
```
antagomir: done for samples -> does this mean that the other transformations are done for the taxa? Clarify or if there is no difference, then remove. I think we should do it per sample by default for every transformation. Otherwise there is larger risk for user misunderstandings and errors.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'     \item{"hellinger" }{Hellinger transform can be used for reducing the impact of
+#'     extreme data points. ($\sqrt{\frac{x}{x_{tot}}}$, where $x$ is a single value and $x_{tot}$ is the sum of
+#'     all values.)}
+#'
+#'     \item{"identity" }{Identity is the data itself without transform.}
+#'
+#'     \item{"clr" }{Centered log ratio (clr) transform can be used for reducing the
+#'     skewness of data and for centering it.
+#'     ($log_{10}x_{r}) - log_{10}µ_{r}$, where $x_{r}$ is a single relative value, $µ_{r}$ is
+#'     the mean of relative values of whole sample.}
```
antagomir: It would be nice to be able to provide access to the transformations in the "compositions" R pkg (ilr, alr, at least; perhaps clr also more efficient there..?). But this would require "compositions" as an imported package. Could we have that @FelixErnst?

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' }
+#'
+#' @references
+#' If there is references Add information here......................................
```
antagomir: Cite [Gloor et al. ](https://www.frontiersin.org/articles/10.3389/fmicb.2017.02224/full) for compositional and clr transformations. In addition, you could cite the original references to CLR (check Gloor et al for that).

Also Hellinger transformation good to cite. Perhaps vegan package has the reference if you do not find otherwise. Ask if not. 

Others are standard, no need to cite anything specifically I think.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' @return \code{transformAbundance} returns \code{x} with additional, transformed
+#' abundance table named \code{*name*} in the \code{\link{assay}}.
+#' \code{getTransformAbundance} returns transformed abundance table.
+#'
+#' @seealso
+#' If there is something Add information here......................................
```
antagomir: I think we can remove this for now

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' @examples
+#' data(esophagus)
+#' x <- esophagus
+#'
+#' # By default, returns object with identity table
+#' x <- transformAbundance(x)
+#' assays(x)$identity
+#'
+#' # By specifying, it is possible to apoly different transformations, e.g. clr transform
```
antagomir: apoly -> apply

antagomir: transform -> transformation

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#' # Pseudocount can be added by giving the value TRUE. Then pseudocount is 1.
+#' x <- transformAbundance(x, transform="clr", pseudocount=TRUE)
+#' assays(x)$clr
+#'
+#' # Name of the stored table, can be specified. Also, the target of transform
+#' # can be specified with "abund_values". I addition to pseudocount=TRUE, it is
```
antagomir: I -> In

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#' # By default, returns object with identity table
+#' x <- transformAbundance(x)
+#' assays(x)$identity
+#'
+#' # By specifying, it is possible to apoly different transformations, e.g. clr transform
+#' # Pseudocount can be added by giving the value TRUE. Then pseudocount is 1.
+#' x <- transformAbundance(x, transform="clr", pseudocount=TRUE)
+#' assays(x)$clr
+#'
+#' # Name of the stored table, can be specified. Also, the target of transform
```
antagomir: table, can -> table can

tests/testthat/test-5transformAbundance.R
```diff
@@ -0,0 +1,12 @@
+context("transformAbundance")
+
+test_that("transformAbundance", {
+
+    # TSE object
+    data(GlobalPatterns)
+    tse <- GlobalPatterns
+
+    # Tests come here.......................
```
antagomir: great!

R/transformAbundance.R
FelixErnst: Maybe also make it a seperate function called `ZTransform`. If you structure the code taking this into account, you could probably reuse code from `transfromAbundance`

R/transformAbundance.R
FelixErnst: How about just simply `method`? The function is called `transformAbundance` and in combination `method` should be meaningful enough

R/transformAbundance.R
FelixErnst: Don't try to reinvent the wheel. Just link and use the existing `relAbundanceCounts` function.

R/transformAbundance.R
FelixErnst: I am not sure.

image
 
doesn't seam to heavy, but is it really worth it just for switching out the implementation? Also keep in mind, that when a `DelayedArray` appears as `assay`, `combinations` cannot be used in any event.

R/transformAbundance.R
FelixErnst: the `relAbundanceCounts` reference should appear here

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
+#' Transform Abundance
+#'
+#' This function applies transformation to abundance table, which is returned.
+#' Additionally, the information can be directly stored in the \code{assay}.
```
FelixErnst: Two function are discussed here. I think you can spell out, which function does the returning and which does the storing

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#' @param transform
+#' A single character value for selecting the transformation method.
+#'
+#' @param name
+#' A single character value specifying the name of transformed abundance table.
+#'
+#' @param pseudocount
+#' Boolean or numeric value deciding whether pseudocount is added. Numerical
+#' value specifies the value of pseudocount. When selecting pseudocount==TRUE,
+#' pseudocount is 1.
```
FelixErnst: I think this should accept, FALSE and and any numeric value. `TRUE` and then translating it two `1` is probably to specific. It robes the user of an option, which is easy to implement

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' @param scale
+#' A numeric value for scaling. Values are multiplied with specified value.
```
FelixErnst: I think `scalingFactor` would be better to avoid any confusion with the case R function `scale` and its arguments

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#' @details
+#' \code{transformAbundance} applies transformation to abundance table.
+#' Transformation methods include
+#'
+#' \itemize{
+#'
+#'     \item{"relabundance" }{Transforms abundances to relative.
+#'     ($\frac{x}{x_{tot}}$, where $x$ is a single value and $x_{tot}$ is the sum of
+#'     all values.)}
```
FelixErnst: I think this whole thing whould be merge with the `relAbundanceCounts` function. The parallels are already there and and additional man page is probably not necessary.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#'     \item{"log10p }{log10 transform can be used for reducing the skewness of the data.
+#'     With log10p, pseudo count is added before log10 transformation.
+#'     ($log_{10}x$, where $x$ is a single value of data.)}
+#'
+#'     \item{"hellinger" }{Hellinger transform can be used for reducing the impact of
+#'     extreme data points. ($\sqrt{\frac{x}{x_{tot}}}$, where $x$ is a single value and $x_{tot}$ is the sum of
+#'     all values.)}
+#'
+#'     \item{"identity" }{Identity is the data itself without transform.}
```
FelixErnst: Is this necessary? Seems to me like a complicated way of wrapping `assay(x,"y")`

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+                   name = transform,
+                   pseudocount = FALSE,
+                   scale = 1,
+                   target = "samples"){
+
+              # Get object with transformed table
+              x <- transformAbundance(x, abund_values, transform, name, pseudocount, scale, target)
+
+              # Get the table
+              mat <- assay(x, name)
```
FelixErnst: writing and extracting is not necessary.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+#'
+#'
+NULL
+
+#' @rdname transformAbundance
+#' @export
+setGeneric("transformAbundance", signature = c("x"),
+           function(x,
+                    abund_values = "counts",
+                    transform = "identity",
```
FelixErnst: If you include all available options like `transform = c("clr", "log10",...)` in the generic and the method, you can easily check the validity via `transform  <- match.arg(transform)`

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+              # Check abund_values
+              .check_abund_values(abund_values, x)
+
+              # Check transform
+              if(!(transform %in%
+                   c("relabundance", "Z", "log10", "log10p", "hellinger", "identity", "clr"))){
+                  stop("'transform' must be one of following methods 'relabundance',
+                       'Z', 'log10', 'log10p', 'hellinger', 'identity', 'clr'.",
+                       call. = FALSE)
+              }
```
FelixErnst: see above

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+              if(!.is_non_empty_string(name)){
+                  stop("'name' must be a non-empty single character value.",
+                       call. = FALSE)
+              }
+
+              # Check pseudocount
+              if(!(pseudocount==TRUE || pseudocount==FALSE || pseudocount>0)){
+                  stop("'pseudocount' must be boolean or positive numeric value.",
+                       call. = FALSE)
+              }
```
FelixErnst: see above

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+            list(assay = assay,
+                 target = target))
+}
+
+.get_relabundance_table <- function(assay, target){
+
+    # Calculates the relative abundances
+    mat <- sweep(assay, 2, colSums(assay), "/")
+    return(mat)
+}
```
FelixErnst: this exists already in `relAbundanceCounts`.

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+    # to calculate log from zeros. If there is no zeros, calculates log.
+    if (any(assay == 0)) {
+        stop("Abundance table contains zero and log10 transformation
+                is being applied without pseudocount. Try log10 with
+                pseudocount=TRUE or log10p transformation.")
+    } else {
+        mat <- log10(assay)
+    }
+    return(mat)
+}
```
FelixErnst: What about `NA`s?

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+
+.get_log10p_table <- function(assay, target){
+
+    # Because a pseudo count was added, there is no zeroes in the abundance table.
+    # Calculates log "directly".
+    mat <- log10(assay)
+
+    return(mat)
+}
```
FelixErnst: Have a look at `normalizeCounts` in the `scuttle` package. I think you should reuse the implementation from there. No need to reinvent the wheel

R/transformAbundance.R
```diff
@@ -0,0 +1,400 @@
...
+
+.get_identity_table <- function(assay, target){
+
+    # Identity table is abundance table itself
+    mat <- assay
+
+    return(mat)
+}
```
FelixErnst: see above

FelixErnst (changes requested):
Thanks for the PR. I think merging it with `relAbundanceCounts` is an apparent simplification to the whole process. Please rework it in this sense, but keep the `relAbundanceCounts` functions. The `relabundance` accessor can also be transferred and maybe you want to provide short cuts for all transformed types, if the are stored in their default slot.

Code structure look quite good so for. I will have a closer look, when its merged with `relAbundanceCounts`. 

However, I would like to point out, that the `getTransformedAbundance` function is really not necessary. For this task two options exist:

1. You provide a calculation function and wrap it in the setter function. e.g. `calcTransformAbundance` and `transformAbundance`.
2. You omit the calc function and prevent the user to waste computing resources and store the result by default and let the user retrieve the data via `assay` themselves. Therefore, `getTransformAbundance` has also the potential to trick an uninformed user into wasting resources like the calc function, but with the additional "waste" of storing the data and discarding a permanent storage as well. This is the save option and should probably be the only option.

R/transformAbundance.R
antagomir: Yes, that's good.

R/transformAbundance.R
antagomir: In fact I was thinking the same, then thought that there was some preference for these more complicated ("descriptive") names. I would go with 'method' as the argument name here, too.

R/transformAbundance.R
antagomir: How does this relate to Z score?

R/transformAbundance.R
antagomir: Right - it would be better to re-implement ilr (and perhaps alr). But this is not fast, perhaps better to open an issue about adding these and done at a later time.

R/transformAbundance.R
FelixErnst: Ìt doesn't. It was meant for the comment above, but somehow ended up here

R/transformAbundance.R
antagomir: Good point. 

Should we consider whether `relAbundanceCounts` is needed as an independent function, if `transformAbundance` does the job and is more generic? Is there added value from having `relAbundanceCounts`?

FelixErnst: Yesterday I would have said yes, now I am not so sure. It is probably the most used one I guess, especially for plotting, so thats why I implemented it

I think `relAbundanceCounts` should most likely be a wrapper around `transformAbundance`, but if we think broad picture I could also see a more generic function name `transformCounts` being used. The names aside, I think the implementation should be largely structured the same as `relAbundanceCounts` and `normalizeCounts`, which is already the case, but the redundancies should be removed

antagomir: Yes to all, I think this would be a clear solution.

antagomir:
Could you also add the method `pa` from [decostand](https://www.rdocumentation.org/packages/vegan/versions/2.4-2/topics/decostand)?

FelixErnst:
> Could you also add the method `pa` from [decostand](https://www.rdocumentation.org/packages/vegan/versions/2.4-2/topics/decostand)?

If this is just prsence-absence I think that is better implement here:
`as.integer(assay(x,"y") > 0)`

antagomir:
> > Could you also add the method `pa` from [decostand](https://www.rdocumentation.org/packages/vegan/versions/2.4-2/topics/decostand)?
> 
> If this is just prsence-absence I think that is better implement here:
> `as.integer(assay(x,"y") > 0)`

It would be more clear to have all standard transformations available through the single transformation wrapper (`transformCounts`), even if some of then can be easily obtained also through other routes (we also have "identity" method currently included). More so, because some classical methods explicitly ask for this particular transformation,.

Presence/absence could use a different threshold than 0, this would add one argument (although this can be of course also achieved by replacing 0 with something else in the above example).

R/transformAbundance.R
TuomasBorman: Removed "target", created function ZTransform()

R/transformAbundance.R
TuomasBorman: --> In

R/transformAbundance.R
TuomasBorman: changed

R/transformAbundance.R
TuomasBorman: scale --> scalingFactor

R/transformAbundance.R
TuomasBorman: Pseudocount can now get FALSE and any numeric values.

R/transformAbundance.R
TuomasBorman: Added information

R/transformAbundance.R
TuomasBorman: Changed

R/transformAbundance.R
TuomasBorman: I deleted those functions that only returned the transformed table.

R/transformAbundance.R
TuomasBorman: getTransformAbundance() deleted

R/transformAbundance.R
TuomasBorman: Changed

R/transformAbundance.R
TuomasBorman: Changed

R/transformAbundance.R
TuomasBorman: --> "'pseudocount' must be FALSE or numeric value."

R/transformAbundance.R
TuomasBorman: Should transform=relabundance still be a choice here in transformAbundance? I changed this part so that row 272 (shown above) is changed to use internal function .calc_rel_abund

antagomir: I think it should be an option.

R/transformAbundance.R
```diff
@@ -0,0 +1,415 @@
...
+# Chooses which transformation function is applied
+.get_transformed_table <- function(assay, transform, pseudocount, scalingFactor){
+
+    # If "pseudocount" is TRUE or over 0 or transform is log10p, add pseudocount
+    if(!pseudocount==FALSE || transform=="log10p"){
+        # In case of log10p, pseudocount can be still FALSE. Then 1 is added.
+        if(pseudocount == FALSE){
+            # Add 1 as a pseudo count
+            assay <- assay + 1
+            warning("Transform was calculated with pseudocount value 1")
```
FelixErnst: Imo this warning should include why a warning is thrown. `Transform was calculated with pseudocount value '1'. Set pseudocount to a numeric value to resolve this warning.`

R/transformAbundance.R
```diff
@@ -0,0 +1,415 @@
...
+    assay <- assay * scalingFactor
+
+    # Function is selected based on the "transform" variable
+    FUN <- switch(transform,
+                  relabundance = .get_relabundance_table,
+                  Z = .get_z_table,
+                  log10 = .get_log10_table,
+                  log10p = .get_log10p_table,
+                  hellinger = .get_hellinger_table,
+                  identity = .get_identity_table,
```
FelixErnst: Identity should be removed, since it is just `assay(x,"y")`

TuomasBorman: Are there situations where user wants to just multiply assay's values by scalingFactor? Should the "identity" be e.g. "scale", or should it be just removed?

R/transformAbundance.R
TuomasBorman: Ok

R/transformAbundance.R
antagomir: can be removed. legacy.

R/transformAbundance.R
antagomir: It is not common and probably not critical. Either remove, or do as proposed (scale).

FelixErnst: I suppose `assay(x,"z") <- assay(x,"y") * a` is the easiest solution and doesn't need a wrapper.

FelixErnst:
Can you update from master due to the upstream changes in TSE?

R/transformAbundance.R
TuomasBorman: Removed identity option

R/transformAbundance.R
TuomasBorman: Identity option removed

R/transformAbundance.R
TuomasBorman: Identity option removed

R/transformAbundance.R
```diff
@@ -359,6 +366,13 @@ setMethod("ZTransform", signature = c(x = "SummarizedExperiment"),
 
 .get_ztransformed_table <- function(assay, pseudocount, scalingFactor){
 
+    # If assay contains missing values
+    if(any(is.na(assay))){
+        # Missing values are replaced with value 0
+        assay[which(is.na(assay))] <- 0
+        warning("Table contains missing values. Missing values are replaced with value 0.")
+    }
+
```
FelixErnst: I think this might not the way to do it, since the result is actually changed by adding zeros. The solution probably hast to be aware of NAs and filter it out at the calculation step.

The idea would be to have a factor for subsetting the data in steps, which cannot handle NAs, like this:
```
f <- is.na(data)
data[f] <- FUN(data[f])
```

antagomir: I agree, this is potentially risky and should be changed now with the update.

antagomir: (or it could be an option to replace NAs by zeroes or imputation but perhaps it is more clear if the user can choose to do that outside of the function)

TuomasBorman: Now, missing values are replaced with 0 only when calculating relative values, but they are changed back to NA. In other cases where missing values cause problems, there are na.rm=TRUE

R/transformAbundance.R
TuomasBorman: Added references

R/transformAbundance.R
TuomasBorman: Warning changed

R/transformAbundance.R
TuomasBorman: NAs taken into account. Only case where NAs are replaced with 0 is when relative values are calculated. In that case, NAs are changed back to NAs after calculation. In other cases, NAs stay but there are na.rm=TRUE so they don't cause problems

R/transformAbundance.R
TuomasBorman: I have thought this a while... Do you mean log1p()? I can not find any other helpful things from the code. If so, that does the natural logarithm

antagomir: We like to do log10 in microbiome data, instead of natural log. But does it work with that change?

TuomasBorman: I am just wondering, what is the functionality that is wanted to be implemented also here. 

Well, my opinion might be very biased, but I think our code is quite straight-forward. All in all, log10 and log10p are the same. Just the pseudocount-adding differs there, and I think it is done in quite logical point, which can be easily found from the code. 

Also, user have choice to choose if the pseudocount is 2 and not 1 like it is by default in log10p.

antagomir: True. Thinking about it now, I do not think that we need both "log10" and "log10p" for this reason. We can just keep the "log10" with the option of pseudocount. I think the original idea had been to have a "quick" shortcut for log10 + pseudocount=1 but I guess it is potentially confusing and "log10p" option can be removed.

TuomasBorman: log10p removed

R/transformAbundance.R
TuomasBorman: Clarified

R/transformAbundance.R
TuomasBorman: "transform" parameter is now "method"

tests/testthat/test-5transformAbundance.R
TuomasBorman: Tests added

R/transformAbundance.R
TuomasBorman: Hi, now the code is quite ready. I think there is only one bigger thing left, and it is this. What do you mean by merging? When user chooses method = "relabundance", should I check that and then just get the table with relAundanceCounts()? 

Like this:

```
if(method=="relabundance"){
    relAbundanceCounts(x, abund_values, name)
}else{
    # Get transformed table
     transformed_table <- .get_transformed_table(assay = assay(x, abund_values),
                                                          method = method,
                                                          pseudocount = pseudocount,
                                                          scalingFactor = scalingFactor)

      # Assign transformed table to assays
      assay(x, name) <- transformed_table

        return(x)
}
```

antagomir: Yes by merging we mean that it should be accessible like this. 

I would not export the relative abundance methods separately (unless we export all specific transformations in the same way but I do not see a clear need).

FelixErnst: I mean with merging, that all the functionality of the `relAbundanceCounts` function and file, should be merged with `transformAbundance`. There should only be one file, one man page and one implementation. You can export a few different methods, but the results should only be generated via one pipeline.

So in order to merge it, copy paste the stuff you need into your file. Make sure, that the man page contains all the details from `relAbundanceCounts`.

I also want to point out, that there is the open conversation whether `transformCounts` wouldn't be a better name for the function.

antagomir: Yes, I agree. Felix put it more exactly.

I have in fact commented in the open conversation that I support `transformCounts`.

FelixErnst: Just to make maybe more descriptive: `transformCounts` should be the main function. An implementation for `relAbundanceCounts` could look like this:

```
setMethod("relAbundanceCounts", signature = c(x = "SummarizedExperiment"),
function(x, ...){
  transformCounts(x, method="relabundance", ...)
})
```

antagomir: Do we really need a separate `relAbundanceCounts` wrapper (and other similar ones) or would just `transformCounts` be enough?

FelixErnst: I would like to keep. It is probably the most used transform functions. In the end it will be 6 lines of code.

antagomir: ok (still thinking if `relAbundanceCounts` is the most descriptive name for this but we can keep it unless anyone comes up with other suggestions)

microsud: I would also agree with @FelixErnst . Maybe a bit controversial, but do you see the possibility of using these options `relProportion` or `proportionalCounts`?

TuomasBorman: relAbundanceCounts merged to transformCounts

R/transformAbundance.R
TuomasBorman: transformAbundance's name changed to transformCounts

R/transformCounts.R
```diff
@@ -28,7 +28,7 @@
 #' A numeric value for scaling. Values are multiplied with specified value.
```
microsud: scalingfactor -> scalingFactor 
Edit: this is throwing one of the warnings during build because of lowercase.

R/transformCounts.R
```diff
@@ -117,21 +117,21 @@
...
                     scalingFactor = 1)
-               standardGeneric("transformAbundance"))
+               standardGeneric("transformCounts"))
 
 
-#' @rdname transformAbundance
+#' @rdname transformCounts
 #' @export
-setMethod("transformAbundance", signature = c(x = "SummarizedExperiment"),
+setMethod("transformCounts", signature = c(x = "SummarizedExperiment"),
```
microsud, line 129: should the `setMethods` in the codes also have `scalingFactor = 1` option.?

R/transformCounts.R
```diff
@@ -1,7 +1,7 @@
-#' Transform Abundance
+#' Transform Counts
 #'
 #' These functions applies transformation to abundance table. By using
```
microsud: Alternative: "These functions provide a variety of options for transforming abundance data." By using ...

R/transformCounts.R
antagomir: I think we previously decided to drop the `scale` transformation. But if there seems to be a clear use/need, then it would be good to add back. Is `scalingFactor` needed anywhere else?

microsud: It was in the build warnings. Therefore, I highlighted it. If not needed then it should be removed. For specific cases the user can do the scaling by themselves.

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+                pseudocount = 1 or other numeric value.")
+    }
+
+    mat <- .calc_rel_abund(assay)
+
+    # In every sample, calculates the log of individual entries. After that calculates
+    # the sample-specific mean value and subtracts every entries' value with that.
+    mat <- apply(mat, 2, function(x) {
+        log(x) - mean(log(x), na.rm=TRUE)
+    })
```
FelixErnst:
```
mat <- matrix(1:200000,2000,100)

FUN1 <- function(){
    apply(mat, 2, function(x) {
        log(x) - mean(log(x), na.rm=TRUE)
    })
}

FUN2 <- function(){
    clog <- log(mat)
    clogm <- colMeans(clog)
    mat <- t(t(clog) - clogm)
    mat
}

bench::mark(FUN1(),FUN2())
```

FUN2 is faster and uses less memory

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+
+    mat <- log10(assay)
+
+    return(mat)
+}
+
+.get_pa_table <- function(assay){
+
+    # If value is over zero, gets value 1. If value is zero, gets value 0.
+    mat <- ifelse(assay > 0, 1, 0)
```
FelixErnst:
```
mat <- matrix(1:200000,2000,100)

FUN1 <- function(){
    ifelse(mat > 0, 1, 0)
}

FUN2 <- function(){
    matrix(as.integer(mat > 0),nrow(mat))
}

bench::mark(FUN1(),FUN2())
```

FUN2 is faster and uses less memory

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+                pseudocount = 1 or other numeric value.")
+    }
+
+    # Gets the log transformed table
+    mat <- .get_log10_table(assay)
+
+    # Performs z transformation
+    mat <- apply(mat, 2, function(x) {
+        (x - mean(x, na.rm=TRUE))/sd(x, na.rm=TRUE)
+    })
```
FelixErnst: This is actually implemented twice. See line 494-496

```
mat <- matrix(1:200000,2000,100)

FUN1 <- function(){
    apply(mat, 2, function(x) {
        (x - mean(x, na.rm=TRUE))/sd(x, na.rm=TRUE)
    })
}

FUN2 <- function(){
    cm <- colMeans(mat, na.rm = TRUE)
    csd <- colSds(mat, na.rm = TRUE)
    t((t(mat) - cm)/csd)
}

bench::mark(FUN1(),FUN2())
```

FUN2 is faster and uses less memory

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+    # Stores the indices of missing values
+    missingValues <- which(is.na(assay))
+    # Missing values are replaced with value 0
+    assay[missingValues] <- 0
+
+    # Calculates the relative abundances. Uses internal function from relabundance.R.
+    mat <- sweep(assay, 2, colSums(assay), "/")
+
+    # If there were missing values, add them back to data
+    mat[missingValues] <- NA
```
FelixErnst:
```
mat <- matrix(1:200000,2000,100)

FUN1 <- function(){
    missingValues <- which(is.na(mat))
    # Missing values are replaced with value 0
    mat[missingValues] <- 0

    # Calculates the relative abundances. Uses internal function from relabundance.R.
    mat <- sweep(mat, 2, colSums(mat), "/")

    # If there were missing values, add them back to data
    mat[missingValues] <- NA
    mat
}

FUN2 <- function(){
    sweep(mat, 2, colSums(mat, na.rm = TRUE), "/")
}

bench::mark(FUN1(),FUN2())
```

FUN2 is only a tiny bit faster, but uses less memory

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+
+    # Z transform for features. Centers the feature data. After that, divides with
+    # the standard deviation of feature.
+    # Performs z transformation
+    trans <- apply(trans, 2, function(x) {
+        (x - mean(x, na.rm=TRUE))/sd(x, na.rm=TRUE)
+    })
+
+    # Transposes the table back to its original form
+    trans <- t(trans)
```
FelixErnst: Since you only use z transformation on a per feature basis, this implementation might actually work better:

```
mat <- matrix(1:200000,2000,100)
FUN1 <- function(){
    t(apply(t(mat), 2, function(x) {
        (x - mean(x, na.rm=TRUE))/sd(x, na.rm=TRUE)
    }))
}

FUN2 <- function(){
    rm <- rowMeans(mat, na.rm = TRUE)
    rsd <- rowSds(mat, na.rm = TRUE)
    (mat - rm)/rsd
}
bench::mark(FUN1(),FUN2())
```

FUN2 is much faster and uses a lot less memory

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+                    scalingFactor = 1)
+               standardGeneric("transformCounts"))
+
+
+#' @rdname transformCounts
+#' @export
+setMethod("transformCounts", signature = c(x = "SummarizedExperiment"),
+          function(x,
+                   abund_values = "counts",
+                   method = c("relabundance", "log10", "pa", "Z", "hellinger", "clr"),
```
FelixErnst: Z transform shouldn't be available through this function anymore, just to avoid any accidents

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+
+              # Multiply values with scalingFactor. By default, scalingFactor is 1, so no changes are made
+              assay(x, abund_values) <- assay(x, abund_values) * scalingFactor
+
+              # Get and store relabundance table
+              assay(x, name) <- .calc_rel_abund(assay(x, abund_values))
+
+              return(x)
+          }
+)
```
FelixErnst:
```
#' @rdname transformCounts
setGeneric("relAbundanceCounts", signature = c("x"),
           function(x, ...)
               standardGeneric("relAbundanceCounts"))

#' @rdname transformCounts
#' @importFrom SummarizedExperiment assay assay<-
#' @export
setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
    function(x, ...){
        transformCounts(x, method = "relabundance", ...)
    }
)
```

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+              # If the method is "relabundance", relAbundanceCounts is used
+              if(method == "relabundance"){
+
+                  # Get relative abundances stored in assays
+                  x <- relAbundanceCounts(x,
+                                          abund_values,
+                                          name,
+                                          pseudocount,
+                                          scalingFactor)
+              }
```
FelixErnst: not necessary anymore. See below

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+    if(!pseudocount==FALSE){
+        assay <- assay + pseudocount
+    }
+
+    # Multiply values with scalingFactor. By default, scalingFactor is 1, so no changes are made
+    assay <- assay * scalingFactor
+
+    # Function is selected based on the "method" variable
+    FUN <- switch(method,
+                  Z = .get_z_table,
```
FelixErnst: replace with 
```
relabundance = .calc_rel_abund,
```

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+# Chooses which transformation function is applied
+.get_transformed_table <- function(assay, method, pseudocount, scalingFactor){
+
+    # If "pseudocount" is not FALSE, it is numeric value specified by user. Then add pseudocount.
+    if(!pseudocount==FALSE){
+        assay <- assay + pseudocount
+    }
+
+    # Multiply values with scalingFactor. By default, scalingFactor is 1, so no changes are made
+    assay <- assay * scalingFactor
```
FelixErnst: Since you need this for Z transform as well, put this into a seperate functions


```
    # If "pseudocount" is not FALSE, it is numeric value specified by user. Then add pseudocount.
    if(!pseudocount==FALSE){
        mat <- mat + pseudocount
    }
    mat
}

.apply_scalingFactor <- function(mat, scalingFactor){
    # Multiply values with scalingFactor. By default, scalingFactor is 1,
    # so no changes are made
    if(scalingFactor != 1){
        mat <- mat * scalingFactor
    }
    mat
}
```

Use this where needed

```
assay <- .apply_pseudocount(assay, pseudocount)
assay <- .apply_scalingFactor(assay, scalingFactor)
```

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+        stop("Abundance table contains zero and log10 transformation
+                is being applied without pseudocount. Try log10 with
+                pseudocount = 1 or other numeric value.")
+        }
+    if (any(assay < 0, na.rm = TRUE)) {
+            stop("Abundance table contains negative values and log10 transformation
+                is being applied without pseudocount. Try log10 with
+                pseudocount = 1 or other numeric value.")
+    }
+
```
FelixErnst: Creating the logical vector twice is probably to much overhead. Just do

```
# Log10 can not be calculated if there is zero
    if (any(assay <= 0, na.rm = TRUE)) {
        stop("Abundance table contains zero or negative values and ",
             "Z-transformation is being applied without pseudocount. ",
             "Try ZTransform with pseudocount = 1 or other numeric value.",
             call. = FALSE)
    }
```

Applies to all other cases in these functions as well

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+        stop("Abundance table contains zero and Z-transformation
+            is being applied without pseudocount. Try ZTransform with
+            pseudocount = 1 or other numeric value.")
+    }
+    # If there is negative values, gives an error.
+    if (any(assay < 0, na.rm = TRUE)) {
+        stop("Abundance table contains negative values and Z-transformation
+                is being applied without pseudocount. Try ZTransform with
+                pseudocount = 1 or other numeric value.")
+    }
```
FelixErnst: Not needed. Gets tested in `.get_z_table`

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+            under detectable limit.")
+
+        # Some features are undetectable in all samples, they are NA when scaled.
+        # 0 is assigned to those features.
+
+        trans[names(undetectables), ] <- 0
+
+        mat <- trans
+
+    }
```
FelixErnst: I think is too much sugar on top of the function. If any rows are not detectable before the transformation, they are not after. `NA`s shouldn't be introduced by the transformation in itself, should they?

R/transformCounts.R
```diff
@@ -0,0 +1,523 @@
...
+              # If the method is something else, internal function is used
+              else{
+              # Get transformed table
+              transformed_table <- .get_transformed_table(assay = assay(x, abund_values),
+                                                          method = method,
+                                                          pseudocount = pseudocount,
+                                                          scalingFactor = scalingFactor)
+
+              # Assign transformed table to assays
+              assay(x, name) <- transformed_table
```
FelixErnst: You probably need to set `withDimnames` to `FALSE` since some transformations strip the names:

```
assay(x, name, withDimnames=FALSE) <- transformed_table
```

FelixErnst (changes requested):
Hi Tuomas @tvborm 

I added some suggestions to improve performance and code structure. Have a look and let me know, if any points don't make sense.

Felix

R/transformCounts.R
TuomasBorman: Oh, thanks. I did not see that

Changed

R/transformCounts.R
TuomasBorman: Ok, I will remove scalingFactor

R/transformCounts.R
TuomasBorman: That sounds better, changed

R/transformCounts.R
TuomasBorman: Removed

R/transformCounts.R
TuomasBorman: Thanks for the suggestions. As you said, it seems that FUN2 makes the same job more efficiently.

Changed

R/transformCounts.R
TuomasBorman: Yep, same job but little bit quicker and with half of the space

Changed

R/transformCounts.R
TuomasBorman: Added

R/transformCounts.R
TuomasBorman: I thought different way, but there is much more sense to do it this way

Changed

R/transformCounts.R
TuomasBorman: Yep, it probably does not give any additional value to check them separately. User get enough information with that

Changed

R/transformCounts.R

R/transformCounts.R
TuomasBorman: Created only pseudocount applying function. There is no scalingFactor anymore.

R/transformCounts.R
TuomasBorman: So, now transformCounts(x, method="Z") does z-transformation for columns. ZTransform(x) does z-transformation for rows. 

So, there should be only one function (ZTransform) that does z-transformation for rows? Z-transformation for columns should not be implemented?

R/transformCounts.R
TuomasBorman: Removed

R/transformCounts.R
antagomir: This is not common in this context, so I propose removing it now.

R/transformCounts.R
TuomasBorman: yeah, it was much of quicker

Changed

R/transformCounts.R
TuomasBorman: replaced

R/transformCounts.R
TuomasBorman: You're right, values should not be changed that way

Now, code does not do that

R/transformCounts.R
TuomasBorman: OK, it is now removed. There is only z transformation for features

R/transformCounts.R
TuomasBorman: Z transformation for features is removed, so no need for this anymore

R/transformCounts.R
TuomasBorman: There is now only one function for z transformation, and it does transformation for features.

FelixErnst (approved):

FelixErnst:
Thanks Tuomas!

antagomir:
What a journey. Whoa!

PR #47, opened 2021-01-27
Added function that converts TSE to phyloseq

TuomasBorman:
Hi!

Here is the function that creates a phyloseq object from TSE. There is one problem with tests. Even though phyloseq functions are like "phyloseq::otu_table()", they are not found. I tested run those same codes outside from "testthat" and they work just fine, and everything seems to be in right place in the phyloseq object. 

Also, because I did not found any data that contains referenceseq, I could not test that it goes to the right place in the phyloseq object.

-Tuomas


antagomir:
Did you try to add: `library(phyloseq)` to load it in the test file?

antagomir (commented):
Seems fine. We can discuss the problem with tests separately in the PR comments.

TuomasBorman:
> Did you try to add: `library(phyloseq)` to load it in the test file?

Ok, found the error. The error was that the function itself did not have "phyloseq::" when using phyloseq's functions

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
+#' Create a phyloseq object from a TreeSummarizedExperiment object
+#'
```
FelixErnst, line 2: I know its quite redundant, but a description is needed.

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+#' @name makePhyloseqFromTreeSummarizedExperiment
+#' @export
+#'
+#' @author Leo Lahti and Tuomas Borman. Contact: \url{microbiome.github.io}
+#'
+#' @examples
+#' # Get TSE object
+#' data("GlobalPatterns")
+#' TSE <- GlobalPatterns
```
FelixErnst: TSE -> tse?

Consistency is king

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+#'
+#' # Create a phyloseq object from it
+#' phy <- makePhyloseqFromTreeSummarizedExperiment(TSE)
```
FelixErnst: For generating copy & paste examples and also generating nice pkgdown output, add a line printing the `phy`object.
Like this:

```
phy <- makePhyloseqFromTreeSummarizedExperiment(TSE)
phy
```

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+#'
+#' # By default the chosen table is counts, but if there is other tables,
+#' # they can be chosen with abund_values.
+#'
+#' # Counts relative abundances table
+#' TSE <- transformCounts(TSE, method = "relabundance")
+#' phy2 <- makePhyloseqFromTreeSummarizedExperiment(TSE, abund_values = "relabundance")
```
FelixErnst: See above

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+#' @rdname makePhyloseqFromTreeSummarizedExperiment
+#' @export
+setMethod("makePhyloseqFromTreeSummarizedExperiment", signature = c(x = "SummarizedExperiment"),
+          function(x, abund_values = "counts"){
+
+            # Input check
+            # Check object
+            if(!is(x,"TreeSummarizedExperiment")){
+                stop("'x' must be a 'TreeSummarizedExperiment' object")
+            }
```
FelixErnst: This has to be restructured. Since you know, which type of object your are handling, you can handle all the different `SE` classes in a nested way.

```
setGeneric("makePhyloseqFromTreeSummarizedExperiment", signature = c("x"),
            function(x, ...)
                standardGeneric("makePhyloseqFromTreeSummarizedExperiment"))

setMethod("makePhyloseqFromTreeSummarizedExperiment", signature = c(x = "SummarizedExperiment"),
           function(x, abund_values = "counts"){
        # do conversion to all the counts data, sample data and feature data
    }
)

setMethod("makePhyloseqFromTreeSummarizedExperiment", signature = c(x = "TreeSummarizedExperiment"),
           function(x, ...){
        ans <- callNextMethod()
        # add tree data if present, add sequence data if present
        ans
    }
)
```

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+
+            # If rowData has information, stores it to tax_table and converts is to phyloseq's tax_table.
+            # Otherwise, tax_table is NULL, and it is converted to phyloseq's tax_table.
+            if(nrow(rowData(x))>0 && ncol(rowData(x))>0){
+                tax_table <- as.matrix(rowData(x))
+                tax_table <- phyloseq::tax_table(tax_table)
+            } else {
+                tax_table <- NULL
+                tax_table <- phyloseq::tax_table(tax_table, errorIfNULL = FALSE)
+            }
```
FelixErnst: Does phyloseq support non-taxonomic data as tax_table? I don't think so, since it is a matrix and therefore can only be of type `character`.

It would be best to subset to taxonomic data only using `rowData(x)[,taxonomyRanks(x)]`

the else is also not necessary.

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+            .check_abund_values(abund_values, x)
+
+            # Get the assay from x
+            otu_table <- assay(x, abund_values)
+            # Convert to otu_table
+            otu_table <- phyloseq::otu_table(otu_table, taxa_are_rows = TRUE)
+
+            # If rowData has information, stores it to tax_table and converts is to phyloseq's tax_table.
+            # Otherwise, tax_table is NULL, and it is converted to phyloseq's tax_table.
+            if(nrow(rowData(x))>0 && ncol(rowData(x))>0){
```
FelixErnst: `nrow(rowData(x))` is never zero. (Except, if the SE has zero rows). In this case I would just refuse to convert the SE with an initial check

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+                tax_table <- as.matrix(rowData(x))
+                tax_table <- phyloseq::tax_table(tax_table)
+            } else {
+                tax_table <- NULL
+                tax_table <- phyloseq::tax_table(tax_table, errorIfNULL = FALSE)
+            }
+
+            # If colData has information, stores it to sample_data and converts is to phyloseq's sample_data.
+            # Otherwise, sample_data is NULL, and it is converted to phyloseq's sample_data.
+            if(nrow(colData(x))>0 && ncol(colData(x))>0){
```
FelixErnst: See above. Just check that `all(dim(x) > 0)` at the beginning and be done with it

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+                sample_data <- data.frame(colData(x))
+                sample_data <- phyloseq::sample_data(sample_data)
+            } else {
+                sample_data <- NULL
+                sample_data <- phyloseq::sample_data(sample_data, errorIfNULL = FALSE)
+            }
```
FelixErnst: Not necessary.

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+
+            # If rowTree has information, stores it to phy_tree and converts is to phyloseq's phy_tree.
+            # Otherwise, phy_tree is NULL, and it is converted to phyloseq's phy_tree.
+            if(!is.null(rowTree(x))){
+                phy_tree <- rowTree(x)
+                phy_tree <- phyloseq::phy_tree(phy_tree)
+            } else {
+                phy_tree <- NULL
+                phy_tree <- phyloseq::phy_tree(phy_tree, errorIfNULL = FALSE)
+            }
```
FelixErnst: Not necessary

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+
+            # If referenceSeq has information, stores it to refseq and converts is to phyloseq's refseq.
+            # Otherwise, refseq is NULL, and it is converted to phyloseq's refseq.
+            if(!is.null(referenceSeq(x))){
+                refseq <- referenceSeq(x)
+                refseq <- phyloseq::refseq(refseq)
+            } else {
+                refseq <- NULL
+                refseq <- phyloseq::refseq(refseq, errorIfNULL = FALSE)
+            }
```
FelixErnst: Needs to be restructured. See above.

else code path is not necessary

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,113 @@
...
+
+            # Creates a phyloseq object
+            phyloseq <- phyloseq::phyloseq(otu_table,
+                                 tax_table,
+                                 sample_data,
+                                 phy_tree,
+                                 refseq)
```
FelixErnst: Since the phyloseq constructor dispatches only to `...`, it would be better to construct a list of arguments and call `do.call(phyloseq::phyloseq,args)`

With this call you don't need to set elements ,if they are not present. This removes close to 50% of the code you need to maintain in this function.

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,47 @@
...
+    skip_if_not_installed("phyloseq")
+
+    # TSE object
+    data(GlobalPatterns)
+    tse <- GlobalPatterns
+
+    phy <- makePhyloseqFromTreeSummarizedExperiment(GlobalPatterns)
+
+    # Test that assay is in otu_table
+    testthat::expect_equal(as.data.frame(phyloseq::otu_table(phy)@.Data), as.data.frame(assays(tse)$counts))
```
FelixErnst: the `testthat::` prefix is not necessary. If you want to test this locally either load `testthat` in your session or use `devtools::test(filter="5makePhyloseq")`

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,47 @@
+context("makePhyloseqFromTreeSummarizedExperiment")
```
FelixErnst: This is an IO function in the most general sense. Please move the testing to the IO test file `test-4IO.R`

FelixErnst (changes requested):
Hi Tuomas,

thanks for the PR. The code needs to be restructured a bit. Let me know, if anything is unclear.

Felix

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
```diff
@@ -0,0 +1,47 @@
...
+    # Test that colData is in sample_table
+    testthat::expect_equal(phyloseq::sample_data(phy),
+                           phyloseq::sample_data(data.frame(colData(tse))))
+
+    # Test that rowTree is in phy_tree
+    testthat::expect_equal(phyloseq::phy_tree(phy), rowTree(tse))
+
+    # Test that referenceSeq is in refseq. There is no referenceseq in TSE, so
+    # referenceseq slot is not made in the phyloseq object
+    #testthat::expect_equal(phyloseq::refseq(phy), referenceSeq(tse))
```
FelixErnst: The referenceSeq slot is present, but the value `NULL`

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Added

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Yep, true

Changed

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Added

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: TSE --> tse
added phy2 printing

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Changed

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Added checking

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Added all(dim(x) > 0) at the beginning

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Removed

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Removed

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Code is now using do.call()

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Restructured

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Restructured

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Removed testthat::

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Moved

tests/testthat/test-5makePhyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Now the function has new structure.

So, if tse's referenceSeq has value NULL, phyloseq's refseq should be also NULL? 

Now the function does not add NULLs, because e.g. 'phyloseq' package's 'GlobalPatterns' data does not have NULL in refseq even though the corresponding tse have NULL.

FelixErnst: The comment was a comment on the comment 😄 

`TSE` has a `referenceSeq` slot, which by default is `NULL`. So, yes, if `is.null(referenceSeq(tse))`, the `phyloseq` object also shouldn't have sequence data. How `phyloseq` internally stores the sequence, I have never bothered to look up.

TuomasBorman: Oh, okay, that explains...

FelixErnst:
Can you fix the error from the builds?

FelixErnst (approved):

PR #50, opened 2021-01-28
Fixed a bug of loadFromQIIME2

TuomasBorman:
Hi!

I found one bug from the code that caused the problem that Chandler had. The problem was that, if there was no row that had "#q2:types" information, there was no value for idx. When there was no value for idx, the whole table was accidentally deleted. 

I tested the code, and now it seems to work with Chandler's data.

### DIFF
--- a/R/loadFromQIIME2.R
+++ b/R/loadFromQIIME2.R
@@ -233,9 +233,13 @@ loadFromQIIME2 <- function(featureTableFile,
     sam <- read.table(file = file, header = TRUE, sep = "\t", comment.char = "")
     rownames(sam) <- as.character(sam[, 1])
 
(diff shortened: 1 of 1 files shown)


FelixErnst:
Wasn't the issue about a function not found? Can you elaborate on this?

TuomasBorman:
image

I haven't followed this thing that much, but Chandler found the function eventually. After running the function, he got the error that is in the picture. Then he send the data to me, and I got the same error. 

Then I found that the problem was in the sample metadata file

I compared the data with example data, and the difference was that Chandler's data did not have that "#q2:types" row. Absence of that row caused that whole table was deleted, and that empty table caused that error.

PR #53, opened 2021-02-02
Changes to man page

TuomasBorman:
Hi!

Now different methods in transformCounts and different indices in estimateDominance are correctly indented. The problem right now is that \frac or \sum is not recognized which gives warnings. I tried to resolve the problem, but I don't know what is the problem and google didn't really help that much... 

They were not recognized before either, but it didn't cause warnings.

-Tuomas

image

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -24,6 +24,9 @@
 #' FALSE or numeric value deciding whether pseudocount is added. Numerical
 #' value specifies the value of pseudocount.
 #'
+#' @param threshold
+#' A numeric value for setting threshold for pa transformation. By default it is 0.
+#'
 #' @param ... additional arguments
 #'
 #' @details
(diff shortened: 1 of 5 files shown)


antagomir:
Does the [maths section in R extensions manual](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Mathematics) help?

R/transformCounts.R
```diff
@@ -28,38 +28,36 @@
...
 #' \code{transformCounts} applies transformation to abundance table.
-#' Transformation methods include
+#' Provided transformation methods include:
 #'
 #' \itemize{
+#' \item{"relabundance" }{Transforms abundances to relative. Generally, all microbiome
+#' data are compositional. That is, e.g., because all measuring instruments have their capacity limits.
+#' To make results comparable with other results, values must be relative.
+#' ($\frac{x}{x_{tot}}$, where $x$ is a single value and $x_{tot}$ is the sum of
+#' all values.) (Gloor et al. 2017.)}
```
antagomir: see e.g. Gloor et al. 2017

FelixErnst:
> Does the [maths section in R extensions manual](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Mathematics) help?

Exactly. Thats section 2.6. 

The examples given there include `\frac`, so I assume you have to adjust the latex commands to the Rd flavour for them to work correctly.

R/transformCounts.R
```diff
@@ -28,38 +28,36 @@
...
-#'     To make results comparable with other results, values must be relative.
-#'     ($\frac{x}{x_{tot}}$, where $x$ is a single value and $x_{tot}$ is the sum of
-#'     all values.) (Gloor et al. 2017.)}
+#' \item{"log10" }{log10 transformation can be used for reducing the skewness of the data.
+#' ($log_{10}x$, where $x$ is a single value of data.)}
 #'
-#'     \item{"log10" }{log10 transformation can be used for reducing the skewness of the data.
-#'     ($log_{10}x$, where $x$ is a single value of data.)}
+#' \item{"pa" }{Transforms table to presence/absence table. If value is over 0,
+#' then value is 1. If value is 0, then value is 0.)}
```
antagomir: Is there no threshold option? By default 0, yes, but it could be useful to accept changing that threshold.

Then you can say: "All abundances higher than epsilon are transformed to 1 (present), otherwise 0 (absent)."

R/transformCounts.R
```diff
@@ -28,38 +28,36 @@
...
 #'
-#'     \item{"pa" }{Transforms table to presence/absence table. If value is over 0,
-#'     then value is 1. If value is 0, then value is 0.)}
+#' \item{"Z" }{Z-transformation or Z-standardization can be used for normalizing the data.
```
antagomir: Z-transformation or Z-standardization normalizes the data by shifting (to mean \mu) and scaling (to standard deviation \sigma).

R/transformCounts.R
```diff
@@ -28,38 +28,36 @@
...
+#' extreme data points. It can be utilize for clustering or ordination analysis.
+#' ($\sqrt{\frac{x}{x_{tot}}}$, where $x$ is a single value and $x_{tot}$ is the sum of
 #'     all values.) (Legendre & Gallagher 2001.)}
 #'
-#'     \item{"clr" }{Centered log ratio (clr) transformation can be used for reducing the
-#'     skewness of data and for centering it.
-#'     ($log_{10}x_{r}) - log_{10}µ_{r}$, where $x_{r}$ is a single relative value, $µ_{r}$ is
-#'     the mean of relative values of whole sample.) (Gloor et al. 2017.)}
-#'
+#' \item{"clr" }{Centered log ratio (clr) transformation can be used for reducing the
```
antagomir: for reducing -> to reduce

R/transformCounts.R
```diff
@@ -28,38 +28,36 @@
...
+#' skewness of data and for centering it.
+#' ($log_{10}x_{r}) - log_{10}µ_{r}$, where $x_{r}$ is a single relative value, $µ_{r}$ is
+#' the mean of relative values of whole sample.) (Gloor et al. 2017.)}
```
antagomir: or just "mean relative value"

R/transformCounts.R
TuomasBorman: --> see e.g.

R/transformCounts.R

R/transformCounts.R
TuomasBorman: --> to reduce

R/transformCounts.R

R/transformCounts.R
TuomasBorman: That would be a nice addition

TuomasBorman: Added

TuomasBorman:
I added threshold for pa. 

Now man page looks like this. If I didn't understand wrong e.g. frac will be rendered better in html version... 

image

FelixErnst:
The `$x$` still looks weird to me. Wouldn`t it be best to use `\eqn{x}` to generate the same italic format?

> Whereas \eqn is used for “inline” formulae (corresponding to TeX’s $…$),

PR #55, opened 2021-02-17
implemented evenness estimate

FelixErnst:


R/estimateEvenness.R
```diff
@@ -0,0 +1,177 @@
+#' Estimate Evenness
+#'
+#' Several functions for calculation of evenness estimates are available.
+#'
+#' These includes the \sQuote{Camargo}, \sQuote{Pielou}, \sQuote{Simpson},
```
antagomir: includes -> include

R/estimateEvenness.R
```diff
@@ -0,0 +1,177 @@
...
+#'
+#' @param abund_values the name of the assay used for calculation of the
+#'   sample-wise estimates
+#'
+#' @param index a evenness measurement
+#'
+#' @param name a name for the column of the colData the results should be stored
+#'   in.
+#'
+#' @param threshold a numeric threshold. assay values below or equal to this
```
antagomir: I find this confusing that lowercase is used when starting sentences. Or is some convention to use in manpages?

R/estimateEvenness.R
```diff
@@ -0,0 +1,177 @@
...
+    # Bulla's Evenness
+    (O - 1/S)/(1 - 1/S)
+}
+
+# Camargo's eveness x: species counts zeroes: include zeros Inspired
+# by code from Pepijn de Vries and Zhou Xiang at
+# researchgate.net/post/How_can_we_calculate_the_Camargo_evenness_index_in_R
+# but rewritten here
+.calc_camargo_evenness <- function(mat) {
+    N <- colSums2(mat > 0, na.rm = TRUE)
```
antagomir, line 96: This operation is also used in several other evenness indices. Some performance gains are possible if this would be calculated before calling these functions (as they are all ran by default ie this operation gets executed multiple times).

R/estimateEvenness.R
```diff
@@ -0,0 +1,177 @@
...
+            for (j in s) {
+                xx <- xx + sum(abs(mat[(j + 1):n,i] - mat[j,i]))
+            }
+            xx
+        },
+        seq_along(N),
+        N,
+        seq)
+    # Return
+    1 - x/(colSums2(mat, na.rm = TRUE) * N)
```
antagomir, line 112: same comment with colSums2(mat, na.rm=TRUE) - used multiple times in the different evenness calculators.

R/estimateEvenness.R
```diff
@@ -0,0 +1,177 @@
...
+
+    # Log abundance
+    a <- log(mat)
+    a[is.na(a) | is.infinite(a)] <- 0
+
+    # Richness
+    S <- colSums2(mat > 0, na.rm = TRUE)
+
+    b <- t(a)/S
+    c <- rowSums2(b, na.rm = TRUE)
```
antagomir: Instead of t()/S + rowSums2 one could use a + colSums2 / t -> perhaps more efficient..?

antagomir (commented):
Very nice.

R/estimateEvenness.R
FelixErnst: The transpose has to be done before the division, since the division only works row-wise

R/estimateEvenness.R
FelixErnst: Wrapping these single functional calls, wouldn't have any performance effects and maintainability would also not be better. Arguments to function would need to be fix and the interdepences between the indicies would make it hard to reason about.

R/estimateEvenness.R
FelixErnst: see below

R/estimateEvenness.R
FelixErnst: changed

R/estimateEvenness.R
FelixErnst: fixed

R/estimateEvenness.R
antagomir: right but after this the sum is taken over rows so the division by S could happen also after the rowSums..?

R/estimateEvenness.R
antagomir: Performance is fine already, so this is a minor issue. But I do not see how it could not lead to performance improvement as the same operation gets calculated now multiple times in different functions. Instead of calculating this 4-5 times it would be possible to do just once and feed into the functions. I agree that this does not necessarily improve maintainability.

FelixErnst: I wouldn't touch it.

Any optimisation in this regard, would have to be done by the introduction of if clauses and I really don't want to catch and arrange all the cases needed to be covered

R/estimateEvenness.R
FelixErnst: I will look into this

FelixErnst: done. Thanks for the tip

microsud (approved):

antagomir (approved):

PR #71, opened 2021-03-03
Implemented coverage index

TuomasBorman:
Hi!

I implemented coverage index. 

-Tuomas


tests/testthat/test-0diversity.R
```diff
@@ -3,12 +3,13 @@ context("diversity estimates")
 test_that("diversity estimates", {
     skip_if_not(requireNamespace("vegan", quietly = TRUE))
     data("esophagus")
-    esophagus <- estimateDiversity(esophagus)
+    esophagus <- estimateDiversity(esophagus, threshold = 0.473)
```
antagomir, line 6: I propose that we use a specific default value for "threshold". This helps to keep it standardized for users who are not sure what to do with it and avoids the need to define it explicitly every time in the function call. A suitable default value is threshold = 0.9

R/estimateDiversity.R
```diff
@@ -104,31 +107,44 @@ setGeneric("estimateRichness",signature = c("x"),
...
+              }
+              #
+              dvrsts <- BiocParallel::bplapply(index,
+                                               .run_dvrsty,
+                                               x = x,
+                                               mat = assay(x, abund_values),
+                                               threshold = threshold,
+                                               BPPARAM = BPPARAM, ...)
+              .add_values_to_colData(x, dvrsts, name)
+          }
```
FelixErnst: I think just changed the indentation. Please revert this and at every other place

R/estimateDiversity.R
```diff
@@ -70,8 +73,8 @@ NULL
 setGeneric("estimateDiversity",signature = c("x"),
            function(x, abund_values = "counts",
                     index = c("shannon","simpson","inv_simpson", "richness",
-                              "chao1", "ACE"),
-                    name = index, ...)
+                              "chao1", "ACE", "coverage"),
+                    name = index, threshold = NULL, ...)
```
FelixErnst: I think you should pick up the optional argument `threshold` later on

R/estimateDiversity.R
```diff
@@ -199,15 +223,31 @@ setMethod("estimateRichness", signature = c(x = "SummarizedExperiment"),
     ans
 }
 
+.get_coverage <- function(x, threshold, ...){
```
FelixErnst: Set the default value for threshold here and remove it from the `setMethod` call. Should work without any other changes. Please describe the argument in the man page in any event

R/estimateDiversity.R
```diff
@@ -199,15 +223,31 @@ setMethod("estimateRichness", signature = c(x = "SummarizedExperiment"),
...
                          shannon = .get_shannon,
                          simpson = .get_simpson,
                          inv_simpson = .get_inverse_simpson,
                          richness = .get_observed,
                          chao1 = .get_chao1,
-                         ACE = .get_ACE)
-    dvrsty <- dvrsty_FUN(mat, ...)
+                         ACE = .get_ACE,
+                         coverage = .get_coverage)
+    dvrsty <- dvrsty_FUN(mat, threshold = threshold, ...)
```
FelixErnst: Remove the optional argument and let the `...` do their work

FelixErnst (changes requested):
Hi Tuomas,

thanks for the PR. Some changes are required. See the comments for details and let me know, if anything is unclear. Thanks!

Felix

microsud:
FYI: We have this open discussion regarding alpha diversity and how it should be structured  https://github.com/microbiome/MiaBook/pull/14
Not sure it should be done here but to keep it in mind.

R/estimateDiversity.R
TuomasBorman: image

I'm wondering, why the `threshold` is not passed to `.get_coverage`. I think it should be because `x` is passed there. 

In `.run_diversity`, it is still 0.5 like it should be, but it is not passed to forward. In `.get_coverage` `threshold `is the default value, 0.9.

FelixErnst: Because the argument is still named in the function definition of `.run_dvrsty` (line 251). If you remove it there, it will work. (and of couse from the lines 252-254)

If you `debug` this, you will noticed, that `list(...)` does not contain the threshold variable inside `.run_dvrsty`.

FelixErnst: You can have a look at a R specific explanation of the behavior of `...` here: https://adv-r.hadley.nz/functions.html?q=...#fun-dot-dot-dot

tests/testthat/test-0diversity.R
TuomasBorman: Threshold is 0.9

R/estimateDiversity.R
TuomasBorman: I copy pasted the content of file, and it seems that it wasn't smart move, because it messed the indentation. Now it is fixed.

R/estimateDiversity.R
TuomasBorman: Changed

R/estimateDiversity.R
TuomasBorman: Removed

R/estimateDiversity.R
TuomasBorman: Ok, thanks for the help. I think now it is working like it supposed to be

TuomasBorman: One thing, that causes an error. How that `threshold` should be described, because now it is an additional parameter, and `@param threshold `is not possible?

FelixErnst: I usually describe it in the `@param ...` section using `\itemize{\item{\code{threshold}{... description ...}}`. There are probably some examples in `mia` or `miaViz`

TuomasBorman: Yep, I found one example, which I used as a reference

R/estimateDiversity.R
```diff
@@ -19,9 +19,13 @@
 #' @param BPPARAM A
 #'   \code{\link[BiocParallel:BiocParallelParam-class]{BiocParallelParam}}
 #'   object specifying whether calculation of estimates should be parallelized.
-#'   (Currently not used)
 #'
-#' @param ... additional parameters passed to \code{estimateDiversity}
+#' @param threshold,... additional parameters passed to \code{estimateDiversity}
```
FelixErnst: You can remove the mention of `threshold` here as well as in `setMethod`.

The rational behind it, that this a specific argument for only one index.

TuomasBorman: I removed it

FelixErnst:
Thanks Tuomas

PR #76, opened 2021-03-12
Get rare taxa

TuomasBorman:
Hi!

I made a function that returns taxa whose abundance is under prevalence threshold. https://github.com/FelixErnst/mia/issues/59

Couple of things:

- I think there are no function that returns `SE` object with the most prevalent taxa. Should there be or is that `getRare `unnecessary?

- `getPrevalentTaxa` returns vector that contains names of most prevalent taxa. It has duplicates. Should there be duplicates? `getRareTaxa` does not have duplicates.

-Tuomas


FelixErnst:
Thanks for the PR. I think there are a few things to unpack.

1. return value: I think a character value as return value is more flexible. It can be used to subset a `SE` to rare taxa, so I think there is to much redundancy in `getRare`. Also the use in which no agglomeration should be performed is not implemented. I think it is best, for now, to keep it simple and just implement `getRareTaxa`. You can add a subsetting step in the examples.
2. duplicated or not: I am not sure, but IMO no. Any objections on this @microsud @antagomir? It should be changed then also for `getPrevelentTaxa`
3. `getRareTaxa` is the inverse of `getPrevalentTaxa`. This does not need to be wrapped in a function, at least not one so comprehensive. In my mind these are two or three lines of code along the line of

```r
prev_taxa <- getPrevalentTaxa(...)
taxa <- rowData(x)[,rank]
taxa[!(taxa %in% prev_taxa)]
```

Any input error gets caught by `getPrevalentTaxa` and the subsequent code just reuses `rank`, but that is sanitized, since it passed the checks in `getPrevalentTaxa`.

Therefore, I don't see the benefit of a seperate man page. Maybe just add it to the `getPrevalentTaxa` file and man page?

Technical aspects:
Indentation is different compared to the rest of the package. 4 space indentation is required.

antagomir:
Good to have this.

1. I agree that we could now keep it simple and only add `getRareTaxa` (and not `getRare`). We used to have this in microbiome pkg and I found it handy. But it is indeed just one-liner to do the subsetting, so perhaps better to keep things simple.
2. `getPrevalentTaxa` should not return duplicates as far as I can see; in what situation that occurs?
3. The `getRareTaxa` is the complement of `getPrevalentTaxa` and I agree with @FelixErnst about the implementation.

I could review the PR after these changes.

TuomasBorman:
1. Yep, it might be best to add `getRareTaxa` to same file as `getPrevalentTaxa`. I will make these changes.

2. Should the default value of `rank` in `getPrevalentTaxa` be `NULL`? Now the agglomeration is done by default. At least, I did not realize until now that it is possible to get taxa without agglomeration. Though it is said there in additional arguments, but which way is better?

3. I think there is no error with the function. There are duplicates because `rowData` is missing information. So this was false alarm.

Screenshot from 2021-03-14 10-30-48

antagomir:
2. Yes, default rank should be `NULL` (no aggregation). Then this should use the lowest available level (ie rows).
3. Ah ok. Checking this I noticed that `head(getPrevalentTaxa(GlobalPatterns, rank = "Species"))` returns taxa at various levels (Family, Genus, Species). Intuitively, this should only return prevalent Species? It this intended, or an error?
4. I noticed that the roxygen docs do not mention clearly what the default values for `detection` is for getPrevalentTaxa; perhaps we should mention this in the function documentation (@param detection field)?

TuomasBorman:
3. As seen also in the picture, it has those different levels also there. Those are agglomerated by `Species`, but their `Species `information is missing, so they get their lowest level available. Because `getPrevalentTaxa` uses directly that information, it also gets those different levels. 

I think user does not want information about different levels, when user want information about `Species`, so this should be changed. E.g. if there is no `Species` information, that taxa is not taken into account.

FelixErnst:
Do I understand it correctly, that you want to use the lowest non-empty rank by default?

If that's a yes, then consider implementing a `lowestAvailableRank` function. This is I think straightforward, since it requires just the combination of `taxonomyRanks` and `taxonomyRankEmpty`.

antagomir:
Yes, I agree with @tvborm - if the user aggregates to Species or other level, this should only return Species

Therefore we should **not** use the lowest non-empty rank by default. 

We can consider having an option to include the lowest available level, if the desired level is not available. But this should not be the default.

FelixErnst:
I don't follow. Is this the functions fault (with the current implementation) or is the problem of the dataset?

Remember: Garbage in, garbage out, not matter what.

antagomir:
In my view this is undesirable function behavior. It is very common in taxonomic profiling data sets that classifications for certain taxa are only available for Family or Phylum level, and not down to more detailed levels (Genera, Species). This missing information varies by taxonomic group, and it is often arising from missing classifications for certain sequence reads due to incomplete sequence databases. 

As a result, we have many OTUs/ASVs/sequence clusters in the data that are missing Species level classification. This is very common.

If the user wants to get a list of prevalent Species, then it would be expected that the function returns that data for Species, and not for all kind of other taxonomic groups. 

It could be an option to return the lowest available level, when the desired level was not available. But this should not be the default behavior.

microsud:
I agree. IMO `getPrevalence` or `getRare` should not have an agglomeration option. Instead it can be `getPrevalentTaxa(se, det, prev, returnTaxonomy=T/F)` . 
We can use just the IDs(rownames) for subsetting within and outside certain functions e.g. plotting, getTop, etc. `returnTaxonomy=T/F` is for the user to investigate taxonomy and maybe have interpretations.
Also `GlobalPatterns` is not ideal for this, it has too diverse samples. Instead, check for single ecosystems data like Sprockett el al.
Edit. by return taxonomy i mean joining rowData without any agglomeration. In this way the user can check which levels are assigned for each OTU/ASV.

antagomir:
It is indeed the case that agglomeration might be somewhat confusing. 

Agglomeration could be the option in `getPrevalentTaxa`, however. So that one can get the list of unique Species that we can detect to be rare.

We can leave out agglomeration from `getPrevalence` (and `getRare`). The user can choose to subset to only those rows that have species before calling this function if they like. We can add examples to highlight this.

I do not get what is the idea of the  `returnTaxonomy=T/F` option. The `SE` object in our case contains the taxonomy anyway (at least in most cases) and running `getPrevalence` should just keep it as is.

FelixErnst:
I am still not sure about the result you want to achieve or what is currently going wrong. To get this discussion on point, please show a sample input and the desired sample output. In addition, please summarise the connection between input and output as bullet point notes.

On the fact, that `getPrevalentTaxa` returns values formatted like `Family:...`, `Genus:...`: this is due to `na.rm` not being set `TRUE` in the agglomeration step. If `na.rm = TRUE` (see line 178-180 in getPrevalence.R), then all features with an `NA` on the selected would get removed and the result would contain not the taxonomic level prefixes. @tvborm Can you redo you example with the addition of `na.rm = TRUE`?

edit:
> In my view this is undesirable function behavior.

I don'T get what `this` is

antagomir:
The function `getPrevalence` should, by default, calculate the prevalence (i.e. population frequency) for each lowest level taxonomic group (row). 
- Input: (t)se object
- Output: a named numeric vector that gives the prevalence value for each taxonomic group; if the user provides `rank` argument, the data would be agglomerated to the specified rank (level) before calculating the prevalence

We often need a list of prevalent species or genera. A wrapper that would allow us to just pull that list out of a `TSE` object would be very handy. 

Another option is to implement `estimatePrevalence` that would add prevalence as a new field in colData. In that case it is not immediately clear how or whether the agglomeration should be implemented, and perhaps the best way would to demand that the user themselves agglomerates the data as they wish, before calling this function in the first place.

But this is not very handy for those who just like to quickly get quantification of prevalences at a specified taxonomic rank, for a given data.

I am not sure if we need `getRare` at all.

The function `getPrevalentTaxa` should return a vector of names for those taxonomic groups that are prevalent (above a given prevalence threshold and estimated with `getPrevalence`): 

- No taxonomic level specified -> return the lowest rank i.e. row ID for those rows that fulfil the criteria of being 'prevalent' (as specified in the function call); it could be an option to return a logical vector instead of row IDs.
- Taxonomic level specified -> agglomerate to the specified level, estimate prevalence, return names of the taxonomic groups (in this given rank) that are prevalent; these names should be unique after agglomeration and NA name/s should be dropped out
- It could be an option to add the lowest available rank if the specified level is not available but this should not be default and I am not sure how useful this would be

The function `getRareTaxa` is just a complement of `getPrevalentTaxa`.

TuomasBorman:
> @tvborm Can you redo you example with the addition of na.rm = TRUE

I made the addition. 

With addition, `getPrevalentTaxa ` is working correctly. If `Species` levels is NA, that taxa is not returned. 
In the picture I first used the current implementation and took only `Species `from it. Then I made the addition, and as seen in the picture `Species` before equals to `Species` after the addition. 

image

FelixErnst:
Thanks for the summary.

Which part does not work? `getPrevalence` returns a named numeric vector and`getPrevelantTaxa ` returns the names as a character vector for the settings specified.

In addition the simplified implementation of `getRareTaxa` is not in the repo at the moment. I am not 100% sure, whether this has something to do with the output or not. I also am not sure, what changed for the results in `before` and `after`, since the function calls are identical.
@tvborm can you push the recent changes you made? Thanks!

Did you try setting `na.rm = TRUE` for `getPrevalentTaxa`?

antagomir:
As far as I can see this is ok now (`getRareTaxa` could be added). I will test more once the final version is in place.

antagomir (approved):

TuomasBorman:
Yep, I planned to add the `getRareTaxa` to `getPrevalence.R` and delete those additional files

microsud:
Sorry if I am complicating things here, but I have queries.
```
library(mia)
library(microbiomeDataSets)
library(dplyr)
tse <- microbiomeDataSets::SprockettTHData()
#tse <- microbiomeDataSets::SilvermanAGutData() # prevalence function does not work on this dataset??

# just anonymize rownames as usually they will not be taxonomicaly named 
rownames(tse) <- paste0("ASV-", seq(1:nrow(tse)))
```

**Get prevalence estimates for individual ASV/OTU**
`getPrevalence` returns both id/name and prevalence
This also returns ASVs with 0.0000000000 prevalence values! 
```
prevs.asv <- getPrevalence(tse,
                           detection = 0.1,
                           prevalence = 0.5,
                           sort = TRUE,
                           as_relative = TRUE)

prevs.asv[1:5]

#output 
  ASV-1370   ASV-2302   ASV-1362   ASV-1770    ASV-978 
0.32933845 0.19367210 0.10450623 0.08916587 0.08581016 
```

The user can get taxonomic classification to make sense of it.
```
prevs.asv.taxo <- rowData(tse) %>% as.data.frame() 
prevs.asv.taxo <- prevs.asv.taxo[names(prevs.asv[1:5]),]
head(prevs.asv.taxo)

# output 
          Kingdom         Phylum               Class             Order             Family                Genus
ASV-1370 Bacteria Actinobacteria      Actinobacteria Bifidobacteriales Bifidobacteriaceae      Bifidobacterium
ASV-2302 Bacteria Proteobacteria Gammaproteobacteria Enterobacteriales Enterobacteriaceae Escherichia/Shigella
ASV-1362 Bacteria Actinobacteria      Actinobacteria Bifidobacteriales Bifidobacteriaceae      Bifidobacterium
ASV-1770 Bacteria  Bacteroidetes         Bacteroidia     Bacteroidales     Prevotellaceae         Prevotella_9
ASV-978  Bacteria     Firmicutes             Bacilli   Lactobacillales   Streptococcaceae        Streptococcus
                                                       Species                            lowest_rank
ASV-1370                   breve/kashiwanohense/longum/saguini Bifidobacterium_breve/kashiwanohense+2
ASV-2302                                                  <NA>                   Escherichia/Shigella
ASV-1362 catenulatum/gallicum/kashiwanohense/pseudocatenulatum Bifidobacterium_catenulatum/gallicum+2
ASV-1770                                                 copri                     Prevotella_9_copri
ASV-978                                                   <NA>                          Streptococcus
```

 `getPrevalentTaxa`  is expected to return names. 
This does not return any values, I was expecting a return of only ASVids
```
taxa <- getPrevalentTaxa(tse,
                         rank= NULL, # when null
                         detection = 0.1,
                         prevalence = 0.5,
                         as_relative = TRUE)

taxa
#output
character(0)
```

When agglomerated, this returns values mixed genus and family name 

```
taxa <- getPrevalentTaxa(tse,
                         rank= "Genus",
                         detection = 0.01,
                         prevalence = 0.2,
                         as_relative = TRUE)

taxa

#output
 [1] "Family:Lachnospiraceae"            "Genus:Dorea"                       "Genus:Blautia"                     "Genus:Faecalibacterium"           
 [5] "Genus:Subdoligranulum"             "Genus:Clostridium_sensu_stricto_1" "Genus:Streptococcus"               "Genus:Lactobacillus"              
 [9] "Genus:Veillonella"                 "Genus:Dialister"                   "Genus:Bifidobacterium"             "Genus:Prevotella_9"               
[13] "Genus:Bacteroides"                 "Genus:Escherichia/Shigella" 
```

Is this ideal? At this point, ASVs are agglomerated and then prevalence is calculated. This means at the genus level you get a certain list. 
For example, we have 2 ASVs classified at Genus level Blautia
Prevalence of ASV1 is 50%, ASV2 is 20% (assuming present in diff samples) and so on, when these are agglomerate, these get overall higher prevalence (70%) at Genus level.
Now for 5 ASVs unclassified at genus level but at family level `Lachnospiraceae`, How do we calculate prevalence? If each ASV is maintained as unique then they will have lower prevalences. If agglomerated and then calculated, they may have higher prevalence because you club many more ASVs together?

With the current update having `na.rm = TRUE` the user needs to be aware that potentially several unclassified genera (that could have high prevalence) have been discarded and that the prevalence estimate is only for those classified until the genus level. Needs to be in rdfiles. So the user cannot directly claim genus XYZ are most prevalent in their data.

FelixErnst:
> This also returns ASVs with 0.0000000000 prevalence values!

Well it says so on the function name. It returns the prevalence. And if the prevalence is 0, than it is 0, isn't it?

> ```
> prevs.asv.taxo <- rowData(tse) %>% as.data.frame() 
> prevs.asv.taxo <- prevs.asv.taxo[names(prevs.asv[1:5]),]
> head(prevs.asv.taxo)
> ```

Why this aversion against `DataFrame`s?

This works as well:
```
prevs.asv.taxo <- rowData(tse)[names(prevs.asv[1:5]),]
prevs.asv.taxo 
```

> getPrevalentTaxa is expected to return names.
> This does not return any values, I was expecting a return of only ASVids

Well please read the man page:
>prevalence: Prevalence threshold (in 0 to 1). The required prevalence is strictly greater by default. To include the limit, set include_lowest to TRUE.
>...
>getPrevalentTaxa returns taxa that are more prevalent with the given detection threshold for the selected taxonomic rank.

no taxa has a prevalence greater or even than `0.5`. So the result is empty. 

```r
> any(prevs.asv>0.5)
[1] FALSE
```

My result differ from yours, since `na.rm = TRUE` is now set by default
```r
> taxa <- getPrevalentTaxa(tse,
                         rank= "Genus",
                         detection = 0.01,
                         prevalence = 0.2,
                         as_relative = TRUE)
> taxa
 [1] "Dorea"                       "Blautia"                     "Faecalibacterium"            "Subdoligranulum"             "Clostridium_sensu_stricto_1"
 [6] "Streptococcus"               "Lactobacillus"               "Veillonella"                 "Dialister"                   "Bifidobacterium"            
[11] "Prevotella_9"                "Bacteroides"                 "Escherichia/Shigella"       
```

When setting it back to `FALSE` the results match:

```r
> taxa <- getPrevalentTaxa(tse,
                           rank= "Genus",
                           detection = 0.01,
                           prevalence = 0.2,
                           as_relative = TRUE,
                           na.rm = FALSE)
> taxa
 [1] "Family:Lachnospiraceae"            "Genus:Dorea"                       "Genus:Blautia"                     "Genus:Faecalibacterium"           
 [5] "Genus:Subdoligranulum"             "Genus:Clostridium_sensu_stricto_1" "Genus:Streptococcus"               "Genus:Lactobacillus"              
 [9] "Genus:Veillonella"                 "Genus:Dialister"                   "Genus:Bifidobacterium"             "Genus:Prevotella_9"               
[13] "Genus:Bacteroides"                 "Genus:Escherichia/Shigella"  
```

> For example, we have 2 ASVs classified at Genus level Blautia
Prevalence of ASV1 is 50%, ASV2 is 20% (assuming present in diff samples) and so on, when these are agglomerate, these get overall higher prevalence (70%) at Genus level.
Now for 5 ASVs unclassified at genus level but at family level Lachnospiraceae, How do we calculate prevalence? If each ASV is maintained as unique then they will have lower prevalences. If agglomerated and then calculated, they may have higher prevalence because you club many more ASVs together?

Thats the beauty of the function. It can be used either way. The interpretation is also up the user, since we don't know every dataset in the world. Do `getPrevelence` and `getPrevalentTaxa` work as expect? I would say, yes. Can we may be explain it better in the man page? Certainly.

So if it I read this whole thread again, two or three things get thrown into the same pot, which certainly helps the discussion in connection with the broader picture, but does make it easier to find a solution and a conclusion.

Therefore lets get this back on topic:
- does `getPrevalentTaxa` performs as intended? IMO, yes
- does `getRareTaxa` needs to be aadded to `getPrevalentTaxa` man page? IMO, yes, in order to make the description easier and centralize it in one place
- do we need `getRare`? IMO, no, since it is a fancy wrapper aroung `[`

Do we agree on these three points?

microsud:
> Why this aversion against `DataFrame`s?

No detailed experience/understanding of DataFrame as you have so kept it easier from general data.frame for example.

> no taxa has a prevalence greater or even than `0.5`. So the result is empty.

My bad did not change the cut-offs like for genus level

> no taxa has a prevalence greater or even than `0.5`. So the result is empty.

Should there be a message/warning when the result is empty?

> Do we agree on these three points?

I am now clear about these

antagomir:
I also agree about the points. 

I do not see a need for warning. The output of the function is rather logical: it is empty when there are no prevalent taxa. As expected.

FelixErnst:
@tvborm Can you implement the changes as discussed for `getRareTaxa`? (See my first comment)

TuomasBorman:
I implemented most of it. Though, there are still some issues that are related to detection and prevalence thresholds

TuomasBorman:
Or about the detection limit...

If user wants rare taxa, does the user want

1. taxa that are below prevalence threshold BUT above the detection limit
2. or all the taxa that are not included to prevalent taxa

antagomir:
We want the complement of prevalent taxa. Otherwise this gets too complicated concept to grasp.

PR #79, opened 2021-03-13
Added estimateRichness as its own function

antagomir:
I splitted richness indices out from `estimateDiversity` into their own function `estimateRichness` and added unit tests.

Will review all alpha diversity and related indices (richness, evenness, diversity, dominance, rarity) soon, with particular attention to getting all Simpson's right. 

The richness indices ACE and Chao1 generate also standard error information, generating two columns instead of one. This is mentioned in the documentation.

Will also check that the coding style is uniform across these. Noticed some differences.

@tvborm are you working on `estimateRarity`?


R/estimateRichness.R
```diff
@@ -0,0 +1,300 @@
...
+
+    richness <- DataFrame(richness)
+
+    colnames(richness) <- names
+
+    colData(x)[,names] <- richness
+
+    x
+
+}
```
FelixErnst: This function is a bit redundant, since it works in the current implementation as well. Please reuse `.add_values_to_colData`

R/estimateRichness.R
```diff
@@ -0,0 +1,300 @@
...
+                ACE = .get_ACE,
+                Hill = .get_hill,
+                .calc_richness)
+
+    do.call(FUN,
+            list(x = assay,
+            detection = detection,
+            index = index
+        ))
+}
```
FelixErnst: This should not work. The dispatch with `do.call` only contains named arguments, which do not match the expectations for all functions available, for example: `.calc_richness` does not have an argument `x`.

Please streamline all these functions to have the same argument name(s). (I would prefer `mat` instead of `x`). I also would like to have all switch option explicitly named. 

Small hint. the default call to `.calc_richness` is not even available, since the default values in line 189 only contain the explicitly named options.

R/estimateRichness.R
```diff
@@ -0,0 +1,300 @@
...
+                        j = seq_len(ncol(mat)),
+                        MoreArgs = list(x = mat),
+                        SIMPLIFY = FALSE),
+                        sum)
+    ans <- unlist(ans)
+
+    # Adds sample names to the table
+    names(ans) <- colnames(mat)
+    ans
+}
```
FelixErnst: the index argument is not used and can be removed. It also doesn't really make sense in context of the subsequent function

tests/testthat/test-5richness.R
```diff
@@ -0,0 +1,25 @@
...
+
+    esophagus <- estimateRichness(esophagus, detection = 1)
+    cd <- colData(esophagus)
+    expect_equal(unname(round(cd$observed, 0)), c(15, 24, 16))
+
+    # These are unaffected by detection parameter
+    expect_equal(unname(round(cd$Chao1, 4)), c(39.1429, 37.5000, 71.0000))
+    expect_equal(unname(round(cd$ACE, 4)), c(49.0970, 40.9465, 88.9768))
+    expect_equal(unname(round(cd$Hill, 4)), c(9.4817, 15.8376, 7.6331))
+
```
FelixErnst, line 71: Test for the calculated values of `.calc_richness` is missing

FelixErnst (changes requested):
Thanks for the PR.

I think this is a good restructuring and makes conceptually sense.

Small changes required. See comments.

R/estimateRichness.R
antagomir: Removed .calc_richess entirely.

antagomir: Switched from x to mat in the functions that execute final richness calculations.

tests/testthat/test-5richness.R
antagomir: Removed .calc_richess entirely.

antagomir:
Tried to fix the earlier issues.

Unified the coding conventions little bit across `estimateDiversity`, `estimateRichness`, `estimateEvenness`, `estimateDominance`. However, there are still differences. We could see if one coding style should be preferred over others. Too many possible things to unify to list here but perhaps we can have a look and discuss.

The richness indices are now case-sensitive (e.g. ACE, Chao1) but we have not required this from other alpha diversity indices (Simpson, Shannon..). It might be useful for nice default plotting etc. to enforce the standard spelling forms but on the other hand, accepting any matching cases, or just using lowercase for all, could reduce mental load and make life easier for the user (no need to remember or check how to capitalize each index). I propose that we anyway do this in the same way for all these functions. One option is to recommend a best practice in examples, so that user would always provide also the argument "name" ie `estimateDiversity(tse, index = "shannon", name = "Shannon")`. In that case we could keep all technical index names lowercase.

Also checked and fixed all Simpson indices. All should be fine now. Do not mix vegan::diversity(.., index = "simpson") with Simpson index (which is ambiguous concept).. the vegan "simpson" index is 1-lambda and invsimpson is 1/lambda.

FelixErnst:
> We could see if one coding style should be preferred over others. Too many possible things to unify to list here but perhaps we can have a look and discuss.

Can you elaborate on this? Internal or external coding styles?

> The richness indices are now case-sensitive (e.g. ACE, Chao1) but we have not required this from other alpha diversity indices (Simpson, Shannon..).

I think, they are always case-sensitive, since `match.arg` and `switch` are case-sensitive

>It might be useful for nice default plotting etc. to enforce the standard spelling forms but on the other hand, accepting any matching cases, or just using lowercase for all, could reduce mental load and make life easier for the user (no need to remember or check how to capitalize each index).

I think this is not necessarily a requirement. You don't want to force the user to use a specific name to be able to use a plotting function. Therefore, the name of a variable used for potting is always one of the arguments for the plotting function. In addition, the responsibility to perform an analysis is in the hand of the user, so it should be required to pick the right name of variable to be used.

> Also checked and fixed all Simpson indices. All should be fine now. Do not mix vegan::diversity(.., index = "simpson") with Simpson index (which is ambiguous concept).. the vegan "simpson" index is 1-lambda and invsimpson is 1/lambda.

Thanks. The references you added are a life saver I think

antagomir:
> We could see if one coding style should be preferred over others. Too many possible things to unify to list here but perhaps we can have a look and discuss.

-> Internal coding style.

----

Re: index naming - yes. I will add examples of good practice where name argument is being actively used. 

I would change all indices lowercase, so that the users do not need to remember the widely varying capitalizations for various indices.

FelixErnst (approved):

FelixErnst:
If you want to renamed the indices, please go ahead. Let me know, when I should merge

antagomir:
For `estimateDiversity`  we have separate functions for specific indices (`estimateShannon`, `estimateInvSimpson`..).

We don't have these for `estimateRichness`, `estimateEvenness`, `estimateDominance` and I don't see a need for separate functions for specific indices.

For consistency, I will remove these from diversity estimation as well. The diversities are then always obtained through the generic `estimateDiversity`.

antagomir:
Ok for merge now unless there are further comments.

PR #83, opened 2021-03-17
Some updates to documentation and coding style unification

antagomir:


R/estimateDiversity.R
```diff
@@ -102,27 +102,30 @@
 #' se <- GlobalPatterns
 #'
 #' # All index names as known by the function
-#' index_code <- c("shannon","gini_simpson","inverse_simpson", "coverage", "fisher")
+#' index <- c("shannon","gini_simpson","inverse_simpson", "coverage", "fisher")
+#'
+#' # Corresponding polished names
+#' name <- c("Shannon","GiniSimpson","InverseSimpson", "Coverage", "Fisher")
```
FelixErnst, line 108: Do you want to use in the later call to `estimateDiversity`?

R/estimateDominance.R
```diff
@@ -342,18 +342,21 @@ setMethod("estimateDominance", signature = c(x = "SummarizedExperiment"),
...
-                 aggregate = aggregate,
-                 index = index))
+
+    FUN(index, mat = mat, ntaxa = ntaxa, aggregate = aggregate, ...)
+    
+    #do.call(FUN,
+    #        list(mat = assay,
+    #             ntaxa = ntaxa,
+    #             aggregate = aggregate,
+    #             index = index))
```
FelixErnst: Removeable?

R/estimateDominance.R
```diff
@@ -342,18 +342,21 @@ setMethod("estimateDominance", signature = c(x = "SummarizedExperiment"),
...
 }
 
-.get_dominance_values <- function(index, assay, ntaxa = 1, aggregate = TRUE) {
+.get_dominance_values <- function(index, mat, ntaxa = 1, aggregate = TRUE, ...) {
 
     FUN <- switch(index,
                   simpson_lambda = .simpson_lambda,
                   core_abundance = .calc_core_dominance,
                   gini = .calc_gini_dominance,
                   .calc_dominance)
```
FelixErnst: Named option?

tests/testthat/test-5dominance.R
```diff
@@ -97,4 +97,5 @@ test_that("estimateDominance", {
     tse <- esophagus
     assay(tse,"counts") <- DelayedArray(assay(tse,"counts"))
     test_internal_estimateDominance(tse)
+
```
FelixErnst: Empty line?

R/estimateDiversity.R
antagomir: The idea was to give a clear example on how the user can provide both index and name at one go, side by side. They are both used later in the example.

R/estimateDominance.R
antagomir: Yes.

R/estimateDominance.R
antagomir: Do you refer to .calc_dominance here? This is a general option related to dbp, dmn, absolute and relative. They are all closely related with some differing parameterizations but also separately known by their specific names.

FelixErnst: The problem is, that if `index` is `whatever`, an index is calculate as well. This might be just available from internal point of view, but also, when going through the code, it does not make sense without prior knowledge. Therefore, I would suggest explicitly naming each option.

antagomir: right. will do.

antagomir:
The issues fixed.

Issue #82, opened 2021-03-17
getRareTaxa duplicates


antagomir:
The `getPrevalentTaxa` and `getRareTaxa` should work the same way but they do not.

- The values returned by `getRareTaxa` are not unique (should be, as in `getPrevalentTaxa`)
- The values should be complement (no overlap) ie rare should be all taxa minus prevalent taxa; there is overlap however

`
prev <- getPrevalentTaxa(GlobalPatterns,rank = "Phylum")
head(prev)
`
[1] "Crenarchaeota"  "Euryarchaeota"  "Actinobacteria" "Spirochaetes"  
[5] "Proteobacteria" "SBR1093"       
> 

`
rare <- getRareTaxa(GlobalPatterns,rank = "Phylum")
head(rare)
`

[1] "Crenarchaeota" "Crenarchaeota" "Crenarchaeota" "Crenarchaeota"
[5] "Crenarchaeota" "Crenarchaeota"

FelixErnst:
25575d0b85f5451c7ac0559dc04dc6bea6232bb8

PR #84, opened 2021-03-17
getRareTaxa fix

TuomasBorman:
Hi!

Regarding to the issue https://github.com/microbiome/mia/issues/82

Leo, thanks for pointing that out. 

There was an error in getRareTaxa, First of all, function did not return right "version" of taxa variable, modifications was not stored in the returned variable. 

Secondly, there were duplicates, because getRareTaxa just took the complement without dropping duplicates. Now that is also corrected. 

-Tuomas

### DIFF
--- a/R/getPrevalence.R
+++ b/R/getPrevalence.R
@@ -278,6 +278,9 @@ setMethod("getPrevalentTaxa", signature = c(x = "SummarizedExperiment"),
 
 #' @rdname getPrevalence
 #'
+#' @details
+#' \code{getRareTaxa} returns complement of \code{getPrevalentTaxa}.
+#'
 #' @export
 setGeneric("getRareTaxa", signature = "x",
            function(x, rank = NULL, ...)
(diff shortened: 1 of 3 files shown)


antagomir:
I noticed that `getRareTaxa` does not have `prevalence` argument set by default whereas `getPrevalentTaxa` has. These two are complements, so it would be logical to have the same prevalence argument in both by default. 

Unit tests could then additionally verify that they are indeed complement, so that there is not overlap; and that the union corresponds to all taxa.

Could you have one more look @tvborm

FelixErnst:
All arguments get passed to `getPrevalentTaxa` in `getRareTaxa`. From my point of view, just the man page needs to be adjusted, since the implementation in itself ensures the non-overlap,

PS: Beaware, the I simplified the internals yesterday and added another example to the man page

antagomir:
Ok I think it is ok now.

PR #86, opened 2021-03-24
Pd to diversity

### LINKED ISSUE #81: Add Faith's phylogenetic alpha diversity
Add implementation of Faith's phylogenetic alpha diversity in `estimateDiversity`. 

This is tree-based alpha diversity measure and would nicely complement the other (non-tree) based ones.

The `picante::pd()` provides one implementation (GPL2). It would add a dependency here, or we could add our own implementation to reduce dependencies. This measure is not overly complex.

TuomasBorman:
Hi!

I implemented Faith's phylogenetic distance. I think there are still some work to do with the structure and also with the functionality. E.g., this implementation does not take into the account if the tree is rooted or not. 

Now SE and TSE objects have different method.

Issue: https://github.com/microbiome/mia/issues/81

-Tuomas


R/estimateDiversity.R
```diff
@@ -118,8 +127,8 @@
 #'
 #' # It is recommended to specify also the final names used in the output.
 #' se <- estimateDiversity(se,
-#'   index = c("shannon", "gini_simpson", "inverse_simpson", "coverage", "fisher"),
-#'    name = c("Shannon", "GiniSimpson",  "InverseSimpson",  "Coverage", "Fisher"))
+#'   index = c("shannon", "gini_simpson", "inverse_simpson", "coverage", "fisher", "pd"),
+#'    name = c("Shannon", "GiniSimpson",  "InverseSimpson",  "Coverage", "Fisher", "Pd"))
```
antagomir: I would propose naming this faith/Faith - more specific than "pd". There can also be other "phylogenetic diversity" measures (now or in the future).

R/estimateDiversity.R
```diff
@@ -149,23 +158,30 @@ NULL
 
 #' @rdname estimateDiversity
 #' @export
-setGeneric("estimateDiversity",signature = c("x"),
+setGeneric("estimateDiversity",signature = c("x", "tree"),
            function(x, abund_values = "counts",
                     index = c("shannon","gini_simpson", "inverse_simpson",
-                              "coverage", "fisher"),
-                    name = index, ...)
+                              "coverage", "fisher", "pd"),
```
FelixErnst: The default setting in the generic will overwrite the default in setMethod AFAIK.

So I think, it would be good to keep this as a separate function with its own generic for now. (only for an SE not for TSE)

R/estimateDiversity.R
```diff
@@ -149,23 +158,30 @@ NULL
...
-             name = index, ..., BPPARAM = SerialParam()){
+             name = index, tree = "missing", ..., BPPARAM = SerialParam()){
 
         # input check
+        # IF object does not have a tree
+        if( "pd" %in% index ){
+            stop("SummarizedExperiment object does not have a tree. 'PD' is not possible to calculate.",
+                 call. = FALSE)
+        }
+
```
FelixErnst: On the one hand you set the default and on the other hand you don't allow it. The user cannot run the function with the default setting, even though a SE is provided.

This can only be the one or the other and you have to choose. See comment above

antagomir (commented):
Seems good to me. We should aim to ensure that our own implementation does not have notable shortcomings compared to the implementation in the other package, and of course validate by checking that same results are obtained.

R/estimateDiversity.R
```diff
@@ -239,17 +294,70 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
 
+.calc_pd <- function(mat, tree, ...){
+
+    # Gets name of the taxa
+    taxa <- rownames(mat)
+    # Gets name of the samples
+    samples <- colnames(mat)
+
+    # Loops through every sample
+    PDs <- BiocParallel::bplapply(samples, function(sample){
```
FelixErnst: You have to add the default `BPPARAM = SerialParam()` argument to this function, because the otherwise the system default will be used aka. parallelised by default. See other functions for example implementation, `runJSD` or `runUniFrac` for example

R/estimateDiversity.R
```diff
@@ -239,17 +294,70 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+
+        # Stores taxa that are present in the sample, i.e., abundance is over 0
+        present <- taxa[ mat[, sample] > 0 ]
```
FelixErnst: I think this can be precalculated outside the `lapply`. `mat` and `samples` have the same length on the column and `mat` and `taxa` on the rows

R/estimateDiversity.R
```diff
@@ -239,17 +294,70 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+        # Stores taxa that are not present, i.e., those taxa that are not in present taxa
+        absent <- taxa[ which( !(taxa %in% present) ) ]
```
FelixErnst: Same as above

R/estimateDiversity.R
```diff
@@ -239,17 +294,70 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+
+        # If no taxa are preent, then PD = 0
+        if( length(present) == 0 ){
+            PD <- 0
+        }
```
FelixErnst: Same as above

R/estimateDiversity.R
antagomir: @FelixErnst do you mean that we should have different exported functions for alpha diversities that use the tree (TSE) and those that do not require the tree (SE)? Or is this discussion about the internal functions and implementation but user could still call all these functions through `estimateDiversity`?

R/estimateDiversity.R
antagomir: +1

R/estimateDiversity.R
```diff
@@ -239,17 +294,70 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+
+        # If there are taxa that are not present
+        else{
+            # Absent taxa are dropped from the tree
+            sub_tree <- ape::drop.tip(tree, absent)
+            # PD is now calculated based on the sub tree
+            PD <- sum(sub_tree$edge.length)
+        }
+    })
```
FelixErnst: I think there is a chance that you can avoid the `lapply` and work with vectorised subsetting. Maybe like this (untested):

```r
PD <- rep(NA,length(samples))
PD[length(present) == 0] <- 0
f <- length(present) == 1
# the match has to be reworked, since the length of the elements probably differ.
# Maybe this need to put in a `lapply`
PD[f] <- tree$ages[which(tree$edge[, 2] == which(tree$tip.label == present[f]))] 
PD[length(absent) == 0] <- sum(tree$edge.length)
f <- length(absent) == 1
trees <- lapply(absent[f], ape::drop.tip, phy = tree)
PD[f] <- vapply(trees, function(t){sum(t$edge_length)},numeric(1))
```

FelixErnst (changes requested):
Hi Tuomas,

thanks for the PR.

There needs to be some discussion on the conceptual side. Also the implementation can probably be improved. For the latter we require tests to be added early on to always make sure that the implementation is correct.

R/estimateDiversity.R
FelixErnst: Maybe that's the right direction to head in. However, I had something different in mind:

- `estimateDiversity` for `SE` without any mention of `pd`
- `estimatePD` for `SE` and `tree`
- `estimateDiversity` for `TSE` dynamically including `pd` if `rowTree` is not `NULL`

With this the contracts from the objects slots are resolved stringently without including any ambiguous combination of defaults and ability to calculate certain indices.

For even more clarity and easier implementation `pd` can also be omitted for `TSE` by default. This could then be easily implemented like this 

```r
setGeneric("estimatePD", ...)
setMethod("estimatePD", signature = c("SummarizedExperiment","phylo"),
                   ...)
setMethod("estimatePD", signature = c("TreeSummarizedExperiment","missing"),
    function(x, ...){
        if(is.null(rowTree(x))){
            stop("...")
        }
        estimatePD(x, rowTree(x), ...)
    }
)
```

This would then be a separation of diversity functions using trees

R/estimateDiversity.R
TuomasBorman: I have been trying to make this work with this approach. The problem is that, I want to get those taxa that are present in specific sample (e.g. vector that contains taxa for sample 1, taxa for sample 2...), and I haven't been able to do that. 

It is easy to get all the taxa that are present, but then it is just one big vector, so I lost the sample information (which taxa belongs to which sample). 

I can get list of samples, that include TRUE/FALSE if taxa is present, but then it is not possible to subset without using lapply, and then I do not know what is the benefit compared to the first approach. 

So, I'm little bit stuck with this

R/estimateDiversity.R
TuomasBorman: I am not sure about my estimatePD either. Currently, first all but "pd" are added to colData, and after that "pd" is added. For example, if those indices are already calculated, then 2 warnings come because first warning says "shannon, simpson... already in colData", and the second warning says "pd is already in colData". 

Those 2 warnings might be weird to user, so should that be done differently? 

And should estimatePD be just normal internal function?

FelixErnst: Then refactor the code accordingly. These warnings are a new addition, so I guess there needs to be additional maturation

R/estimateDiversity.R
FelixErnst: Use `split` with the `lengths`/`sum` (aka. the number) of taxa found. Should yield a `CharacterList` and allow you to work with that

FelixErnst: Maybe also have a look at `PartitioningByEnd`/`PartitioningByWidth` and the rational behind that. `relist` is also related to that and `split`

TuomasBorman: Ok, thanks for help

R/estimateDiversity.R
TuomasBorman: Changed it to faith

R/estimateDiversity.R
TuomasBorman: Structure changed accordingly

R/estimateDiversity.R
TuomasBorman: Structure changed accordingly

R/estimateDiversity.R
TuomasBorman: Implemented this

R/estimateDiversity.R
TuomasBorman: Structure changed accordingly

TuomasBorman:
There is an error caused by `warning: package 'mia' is not available for this version of R`. Because of the same error, I can't do tests in my computer either. I remember that I asked about similar thing some time ago, but I have forgotten it and unfortunately I did not put that info anywhere. There was some trick but I can't remember what

FelixErnst:
you are using R 4.0 and mia is now requires R4.1 due to the bioc submission. You have two options

1. Clone the repo, switch the dependency in DESCRIPTION back to R 4.0 and build the package. This should work
2. Use docker to spin up a devel environment in a browser (See example here: https://github.com/Bioconductor/bioconductor_docker)
3. Install R-devel version

FelixErnst:
The current errors on GitHub Actions are thrown because the internal function `.check_abund_values` got renamed. It is now `.check_assay_present`

R/estimateDiversity.R
```diff
@@ -239,17 +344,66 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+    absent_combined <- taxa[ mat[, samples] == 0 ]
+
+    # colSums(mat > 0) # Gets how many taxa there are in each sample
+    # as.vector(cumsum(colSums(mat > 0)+1)) # Determines indices of samples' first taxa
+    # seq_along(present_combined) %in% as.vector(cumsum(colSums(mat > 0)+1)) # Gets true if splitting point
+    # as.factor(cumsum((seq_along(present_combined)-1) %in% as.vector(cumsum(colSums(mat > 0)))))) # Determines which taxa belongs to which sample
+    # Split() assigns taxa to right samples based on their number that they got from previous step
+    # unname() deletes unnecessary names
+    present <- unname(split(present_combined, as.factor(cumsum((seq_along(present_combined)-1) %in% as.vector(cumsum(colSums(mat > 0)))))))
+    absent <- unname(split(absent_combined, as.factor(cumsum((seq_along(absent_combined)-1) %in% as.vector(cumsum(colSums(mat == 0)))))))
```
FelixErnst: For clarity, it is always better split the factor generation from actually using it. Also the line length does not exceed 80/100 characters:

```r
split_present <- as.vector(cumsum(colSums(mat > 0)))
split_absent <- as.vector(cumsum(colSums(mat == 0)))
split_present <- as.factor(cumsum((seq_along(present_combined)-1) %in% split_present))
split_absent <- as.factor(cumsum((seq_along(absent_combined)-1) %in% split_absent))
present <- unname(split(present_combined, split_present))
absent <- unname(split(absent_combined, split_absent))
```

R/estimateDiversity.R
```diff
@@ -178,13 +188,108 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+        }
+
+        # If index list contained other than 'faith' index, the length of the list is over 0
+        if( length(index)>0){
+            # Calculates all indices but not 'faith'
+            x <- callNextMethod()
+        }
+
+        # If 'faith_name' exist, then 'faith' index was present in the index list
+        if( exists("faith_name") ){
```
FelixErnst: The `exists` function is a nice feature, but it is not the right solution in this instance. From the meaning of the check you want to answer the question "Do we need to calculate the faith index?". This can be answered with a "Yes" or "No" aka. boolean. Thus, it should be a scalar logical variable defining the answer to this question.

```r
calc_faith <- FALSE
if(...){
  ...
  calc_faith <- TRUE
}
...
if(calc_faith){
  ...
}
```

Also `exists` can walk environments, which leads to this behavior.

```
> FUN <- function(){
+     exists("abc")
+ }
> FUN()
[1] FALSE
> abc <- "meep"
> FUN()
[1] TRUE
```

If the user defines a variable `faith_name` in the global environment, the index will always be tried to calculated. Thus, the `exists` function is dangerous, especially if you don't define the `where` and/or `envir` argument. I would be very very careful with this function.

R/estimateDiversity.R
```diff
@@ -178,13 +188,108 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+#' @export
+setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
+    function(x, abund_values = "counts",
+             index = c("shannon","gini_simpson","inverse_simpson",
+                       "coverage", "fisher", "faith"),
+             name = index, tree = rowTree(x), ..., BPPARAM = SerialParam()){
+
+        # input check
+        # IF object does not have a tree
+        if( ("faith" %in% index) && (is.null(tree) || is.null(tree$edge.length)) ){
```
FelixErnst, line 213: `tree` is not defined

R/estimateDiversity.R
TuomasBorman: Good point. Much easier to follow the code, if user wants to

Changed

R/estimateDiversity.R
TuomasBorman: Changed

R/estimateDiversity.R
FelixErnst: And for yourself in two years 😄

TuomasBorman: Well that's also true :D

R/estimateDiversity.R
TuomasBorman: Added

R/estimateDiversity.R
```diff
@@ -178,13 +188,120 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+#' 
+#' @param tree A phylogenetic tree that is used to calculate 'faith' index.
+#' By default, it is \code{rowTree(x)}.
+#' 
+#' @export
+setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
+    function(x, abund_values = "counts",
+             index = c("shannon","gini_simpson","inverse_simpson",
+                       "coverage", "fisher", "faith"),
+             name = index, tree = rowTree(x), ..., BPPARAM = SerialParam()){
```
FelixErnst: Since you supply a potential tree with `x`, I think you can drop the tree argument here. Just add:

```r
tree <- rowTree(x)
```

directly in the function. Otherwise the user could overwrite the tree and that doesn't make sense

TuomasBorman: Removed

R/estimateDiversity.R
```diff
@@ -178,13 +188,123 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+)
+
+#' @rdname estimateDiversity
+#' 
+#' @param tree A phylogenetic tree that is used to calculate 'faith' index.
+#' By default, it is \code{rowTree(x)}.
+#' 
+#' @export
+setGeneric("estimateFaith",signature = c("x", "tree"),
+           function(x, tree = "missing", abund_values = "counts",
```
FelixErnst, line 259: The missingness cannot be defined in the generic.

Currently this defines a character vector with the value `missing` as default.

R/estimateDiversity.R
```diff
@@ -178,13 +188,123 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+#' @rdname estimateDiversity
+#' @export
+setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
+    function(x, abund_values = "counts",
+             index = c("shannon","gini_simpson","inverse_simpson",
+                       "coverage", "fisher", "faith"),
+             name = index, ..., BPPARAM = SerialParam()){
+        
+        # Gets the tree 
+        tree <- rowTree(x)
```
FelixErnst, line 210: The implementation almost works now. However, since you don't define `tree` for the `SummarizedExperiment` method, line 191 will not work. If you just remove it, it will work again.

TuomasBorman: removed

R/estimateDiversity.R
TuomasBorman: Should it be defined somewhere? That was just an attempt to fix a warning that said `tree is not defined`

TuomasBorman: I removed the definition from the current version

R/estimateDiversity.R
```diff
@@ -178,13 +188,123 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+        # Adds calculated Faith index to colData
+        .add_values_to_colData(x, faith, name)
+
+          }
+)
+
+#' @rdname estimateDiversity
+#' @export
+setMethod("estimateFaith", signature = c(x="TreeSummarizedExperiment", tree="missing"),
+    function(x, tree = rowTree(x), abund_values = "counts",
```
FelixErnst: The same thing as for you most recent changes applies here. If you remove the tree argument from the function definition alltogether, which you are allowed to do, since `tree` is defined as missing, and add `tree <- rowTree(x )` in the function it will be more consistent.

The rational behind it is as follow:
1. SE + tree: dispatch directly to correct method
2. TSE: tree gets extracted and next call to `estimateFaith` is issued. Since a TSE is also a SE, it matches the signature for 1.
1. If run with a TSE + tree, it gets dispatched to 1. directly as done internally in 2. This also works it the TSE does not contain a `rowTree`.

This is the beauty of S4 dispatch 😄 to cover all the possible combinations of available data

tests/testthat/test-0diversity.R
```diff
@@ -38,4 +38,19 @@ test_that("diversity estimates", {
...
+
+        # Absent taxa are dropped from the tree
+        sub_tree <- ape::drop.tip(rowTree(tse_idx), absent)
+        # Faith is now calculated based on the sub tree
+        faith <- sum(sub_tree$edge.length)
+
+        expect_equal(cd$faith[i], faith)
+    }
+
+
```
FelixErnst: Can you add test for all the combination mentioned above?
- SE + tree
- TSE 
- TSE + tree

hint: you can always "downgrade" a TSE to SE using `as`.

R/estimateDiversity.R
FelixErnst: That warning was the result of line 191 contain the tree variable, which is not defined in that function (and also is not used)

TuomasBorman: Ok, clear

R/estimateDiversity.R
TuomasBorman: Thanks for clarifying, I think that I got it

tests/testthat/test-0diversity.R
TuomasBorman: Added

FelixErnst:
Can you solve the warning? I would merge the PR once it passes the GHA

TuomasBorman:
Now it should work

FelixErnst:
I meant this warning:

```
❯ checking Rd \usage sections ... WARNING
  Undocumented arguments in documentation object 'estimateDiversity'
    ‘tree’
  
  Functions with \usage entries need to have the appropriate \alias
  entries, and all their arguments documented.
  The \usage entries must correspond to syntactically valid R code.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.
```

R/estimateDiversity.R
```diff
@@ -184,7 +194,123 @@ setMethod("estimateDiversity", signature = c(x = "SummarizedExperiment"),
...
+#' @rdname estimateDiversity
+#' @export
+setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
+    function(x, abund_values = "counts",
+             index = c("shannon","gini_simpson","inverse_simpson",
+                       "coverage", "fisher", "faith"),
+             name = index, ..., BPPARAM = SerialParam()){
+        
+        # Gets the tree 
+        tree <- rowTree(x)
```
TuomasBorman, line 210: I was thinking if this should be removed, because estimateFaith is already doing this same job

FelixErnst:
> Now it should work

944920d changes the implementation actually. A different dispatch is used. I would revert this

R/estimateDiversity.R
FelixErnst: But then you cannot use `tree` in the check in line 210 anymore...

FelixErnst:
The warning is issued because the man does not contain an entry for `tree`. Since the man page is also for `SE` for which you need a tree, the param `tree` has to be described.

R/estimateDiversity.R
TuomasBorman: That's true, I take one step back

TuomasBorman:
> The warning is issued because the man does not contain an entry for `tree`. Since the man page is also for `SE` for which you need a tree, the param `tree` has to be described.

Ok, I got little bit confused, because there was param for `tree `in one point in general of `estimateFaith`. So, is that the right place for `tree` param (because general is also for `estimateFaith` of TSE without tree)? Or could it be added to method of SE?

R/estimateDiversity.R
TuomasBorman: reverted to previous state

FelixErnst:
This is not about the implementation, it is about the documentation aka. the roxygen block for `estimateDiversity`. Since all function are tied to that man page via `@rdname estimateDiversity`, all arguments also need to described. Just add the `@param tree` block with a suitable description and update the man pages.

PR #89, opened 2021-04-06
Load from mothur

### LINKED ISSUE #88: Loading Mothur data
We seem to miss Mothur importer, this could be added similar to [mia::loadFromQIIME2](https://github.com/microbiome/mia/blob/master/R/loadFromQIIME2.R).

We have one in [microbiome::read_mothur2phyloseq](https://github.com/microbiome/microbiome/blob/master/R/read_mothur2phyloseq.R) and this might help to get it started.

TuomasBorman:
Hi,

here is the implementation of `loadFromMothur`. I used loadFromQIIME2 and `microbiome::read_mothur2phykoseq` as references. 

I have tested this with the data (Baxter_FITs_Microbiome_2016) that I found from one of the older versions of `microbiome.` However, it did not include phylotree nor refseq, so I haven't tested those functionalities. 

With those abundance table, taxa data, and sample meta data, `loadFromMothur` seems to work. 

Issue: https://github.com/microbiome/mia/issues/88

-Tuomas


R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+                           sampleMetaFile = NULL,
+                           featureNamesAsRefSeq = TRUE,
+                           refSeqFile = NULL,
+                           phyTreeFile = NULL,
+                           ...) {
+  
+    # input check
+    if(!.is_non_empty_string(featureTableFile)){
+      stop("'featureTableFile' must be a single character value.",
+           call. = FALSE)
```
antagomir: Are the indentations here and below correct? Shouldn't it always be multiples of 4 by Bioc standards?

antagomir (commented):
Seems good! Can be merged after my comments, if @FelixErnst agrees.

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
+#' Import Mothur results to \code{TreeSummarizedExperiment}
+#'
+#' @param featureTableFile a single \code{character} value defining the file
```
FelixErnst: Description text is missing. Please add it, even though it might be only one sentence

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+#'   will be disregarded, if \code{refSeqFile} is not \code{NULL}. If the
+#'   feature names do not contain valid DNA characters only, the reference
+#'   sequences will not be set.
+#'
+#' @param refSeqFile a single \code{character} value defining the file path of
+#'   the reference sequences for each feature. (default: \code{refSeqFile =
+#'   NULL}).
+#'
+#' @param phyTreeFile a single \code{character} value defining the file path of
+#'   the phylogenetic tree. (default: \code{phyTreeFile = NULL}).
```
FelixErnst: I am not really familiar with mothur output. Are those data types available from a typical mothur run?

If not, please remove them for clarity.

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+#'
+#' @param ... additional arguments:
+#'
+#' @details
+#' Results exported from Mothur can be imported as a
+#' \code{TreeSummarizedExperiment} using \code{loadFromMothur}. Except for the
+#' \code{featureTableFile}, the other data types, \code{taxonomyTableFile},
+#' \code{refSeqFile} and \code{phyTreeFile}, are optional, but are highly
+#' encouraged to be provided.
```
FelixErnst: Adjust if necessary based on comment above

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+#' @return  A
+#' \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#' object
+#'
+#' @name loadFromMothur
+#' @seealso
+#' \code{\link[=makeTreeSummarizedExperimentFromphyloseq]{makeTreeSummarizedExperimentFromphyloseq}}
+#' \code{\link[=makeTreeSummarizedExperimentFromBiom]{makeTreeSummarizedExperimentFromBiom}}
+#' \code{\link[=makeTreeSummarizedExperimentFromDADA2]{makeTreeSummarizedExperimentFromDADA2}}
+#' \code{\link[=makeTreeSummarizedExperimentFromDADA2]{loadFromQIIME2}}
```
FelixErnst: wrong link

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+#' # Abundance table
+#' counts <- "~/Desktop/Baxter_FITs_Microbiome_2016_fit.final.tx.1.subsample.shared"
+#' # Taxa table
+#' taxa <- "~/Desktop/Baxter_FITs_Microbiome_2016_fit.final.tx.1.cons.taxonomy"
+#' # Sample meta data
+#' meta <- "~/Desktop/Baxter_FITs_Microbiome_2016_mapping.csv"
+#' 
+#' # Creates tse object from files
+#' \dontrun{tse <- loadFromMothur(counts, taxa, meta)}
+#'
```
FelixErnst: Hard coded files are a no-go. Please add them as example data and reference them in the examples via `system.file`. Then the data can actually be loaded and the `\dontrun` section can be "removed" aka. the code actually be run

R/loadFromMothur.R
FelixErnst: yep those are not correct.

edit: the indentation

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    
+    # If colData information exists, gets that. Otherwise, sample_tab is just data frame without information
+    if (!is.null(sampleMetaFile)) {
+      sample_meta <- .read_mothur_sample_meta(sampleMetaFile, feature_tab)
+    } else {
+      sample_meta <- S4Vectors:::make_zero_col_DataFrame(ncol(feature_tab))
+    }
+    
+    if (!is.null(phyTreeFile)) {
+      tree <- .read_qza(phyTreeFile, ...)
```
FelixErnst: This is not applicable here, since the `.read_qza` function is `QIIME2` output specific. You probably just need to add another function as warpper for `ape::read.tree`

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      tree <- NULL
+    }
+    
+    # if row.names(feature_tab) is a DNA sequence,  set it as refseq
+    if (!is.null(refSeqFile)){
+      refseq <- .read_qza(refSeqFile, ...)
```
FelixErnst: See comment above. This requires your own kind of wrapper, since the assumption for `.read_qza` are not assumable for mothur output.

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    # Reads the file
+    assay <- read.table(featureTableFile, check.names=FALSE, header=TRUE,
+                      sep="\t", stringsAsFactors=FALSE)
+  
+    # File contains additional columns "label", "numOtus", and "Group". They are removed
+    assay$label <- NULL
+    assay$numOtus <- NULL
+    # Information of "Group" column is saved, it includes taxa information
+    x <- assay$Group
+    assay$Group <- NULL
```
FelixErnst: Why? They are supposed to end up in the `rowData`

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    
+    # Initializes rownames
+    rownames(assay) <- NULL
+    # Saves taxa to rownames
+    rownames(assay) <- x
```
FelixErnst: Where does `x` come from?

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    
+    # Transposes and converts assay from dataframe to matrix
+    assay <- t(as.matrix(assay))
+    
+    return(assay)
+
```
FelixErnst, line 218: Please remove those empty lines. (All across the PR)

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+                               extra="merge")
+    # Deletes ";" from the end of Genus names
+    rowData$Genus <- gsub(";", "", rowData$Genus)
+    # Deletes addition "Size" column
+    rowData$Size <- NULL
+    
+    # Adds taxa to rownames
+    rownames(rowData) <- rowData$OTU
+    # Deletes addition "OTU" column
+    rowData$OTU <- NULL
```
FelixErnst: This section is a bit weird in the sense that the data stored as `rownames` here must match the `rownames` of the assay, if they are set. From the code this is not clearly managed which is the actual true feature name in the end returned by `rownames(tse)`.

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    
+    # Creates a matrix from the table
+    rowData <- as.matrix(rowData)
```
FelixErnst: `rowData` will be a `DataFrame` in the end, so this conversion does make sense

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    
+    # Gets only those taxa that are included in feature_tab
+    rowData <- rowData[rownames(feature_tab),]
```
FelixErnst: Why? Additional data is allowed isn't it? Why not import it as well?

And also, why is the subset necessary? Shouldn't the dimension of the data match out of the box?

R/loadFromMothur.R
```diff
@@ -0,0 +1,206 @@
...
+    return(rowData)
+
+}
+
+.read_mothur_sample_meta <- function(sampleMetaFile, feature_tab, ...){
+  
+    # Reads the file
+    colData <- read.csv(sampleMetaFile, row.names=1, check.names=FALSE)
+    # Gets only those samples that are included in feature_tab
+    colData <- colData[colnames(feature_tab), ]
```
FelixErnst: Same as above. Why don't the dimension match? Or the order?

I don't think that this hand-holding is necessary. If mothur supplies data which matches, we don't need to make sure, the user didn't jumble the data around.

If mothur supplies jumbled-up data, then I see the point.

FelixErnst (changes requested):
Hi Tuoams,

thanks for the PR.

Some thing need to be done and some need to be discuss. Have a look the individual comments and let me know, what you think about the issues raised. Thanks!

Felix

antagomir:
Great, I agree with all. Somehow many of these had escaped my attention, need to try and be more sharp next time!

R/loadFromMothur.R
TuomasBorman: Fixed

R/loadFromMothur.R
TuomasBorman: Added

R/loadFromMothur.R
TuomasBorman: I'm not familiar either with Mothur, so that is why I basically tried to just implemented the microbiome's function

It seems that Mothur output files can include:

- assay: https://mothur.org/wiki/shared_file/
- rowData: https://mothur.org/wiki/taxonomy_file/ or https://mothur.org/wiki/constaxonomy_file/
- colData: https://mothur.org/wiki/design_file/

(Full file type list: https://mothur.org/wiki/tags/#file_types)

So, these file types above should be allowed as input?
I should probably try to find more presentative data so I don't have to do this blindfolded

R/loadFromMothur.R
TuomasBorman: Ok, should I avoid empty rows also in the future? Personally, I find it easier to read code when it is little bit grouped, but for others it might be confusing, unnecessary, or annoying

R/loadFromMothur.R
FelixErnst: > So, these file types above should be allowed as input?

I guess so

R/loadFromMothur.R
FelixErnst: That is a fine line. Grouping steps is helpful, but a break using `#` with a short description would work for that as well. An empty line after or before a return is not helpful. Have a look at the other code in `mia`. I think you will get it.

R/loadFromMothur.R
microsud: Yes these are standard outputs of mother.

R/loadFromMothur.R
TuomasBorman: According to <https://mothur.org/wiki/shared_file/, 

- " label" is just first value of original data's first column. It is just for comparison
- "Group" contains the name of samples, which will be stored as colnames
- "numOtus" contains just the number of taxa that was found from the sample

R/loadFromMothur.R
TuomasBorman: x is sample names. I have named the variable differently now

R/loadFromMothur.R
TuomasBorman: So, should the `rownames(tse) `be set somehow?  I assumed that it is checked that `rownames(assay) == rownames(rowData)`. Otherwise, I think it throws an error when TSE is created: `"rowdata and assay are not parallel" `or something

R/loadFromMothur.R
TuomasBorman: rownames should match, and subsetting is unnecessary. I checked my data, and for some reason in taxonomic file random taxa are dropped from the middle

R/loadFromMothur.R
FelixErnst: All three variables should end up in `colData`, since you cannot expect all users to not need it.

TuomasBorman: Ok

R/loadFromMothur.R
FelixErnst: Having read the mothur man page, make sure to use their nomenclature. From the names you are using right now, it is for example not clear, in which file the count data is expected to be.

R/loadFromMothur.R
FelixErnst: Exactly. So it is critical, that you have a consistant strategy. If I remember correctly, assay names take precedence, when creating a `SE`.

Currently, you are doing the following steps:

- get assay data
- set the colnames of the `SE` of the assay as rownames, but then transpose the matrix
- get the rowData and set the rownames of the `SE` as rownames of the rowData
- get colData and just keep the columns which match the rownames, ... ah sorry the colnames, of the assay

So the strategy works for sure, but is not really consistent, since you are using three different types of using the dimnames

Good practice would be:

- to extract the assay at a single point
- to extract the colnames at a single point
- to extract the rowData at a single point
- to extract the rownames at a single point
- to extract the colData at a single point
- to combine the data and set the dimnames in a single step

Especially the subsetting/ordering of the colData leaves me with a weird feeling, since in the worst case scenario, `NA` get produced without the user getting notified, that the data is not matching. The second thing is the assay and dimname setting. I would do something along the lines of for the assay extraction:

```r
MOTHUR_NON_ASSAY_COLS <- c("label","numOtus","Group")
data <- read.table(featureTableFile, check.names=FALSE, header=TRUE,
                      sep="\t", stringsAsFactors=FALSE)
if(){
... # test for colnames not NULL
... # test for colnames present
}
assay <- as.matrix(data[,!(colnames(data) %in% MOTHUR_NON_ASSAY_COLS )])
assay <- t(assay)
colData <- DataFrame(data[,MOTHUR_NON_ASSAY_COLS])
return(list(assay = assay,
         colData = colData))
```

This way you don't touch any data, you don't have to and it is really easy to change, since you only have to change `MOTHUR_NON_ASSAY_COLS` to affect the outcome.

FelixErnst: For the rowData extraction, I would do something similar:

```r
 # Reads the file
MOTHUR_TAX_COL <- "Taxonomy"
data <- read.table(taxonomyTableFile, check.names=FALSE,
                      header=TRUE, sep="\t", stringsAsFactors=FALSE)
if(){
... # test for colnames present
... # check for MOTHUR_TAX_COL  present in colnames
}
# split taxa level into seperate columns
data [,MOTHUR_TAX_COL] <- gsub("[\"]", "", data [,MOTHUR_TAX_COL])
data [,MOTHUR_TAX_COL] <- gsub("[(1-100)]", "", data [,MOTHUR_TAX_COL])
tax <- tidyr::separate(data , MOTHUR_TAX_COL, into=c("Kingdom", "Phylum", "Order", "Class", "Family", "Genus"), sep=";", extra="merge")
tax$Genus <- gsub(";", "", tax$Genus)

# 
rowData <- cbind(DataFrame(tax),
                DataFrame(data[, !(colnames(data) %in% MOTHUR_TAX_COL)]))
if(!is.null(rowData$OTU)){
    if(){
    ... # check if the rownames are the same as in assay.
    }
    rownames(rowData) <- rowData$OTU
}
return(rowData)
```

Here again, you don't touch anything you don't need to and you make it clear, from the code which information takes precedence. If I review this code (I know, not the best example), then it is immediately clear, that the assay data rownames have the higher priority and the integrity of the data is ensured by the code.

R/loadFromMothur.R
FelixErnst: They should match, but if someone mixes up their files? A lot of `NA` get produced or a subset out of bound error is triggered, without any clear indication, what the problem is.

If the rownames should match, than you have to make sure, e.g. with this
```r
if(!all(.. == ...)){
   stop()
}
```

R/loadFromMothur.R
```diff
@@ -82,22 +85,28 @@ loadFromMothur <- function(featureTableFile,
...
+    # }
+    # if(!is.null(phyTreeFile) && !.is_non_empty_string(phyTreeFile)){
+    #     stop("'phyTreeFile' must be a single character value or NULL.", 
+    #         call. = FALSE)
+    # }
     #
     
     # Reads the featureTablefile 
-    feature_tab <- .read_mothur_feature(featureTableFile, ...)
+    feature_tab_and_data_to_colData <- .read_mothur_feature(featureTableFile, ...)
```
FelixErnst: As a discussion: The name `featureTableFile` is that used in the context of mothur? Is it a default name for mothur? I know, that the functions is informed a lot by `loadFromQIIME2`, but I think it not required to match the name.

I would put more emphasis on the view point of a typical mothur user. He has to recognize the file type by its name and know immediately, which file from mothur belong to which argument for `loadFromMothur`.

Is that the case for the argument names used here? I never have used mothur, so maybe someone else can offer an opinion? @antagomir @microsud

R/loadFromMothur.R
```diff
@@ -106,27 +115,30 @@ loadFromMothur <- function(featureTableFile,
...
+    # # if row.names(feature_tab) is a DNA sequence,  set it as refseq
+    # if (!is.null(refSeqFile)){
+    #     refseq <- .read_qza(refSeqFile, ...)
+    # } else if (featureNamesAsRefSeq) {
+    #     refseq <- .rownames_as_dna_seq(rownames(feature_tab))
+    # } else {
+    #     refseq <- NULL
+    # }
     
     return(TreeSummarizedExperiment(
```
FelixErnst: Since the input from mothur, cannot require `TSE` to be used (now reference sequence, no tree), I would recommend just returning a `SE`.

R/loadFromMothur.R
TuomasBorman: I agree that it could be wise those change these names to match with Mothur

R/loadFromMothur.R
TuomasBorman: Changed

R/loadFromMothur.R
TuomasBorman: E.g.

assay = shared
rowData = constaxonomy
colData = design

However, I'm not sure if those are the names that are used in reality

FelixErnst: What do they use in their manual? If I remember correctly the `shared` name is also a documentation title, isn't it?

`taxonomyFile` and `designFile` are good names IMO. For the `sharedFile` a detailed entry on the man page might help as well.

TuomasBorman: Yes, those names that I listed are the names that they use as documentation titles

-    https://mothur.org/wiki/shared_file/
-    https://mothur.org/wiki/constaxonomy_file/
-    https://mothur.org/wiki/design_file/

FelixErnst: `constaxonomy` looks like a typo, but, hey, not our cup of coffee. `taxonomyFile` would probably match close enough, wouldn't it?

TuomasBorman: It might, but they actually have two different taxonomy files:

- taxonomy file https://mothur.org/wiki/taxonomy_file/
- constaxonomy file https://mothur.org/wiki/constaxonomy_file/

Now only `constaxonomy `is allowed. Should `taxonomy `file be allowed aswell? I was thinking, if the name is `taxonomyFile,` then it could be misleading, because in reality it is `constaxonomy `file

FelixErnst: Good catch. I think you could extend the current function to support both.
 I would implement like this:
- read in file without header
- check whether the first lines contains valid columns names
- If yes proceed as `constaxonomy`
- If no:
  - check that two columns exist
  - process the first column as rowData column `sequenceName`
  - process the second column into taxonomic groups and check the result
  - check that the dimension of the taxonomy matched the assay

I think that should yield a valid result.

R/loadFromMothur.R
TuomasBorman: tse --> se

DESCRIPTION
```diff
@@ -73,7 +73,8 @@ Suggests:
     biomformat,
     reldist,
     ade4,
-    microbiomeDataSets
+    microbiomeDataSets,
+    rmarkdown
```
FelixErnst, line 77: Why? Thats covered via `BiocStyle` isn't it?

TuomasBorman: There was an error that said that it should be added there

TuomasBorman: image

FelixErnst: Good catch, but "Oh damn". I hope this doesn't crash a large number of the CRAN/Bioconductor builds. I haven't received any mail regarding this from the maintainer of knitr. So lets hope I don't have any other packages to modify.

tests/testthat/test-4IO.R
```diff
@@ -40,6 +40,19 @@ test_that("Importing dada2 objects yield TreeSummarizedExperiment objects", {
...
+    
+    counts <- system.file("extdata", "mothur_example.shared", package = "mia")
+    taxa <- system.file("extdata", "mothur_example.cons.taxonomy", package = "mia")
+    taxa2 <- system.file("extdata", "mothur_example.taxonomy", package = "mia")
+    meta <- system.file("extdata", "mothur_example.design", package = "mia")
+    se <- loadFromMothur(counts, taxa, meta)
+    se2 <- loadFromMothur(counts, taxa2, meta)
+    
+    expect_s4_class(se, "SummarizedExperiment")
+    expect_s4_class(se2, "SummarizedExperiment")
```
FelixErnst, line 66: Can you add a few tests on the dimension? This ist just to make sure, that the `SE` actually conforms to the expectations (An `SE` can in theory have 0 rows or 0 columns).

Please check the following
- dimensions
- assays present
- rowData present + correct dimension
- colData present + correct dimension

Thanks

R/loadFromMothur.R
```diff
@@ -0,0 +1,232 @@
...
+    ex <- strsplit(basename(file), split = ".", fixed = TRUE)[[1]]
+    ex[length(ex)]
+}
+
+#' extract file extension
+#' @noRd
+.get_mothur_taxonomy_file_type <- function(file) {
+    ex <- strsplit(basename(file), split = ".", fixed = TRUE)[[1]]
+    paste(c(ex[length(ex)-1], ".", ex[length(ex)]), collapse= "")
+}
```
FelixErnst: These checks only work based on the name of the file. I think this is not really a solid solution, since problems can occur in to modes:
- wrong file name, right data
- right file name, wrong data

Mode 1 is not a severe problem, but mode 2 will produce unexpexted results. 

I would suggest the following:
- read in the data via `read.table`, because this is the common format (all files are tab-delimited files)
- check, if column names match certain assumptions. From what I can tell, this is actually the only requirement for correct processing of the data in following steps: A `shared` file requires the columns `"label","numOtus","Group"`, a taxonomy file either has the `Taxonomy` column or exactly `two` columns and the same number of rows than the shared file has data and a meta data file has column names and the same number of rows as the shared file has rows

Regarding the check implementation, it would be better to formulate the check differently. Instead of having a function return a type and do the check outside of the function, have a function like `.is_mothur_taxonomy_file` return a `boolean` result. That way the logic is contained in a single function and not separated into to two, the function and the check via `if`. This makes the code better maintainable and splits the logic for the individual files up. Each can be adjusted independently

tests/testthat/test-4IO.R
TuomasBorman: Added tests

R/loadFromMothur.R
TuomasBorman: I modified those file type checks

FelixErnst (approved):
made a few formatting changes. Good to go after the checks clear. Thanks Tuomas for sticking with it! @tvborm

PR #90, opened 2021-04-28
Fixed a bug that appears when file doesn't include all data

TuomasBorman:
Hi!

I found a bug from `makeTreeSummarizedExperimentFromBiom`. I have a biom-file and I wanted to create `TSE` from it. The problem is that it's not include sample metadata, i.e., `colData`. When `TSE` object is created `NULL` leads to an error. I fixed it so that now it is possible to create `TSE` object if biom-file includes only counts.  

I was wondering if this should be constructed like `loadMothur`, i.e., there could be additional parameters for `rowData` and `colData` files. 

1. User could choose if those files are provided. 
2. First, `rowData` and `colData` are fetched from the `biom-file`.
3. If user has provided additional `rowData` and `colData` files, they replace those ones that was fetched from `biom-file`. 
4. `SE` is created instead of `TSE`.

However, I'm not familiar with biom-format. I don't know if the file should normally include `colData` and `rowData`, or is my file just an exception and additional parameters are not needed. 

-Tuomas

### DIFF
--- a/R/loadFromMothur.R
+++ b/R/loadFromMothur.R
@@ -149,7 +149,7 @@ loadFromMothur <- function(sharedFile,
                            stringsAsFactors=FALSE, 
                            col.names = c("OTU", "Taxonomy"))
     }
-    # Else the is not either gives an error
+    # Else the file is not either gives an error
     else{
         stop("The input '", taxonomyFile, "' must be provided in the ",
              "`taxonomy` or `cons.taxonomy` format. In addition, it must ",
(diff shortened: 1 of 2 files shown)


FelixErnst:
Hi Tuomas,

thanks for catching that and the PR.

I think it is only necessary to expand `colData` to a `DataFrame` and not `rowData`, since `rowData = NULL` is the default in the `SummarizedExperiment` constructor `TreeSummarizedExperiment()` redirects to.

Felix

TuomasBorman:
Now empty data frame is added only to `colData`

PR #92, opened 2021-04-29
bugfix for applying pseudocount at wrong point in sequence (fixes #91)

### LINKED ISSUE #91: CLR transform doesn't include relative abundance step
Hi
I noticed that when computing the CLR transform from the transformCounts() function and comparing it to the result from when I first calculate the relative abundances and then apply the CLR transformation, the results were different. Leo said it would be good to mention this to see if you want to keep it this way or change it. Here is an example that reproduces what I mentioned above.

```
library(mia)

data("GlobalPatterns")
tse <- GlobalPatterns

tse1 <- relAbundanceCounts(tse)
tse1 <- transformCounts(tse1, abund_values = "relabundance", method = "clr", pseudocount = 1)

tse2 <- transformCounts(tse, abund_values = "counts", method = "clr", pseudocount = 1)

assays(tse1)$clr == assays(tse2)$clr
```

FelixErnst:


R/transformCounts.R
```diff
@@ -308,88 +303,75 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
+    # Gets the relative abundance
+    mat <- .calc_rel_abund(mat)
+    mat <- .apply_pseudocount(mat, pseudocount)
     # If there is negative values, gives an error.
     if (any(mat < 0, na.rm = TRUE)) {
         stop("Abundance table contains negative values and hellinger ",
-             "transformation is being applied without pseudocount. Try ",
-             "hellinger with pseudocount = 1 or other numeric value.",
+             "transformation is being applied without (suitable) pseudocount. ",
+             "Try hellinger with pseudocount = 1 or other numeric value.",
```
antagomir, line 324: Does this apply pseudocount = 1 on relative abundances - this would be suitable for counts but not relative abundances.

R/transformCounts.R
```diff
@@ -308,88 +303,75 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
-.calc_clr <- function(mat, threshold){
-
+.calc_clr <- function(mat, pseudocount, ...){
+    mat <- .calc_rel_abund(mat)
+    mat <- .apply_pseudocount(mat, pseudocount)
     # If there is negative values, gives an error.
     if (any(mat <= 0, na.rm = TRUE)) {
         stop("Abundance table contains zero or negative values and ",
              "clr-transformation is being applied without pseudocount. ",
              "Try clr with pseudocount = 1 or other numeric value.",
```
antagomir, line 340: Same comment here. If pseudocount is added on relative abundances, then 1 is not necessarily a good option(?)

tests/testthat/test-5transformCounts.R
```diff
@@ -117,9 +112,16 @@ test_that("transformCounts", {
...
             }))
+        
+        
+        tse2 <- relAbundanceCounts(tse)
+        expect_true(all(round(as.matrix(assays(mia::transformCounts(tse2, method = "clr", 
+                                                                    abund_values = "relabundance", pseudocount = pseudonumber))$clr) -
+                                  as.matrix(assays(mia::transformCounts(tse, method = "clr", pseudocount = pseudonumber))$clr),
+                              8) == 0))
 
         #############################################################
```
antagomir, line 126: How about also adding 

```
data("GlobalPatterns")
tse <- GlobalPatterns
tse1 <- relAbundanceCounts(tse)
tse1 <- transformCounts(tse1, abund_values = "relabundance", method = "clr")
tse2 <- transformCounts(tse, abund_values = "counts", method = "clr")
assays(tse1)$clr == assays(tse2)$clr
```

These should have the same default behavior. How to do it with pseudocount, perhaps need to be discussed.

FelixErnst: This will result in an error due to potential rounding errors. Calculating the difference, rounding it and checking whether it is zero is more stable

R/transformCounts.R
FelixErnst: Thats up to the user. He has to decide, what to do. A check whether relative abundances are already present is not up to us. (The man page clearly spells out the formula used)

tests/testthat/test-5transformCounts.R
antagomir: Oh yes, that's fine. I primarily meant to check that this default behavior is the same. Test can be implemented in any way that works best.

R/transformCounts.R
FelixErnst: You can add pseudocount of `-1`. That would however not be suitable in some cases.

R/transformCounts.R
FelixErnst: What is commonly used for `clr`?

tests/testthat/test-5transformCounts.R
FelixErnst: Already done. See line 120-124

R/transformCounts.R
antagomir: Why do we want to recommend using pseudocount 1 here ? - this is potentially misleading.

FelixErnst: I did assume with Chandlers bug report, that this is usually done like this. First relative abundance, then pseudocount and then the rest of the transformation. 

If that is not the case, than the procedure he used shouldn't work or at least should have the same result.

FelixErnst: What kind of pseudocount is commonly used for `clr` transformation?

antagomir: The way this should work is that relative abundance is always done before clr transformation (as a preprocessing step). Without pseudocount by default.

Then this process should return the same result regardless on whether it is being applied on relative abundance or count values.

Unfortunately, clr gets easily broken with zero values and most of the time pseudocount is needed. 

If pseudocount is added, the default choice would be to add it on the count data and use pseudocount = 1.

If relative abundances contain zeroes, then it suggests that relabundances were calculated without pseudocount.

If pseudocount is added on relative abundances instead of count data, a default choice could be `pseudocount = min(x[x>0])`.

That should yield identical result (within tolerance) for those two alternative ways to do this in most cases. 

Regarding warning messages, we should either not give recommendations on pseudocount values, or we should give recommendations that remind the user that default choice will depend on the assay type (count / relabundance). 

In general, clr should not be done for other than count or relabundance assay. Perhaps the user should get a warning is such case.

Later when time allows, we should implement rclr which is a robust version that avoids this. I can open issue about that.

R/transformCounts.R
antagomir: Same comment than below. The default pseudocount choice should be different for counts and for relabundances. Hellinger should also yield identical result for both of these by default.

R/transformCounts.R
FelixErnst: >If pseudocount is added, the default choice would be to add it on the count data and use pseudocount = 1.
>If relative abundances contain zeroes, then it suggests that relabundances were calculated without pseudocount.
>If pseudocount is added on relative abundances instead of count data, a default choice could be pseudocount = min(x[x>0]).

There is no way, we can check for these things, so it is the user who has to make sure. We can just make sure the calculation is correct, but we cannot make sure that the calculation is the correct choice.

However, we can spell it out even more directly on the man page.

>In general, clr should not be done for other than count or relabundance assay. Perhaps the user should get a warning is such case.

How to check for these cases? I don't see bulletproof way.

FelixErnst: A PR making sure, that the man page is the best version possible, would be good, but I don't think the code can adjusted to avoid every pitfall there is.

antagomir: Yes there are limits to how much should be done for the user. But:

1) man page should give clear recommendation, explanation, and examples

2) warning messages should not encourage naive users make bad choices (i.e. if default values are recommended for pseudocount, then let us also spell out that different default values are suitable for different cases)

antagomir:
I will test the behavior of the updated function as soon as I have the chance.

PR #95, opened 2021-04-30
Transform

antagomir:
I did some fixes on the documentation as well as functions so that this matches better with my view on how these should be used.


DESCRIPTION
```diff
@@ -35,7 +35,7 @@ License: Artistic-2.0 | file LICENSE
 Encoding: UTF-8
 LazyData: false
 Depends:
-    R (>= 4.1),
+    R (>= 4.0),
```
FelixErnst: This must be R 4.1

R/transformCounts.R
```diff
@@ -154,7 +168,7 @@ setMethod("transformCounts", signature = c(x = "SummarizedExperiment"),
              name = method,
              pseudocount = FALSE,
              threshold = 0){
-        
+
```
FelixErnst: Remove empty lines

R/transformCounts.R
```diff
@@ -272,18 +290,18 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
-.get_transformed_table <- function(assay, method, pseudocount, threshold){
+.get_transformed_table <- function(assay, method, threshold){
     # Function is selected based on the "method" variable
     FUN <- switch(method,
                   relabundance = .calc_rel_abund,
                   log10 = .calc_log10,
                   pa = .calc_pa,
                   hellinger = .calc_hellinger,
                   clr = .calc_clr)
+
```
FelixErnst, line 314: See above

R/transformCounts.R
```diff
@@ -374,8 +382,10 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
 .apply_pseudocount <- function(mat, pseudocount){
     # If "pseudocount" is not FALSE, it is numeric value specified by user. 
     # Then add pseudocount.
+
```
FelixErnst: See above

R/transformCounts.R
```diff
@@ -374,8 +382,10 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
     if(!pseudocount==FALSE){
         mat <- mat + pseudocount
     }
+
```
FelixErnst: See above

FelixErnst (commented):
Hi Leo,

Thanks for the PR. Especially to example are really helpful. Can you fix the indentation in the tests?

In addition, I have a few questions:

1. Since you refactored the code to avoid user errors as per the discussion on slack, I am not sure, why you would allow a user to add pseudocount to a relative abundance calculation.
2. Wouldn't it be better to have the code acting for a transformation be bundled up in one function instead of being spread out? (pseudocount adding in one place and doing the transformation in another) This is hard to maintain.

antagomir:
Will do the changes later this weekend.

1. pseudocount - it is possible that user only has access to relabundances (from data provider for some reason - we bump into this sometimes). Then CLR would not be possible without adding pseudocount on relative abundances. We would still like to support operating on relabundance when original count data is missing.

2. Does this refer to the Ztransformation or more generally?

FelixErnst:
1. With your implementation, I could run `transfromCounts(..., method = "relabundance", pseudocount = 1000)` and it would work. This is dangerous. However, if the pseudocount argument is available for `clr`, which requires relative counts, I don't think it is a problem (and you explain it in the example).
2. Generally speaking. Therefore and because of point 1, , I think it would be good, if you would revert the changes made for applying the pseudocount. It should be located within the function for the actual calculation and not before in the general wrapper.

antagomir:
Do you mean that pseudocount would not be applicable to certain transformations (in particular the compositional one)?

FelixErnst:
Yes. it shouldn't be applicable per se, just for specific transformations

FelixErnst:
Can you update the branch from master?

R/transformCounts.R
antagomir: In fact I have always thought that code is easier to read (to me at least) when there is some space. I do not see a problem in having space to improve readability but I have now removed the empty lines per this request. Curious to know if there is a justification or is this just a stylistic preference.

antagomir:
Thanks. I am now aiming to finalize this.

Regarding the above discussion: adding pseudocount to relative abundances is sometimes necessary in practice (I am not saying that it should be recommended in general). Hence I would not remove that option. If we think that it is a risk that naive users give problematic values (like pseudocount = 1000 for relabundances), then one solution is to provide a warning in such case.

R/transformCounts.R
FelixErnst: It is generally regarded as a "code smell" signifying other things at play. These could be:
- Not enough compartmentalizing with functions
- A missing comment to summarize a code chunk
- Unclear coding style (short/unclear variable name and other unclear formating)

Because we don't do it generally in the rest of the project, I would like to avoid letting it spread further.

antagomir: Fine, let's do so then. I just find it less comfortable to read but I can live with that.

FelixErnst:
Can you add this warning for `pseudocount != FALSE` for relative abundance calculation?

antagomir:
> Can you add this warning for `pseudocount != FALSE` for relative abundance calculation?

Done

FelixErnst:
> Regarding the above discussion: adding pseudocount to relative abundances is sometimes necessary in practice 

> Can you give an example? I cannot think of any. Maybe this should be an `error` instead of a `warning`?


For instance when willing to analyse log relative abundances. Log0 is not defined.

FelixErnst (approved):

Issue #94, opened 2021-04-30
Add new method to transformCounts for returning ranks of abundances


antagomir:
Rank index is needed in certain non-parametric operations. I suggest to add this as one transformation. 

Abundances would be transformed as: `x <- order(x)`

Related to #93

FelixErnst:
The concept is not clear to me.

In addition the command `x <- order(x)` is ambiguous for any two dimensional object and I also would like point out two things problematic with concept as laidout above:
1. `order` returns an ordering integer vector. Overwriting `x` with it probably does not make sense. Did you mean `sort`?
2. Ordering of `SE` object is already available via `se[new_order_row,new_order_col]`. This makes it very versatile, because the user can create an enormous amount of different orders using data from `rowData` or `colData`. There is no wrapping function needed.

I think this is a case of "use/teach base R behaviour" to avoid any unneccessary sugar.

antagomir:
Rank tests are a routine workhorse for non-parametric tests and having the ability to add such assay easily through a transformation would be useful. Because such transformation needs to be done for certain tests or visualizations.

The `x <- order(x)` was a brief example on how to do this per row or per column. In practice a sweep will be needed. Order returns the rank for each entry. I did mean the rank order and not sort.

The idea is not to reorder rows or columns but transform the abundances to their corresponding ranks per sample (or per taxa).

FelixErnst:
So the idea would be more like function named `getRankOfAbundance`?

FelixErnst:
or more like `transformCounts(..., method = "rank")`?

antagomir:
The latter, yes.

antagomir:
I got to test this now. The `.calc_rank` function internally uses the `ties.method` argument (of colRanks function). This has been set to `ties.method="min"`. This is not the default in colRanks, and the user is not given an option to change this. 

I propose the following changes:
- default should be ties.method="first"
- user should have access to this argument through `transformCounts` (`"..."` option)
- example could be added, including an example on how to change the ties.method in `transformCounts`
- unit tests should be modified to match this

FelixErnst:
I am of two minds on this

1. This method implements a specific fixed rank calculation.
2. If it is just a wrapper of colRanks, maybe we should skip implementing the wrapper and just use `assay(se,"whatever") <- colRanks(assay(se, "whichever"))`

So from that point of view, I thought this was a specific use case, which warrants fixing the `ties.method` argument. If that is not the case, removing the whole function again would make sense

[...]

PR #97, opened 2021-05-03
Rank to transform counts

TuomasBorman:
Hi,

here is the implementation of "rank" method. 

`transformCounts(x, method = "rank") `returns object that includes rank table. Ranks are calculated column-wise, i.e., for every sample, ranks of taxa are calculated. 

Issue: https://github.com/microbiome/mia/issues/94

-Tuomas


TuomasBorman:
In addition to that, I made a small change to `makeSummarizedExperimentFromBiom'`s example. Because now the returned object is `SE` and not `TSE`, I changed the name of object from `tse `to `se`.  (Commit "Minor change to example")

R/transformCounts.R
```diff
@@ -349,6 +357,15 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
     return(mat)
 }
 
+.calc_rank <- function(mat, pseudocount, ...){
+    mat <- .apply_pseudocount(mat, pseudocount)
```
FelixErnst: I think this is not necessary. Whichever transformation is applied to the data, the rank should never ever change, should it? Therefore allowing the addition of a pseudocount not required, since it will not change the end result.

R/transformCounts.R
```diff
@@ -349,6 +357,15 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
+    # For every sample, finds ranks of taxa.
+    # Column-wise, NAs are kept as NAs, and ties get the minimum rank value
+    # as.matrix, because otherwise error occurs if mat is, e.g., DelayedArray
+    mat <- apply(as.matrix(mat), MARGIN = 2, FUN = rank, na.last = "keep", ties.method = "min")
```
FelixErnst: `colRanks` from `DelayedMatrixStats` should be much faster.

FelixErnst (commented):
Hi Tuomas,

thanks for the PR.

Small comments/changes required, otherwise good to go.

Felix

R/transformCounts.R
TuomasBorman: True, that is unnecessary step. Removed.

R/transformCounts.R

PR #98, opened 2021-05-06
Faith: colnames replaced with vector of numbers that represent samples

TuomasBorman:
Hi!

I found a bug from `estimateDiversity`. 

I have data that does not contain `colnames` in assay. Currently Faith is using `colnames` and because of that `samples` are `NULL` --> Faith indices are not calculated. 

However, if `samples` are replaced with just a vector that contains index numbers for samples, `samples` is not `NULL` --> Faith indices are calculated.

`samples `doesn't have any bigger meaning here, so this replacement can be done without errors. However, that error that I got was quite weird and might be hard to resolve for user

-Tuomas

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 0.99.17
+Version: 0.99.18
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
              email = "felix.gm.ernst@outlook.com",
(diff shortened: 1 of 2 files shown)


TuomasBorman:
> samples doesn't have any bigger meaning here,

By this I mean that `colnames` are not used as "names", indices works fine

TuomasBorman:
Though I'm not sure if that's more data-related issue than mia-related

FelixErnst:
Can you bump the version?

PR #99, opened 2021-05-14
Bugfix, wrong variable name

TuomasBorman:
Hi, 

I found a bug. There is a variable named `se`, but it does not exist --> error. 

And if user has object named `se`, information is stored there instead of input object, so it can cause that kind of problems also.

-Tuomas

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 0.99.18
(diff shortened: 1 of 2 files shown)


antagomir:
Seems good.

FelixErnst:
Can you bump the version?

FelixErnst:
Can you sync with master? 0.99.19 was created by yesterdays merge

PR #100, opened 2021-05-17
allows splitByRanks/unsplitByRanks also work on plain SummarizedExper…

FelixErnst:
…iments/SingleCellExperiment because no technical requirement for limiting these function on TreeSummarizedExperiment exist


antagomir:
Seems feasible.

PR #101, opened 2021-05-20
transformCounts into transformSamples and transformFeatures

### LINKED ISSUE #93: Split transformCounts to row and col specific functions
The current `transformCounts` contains operations for both samples (clr, hellinger etc.) and features (Z). In the future, we may like to add some that could be applied in either direction (rank index).

1. Splitting `transformCounts` into `transformSamples` and `transformFeatures` would be a more clear and informative naming scheme.

2. Can we consider changing the argument name "abund_values" into "assay" since this is what it essentially does - conceptually more clear?

TuomasBorman:
Hi,

1. I splitted `transformCounts` into `transformSamples` and `transformFeatures`. `transformSamples` works as an alias for `transformCounts`. The difference between `transformFeatures` and `transformSamples` is that in `transformFeatures` the table is first transposed and after that transformation is calculated. 

2. E.g. `.calc_rank` didn't work with `delayedArray`, so table is converted into normal matrix before transformation is done.

3. z-transformation can done also from `transformSamples` and `transformFeatures`. `ZTransform `works as a shortcut for `transformFeatures`. I thought it is logical that z-transformation is also allowed from these functions like all other transformations.

4. I changed the order of methods to alphabetical

-Tuomas


TuomasBorman:
Issue: https://github.com/microbiome/mia/issues/93

R/transformCounts.R
```diff
@@ -97,9 +101,9 @@
 #'
 #'
 #' @return
-#' \code{transformCounts}, \code{relAbundanceCounts}, and \code{ZTransform} return
-#' \code{x} with additional, transformed abundance table named \code{*name*} in
-#' the \code{\link{assay}}.
+#' \code{transformCounts}, \code{transformSamples}, \code{transformFeatures}, 
+#' \code{relAbundanceCounts}, and \code{ZTransform} return \code{x} with additional,
```
FelixErnst, line 105: ZTransform is not exported anymore. I guess the reference should be removed

NAMESPACE
```diff
@@ -2,7 +2,6 @@
 
 export(IdTaxaToDataFrame)
 export(TAXONOMY_RANKS)
-export(ZTransform)
```
FelixErnst: This is not exported anymore. Why? The roxygen tags still exist

FelixErnst:
Looks good.

Can you sort out/give me hint, why Ztransform is removed from the exports, when the tags are still present?

TuomasBorman:
ZTransform was missing @export in generic. Also, relAbundanceCounts did not have that. I added it to both

FelixErnst:
Can you bump the version?

TuomasBorman:
Version bump

Issue #69, opened 2021-03-01
Summarize TSE object


antagomir:
We could add a summary function that could generate / print out basic summaries of a `TSE` object for microbiome data.

This would be roughly equivalent to [microbiome::summarize_phyloseq](https://github.com/microbiome/microbiome/blob/master/R/summarize_phyloseq.R).

FelixErnst:
Sure. Can you prepare an PR? 

What would be the ouput type? From looking at it right now a small `data.frame` would probably be best

antagomir:
A data.frame would be best I think, too. 

We can prepare PR later. We could start prioritizing issues because there are too many PRs to do at the same time.

FelixErnst:
I think this should be added to `getTopTaxa` since this is a summarizing function as well. Maybe the man page needs to be reworked to reflect the intention of the shown functions better, but the foundation for summarizing data is already available.

@microsud

FelixErnst:
@antagomir @microsud any new insides on this? I think a set of summarization functions should be most applicable. `summarizeTaxa` would describe the idea behind the three linked issues the best.

Do you agree that `getDominantTaxa` can be reworked to `summarizeDominantTaxa`? On the same man page `summarizeTaxa` could reside as well.

antagomir:
I am ok with this at least. It is just useful to have 1) easy way to pick up dominant or "top" taxa (various criteria could be possible, considering abundance or prevalence features): and 2) concise summaries as has been implemented by @microsud earlier in phyloseq context.

microsud:
I will work on this in the coming days, was avoiding updating the R version on my system :P
Some comments: 
The `getDominantTaxa` parts of which are inspired from [microbiomeutilties::dominant_taxa](https://github.com/microsud/microbiomeutilities/blob/master/R/dominant_taxa.R) can be renamed as `summarizeDominantTaxa` as it returns a tibble with overview. On the other hand, `dominantTaxa` adds a column to `colData` and therefore I think this should be renamed to `getDominantTaxa` / `addDominantTaxa`. What are your thoughts? @tvborm because you wrote this function, do you have time to make these changes?
The `getTopTaxa` technically returns the most *dominant*/*abundant* taxa based on all samples using the option `sum`, the `dominantTaxa` function is per/sample which may not be useful option to add to `getTopTaxa`.  
Regarding summarize tse, the print(tse) returns a good amount of information. I have some basic idea for adding relevant information to `summarizeTSE` that I will test and let you know.

[...]

PR #104, opened 2021-05-30
summarizeDominantTaxa and addDominantTaxa

TuomasBorman:
Hi,

1. I added `addDominantTaxa`.

2. I also moved `getDominantTaxa `to summaries.R and renamed it to `summarizeDominantTaxa.` However, there is similar error than last time, but I can't find were `getDominantTaxa` is mentioned / what causes the error

```
> devtools::load_all()
ℹ Loading mia
Error in add_classes_to_exports(ns = nsenv, package = package, exports = exports,  : 
  in ‘mia’ methods for export not found: getDominantTaxa
In addition: Warning messages:
1: In setup_ns_exports(path, export_all, export_imports) :
  Objects listed as exports, but not present in namespace: getDominantTaxa
2: no function found corresponding to methods exports from ‘mia’ for: ‘getDominantTaxa’ 
```

3. I renamed getDominantTaxa.R. Now its name is dominantTaxa.R

-Tuomas


FelixErnst:
Thanks for the PR. You have to manually remove `getDominantTaxa` from `NAMESPACE`. roxygen does not automatically remove entries, just add them. In RStudio you can search the whole project with CTRL+SHIFT+F

microsud:
Hi @tvborm thanks for the PR. I can take over from here, add the summrizeSE and fix the issue. Is that okay with you?
Cheers,
Sudarshan

TuomasBorman:
> Thanks for the PR. You have to manually remove getDominantTaxa from NAMESPACE. roxygen does not automatically remove entries, just add them. In RStudio you can search the whole project with CTRL+SHIFT+F

OK, thanks for quick reply

> Hi @tvborm thanks for the PR. I can take over from here, add the summrizeSE and fix the issue. Is that okay with you?
> Cheers,
> Sudarshan

That is ok, thanks!

microsud:
Build on my system without error or warning but has 2 notes
> checking dependencies in R code ... NOTE
  Unexported object imported by a ':::' call: 'S4Vectors:::make_zero_col_DataFrame'
    See the note in ?`:::` about the use of this operator.

> checking for unstated dependencies in vignettes ... NOTE
  '::' or ':::' import not declared from: 'BiocManager'
  'library' or 'require' call not declared from: 'BiocManager'

R/dominantTaxa.R
```diff
@@ -55,31 +41,45 @@
...
+#' @export
+setMethod("addDominantTaxa", signature = c(x = "SummarizedExperiment"),
+          function(x,
+                abund_values = "counts",
+                rank = NULL,
+                name = "dominant_taxa"){
+              dominantTaxa(x, 
+                           abund_values = abund_values,
+                           rank = rank,
+                           name = name)
```
FelixErnst: I am not sure, but this wasn't the idea, was it?

From the discussion I think the following requirements need to be met:
-  `dominantTaxa` returns a `DataFrame`
- `addDominantTaxa` returns the input `x`, with output of `dominantTaxa` added to `colData`

This allows the user to decided, whether he/she/they need the data separate or combined within a `SE`

PS: tests also need to be changed accordingly.

R/summaries.R
```diff
@@ -12,36 +12,54 @@
...
 #' @return
-#' For \code{getTopTaxa}: A vector of the most \code{top} abundant
+#' The \code{getTopTaxa} returns a vector of the most \code{top} abundant
 #' \dQuote{FeatureID}s
 #'
 #' @seealso
 #' \code{\link[=getPrevalence]{getPrevalentTaxa}}
 #'
-#' @name getTopTaxa
+#' @name summaries-basic
```
FelixErnst: So what is an non basic summary? Is there a plan which distinguishes basic and non-basic?

I know, that the questions might come across as rhetorical, but they are not. The aim is to get to the bottom of this and thus keep it simple. It also implies, that if there is the need to distinguish summaries and have different man pages, this is totally fine. But if it is currently not needed, there is no need for two man pages or one man page with a too complicated name.

R/summaries.R
```diff
@@ -12,36 +12,54 @@
...
 #'
 #' @author
-#' Sudarshan A. Shetty
+#' Leo Lahti, Tuomas Borman and Sudarshan A. Shetty
 #'
 #' @examples
-#' data(GlobalPatterns)
+#' data("GlobalPatterns")
```
FelixErnst: I changed this to match the majority of uses.

`data` also accepts the name of the dataset, and does not require a `character` as input. More info: `?data`

R/summaries.R
```diff
@@ -87,3 +105,231 @@ setMethod("getTopTaxa", signature = c(x = "SummarizedExperiment"),
...
+#' @export
+setMethod("summarizeSE",
+          signature = c(x = "SummarizedExperiment"),
+          function(x, abund_values = "counts"){
+            .check_rel_neg(x, abund_values)
+            sample.summary <- .get_summary_col_data(x, abund_values)
+            feature.summary <- .get_summary_rowl_data(x, abund_values)
+            return(list("samples" = sample.summary, "features" = feature.summary))
+          }
+)
```
FelixErnst: I got a problem with the name `summarizeSE`.

Two things I would like to throw into the mix:
1. The object class type is in the method name. This is redundant, since we are using S4 dispatches all around. Therefore, this is already done for us.
2. Is this really a summary or is a quality control function implemented here?

From these two things, I am thinking about the verb for this function. If it is really a summary, one could just the verb `summary`, for which the generic is defined in base R. If it is about QC maybe the naming scheme used elsewhere might be better followed, e.g. `perTaxaQCMetrics`/`addPerTaxaQC` and `perSampleQCMetrics`/`addPerSampleQC`

R/summaries.R
```diff
@@ -87,3 +105,231 @@ setMethod("getTopTaxa", signature = c(x = "SummarizedExperiment"),
...
+                stop("'group' variable must be in colnames(colData(x))")
+                }
+            }
+        # Adds dominant taxas to colData
+        tmp <- dominantTaxa(x, abund_values, rank, name)
+        # Gets an overview
+        overview <- .get_overview(tmp, group, name)
+        overview
+  }
+)
```
FelixErnst: This function can basically summarize anything, which is a `character`/`factor`.

If `name = "whatever"` is set, the function no longer summarizes the dominant taxa. It summarizes whatever. From this I get the feeling that the name of the function and intended purpose of this function still needs to be discovered.

From the results of the function, I think this is a fancy version of `table` or specific version `tally`. Is that correct?

From the man page of `tally`, I got the idea, that a better function name for this function might actually by `countTaxa`. What are your thought on this?

edit: Actually it doesn't do anything taxa related, since it only works on the `colData`. The taxa is stored in `rowData`. So a more accurate name would actually be `countSamples`

edit2: to match the intended use case implied here, the matching function name would/could be `countDominantTaxa`. But the implementation would need to be drastically changed, to really only allow the taxa values to be used.

R/summaries.R
```diff
@@ -87,3 +105,231 @@ setMethod("getTopTaxa", signature = c(x = "SummarizedExperiment"),
...
+
+#' @rdname summaries-basic
+#' @export
+setMethod("getUniqueTaxa",
+          signature = c(x = "SummarizedExperiment"),
+          function(x, rank = NULL){
+            .check_taxonomic_rank(rank, x)
+            unique(rowData(x)[,rank])
+          }
```
FelixErnst: Indentation needs to be fixed

FelixErnst (changes requested):
Hi @microsud, @tvborm 

I think this a good addition, but it needs some refinement, clarification.

Have a look at the comments and let me know, what's your take on those.

Felix

R/summaries.R
microsud: These are more like stats for overall denoised sequences and inferred ASVs. QC would include more information and covered by `perCellMetrics` and `perFeatureMetrics`. Would `seqeunceStats()` be a more appropriate name? 
For `perFeatureMetrics` I was thinking of expanding it to include other information shown in **2b** eg., here https://github.com/microbiome/mia/issues/69 . This will be in qc-functions.rd file.

FelixErnst: So there are two aspects these functions need to provide as far as I understand it now:
- more then QC (so the output of `perFeatureQCMetrics` does not suffice as explained by example figure in the issue)
- but the layout of the result should be along the dimension of the input

Taking this into account, the functionality would by described by functions named `perTaxaSummary` and `perSampleSummary`? (plus the additional `add*` wrappers) This would marry the idea of a summary with the granuality provided by other "Metric" functions, provide the context what is summarized and be granular enough, so that either summary could be opbtained individually.

I am not sure `seqenceStats` describes the functionality properly. What is `sequence` in this context? What would expected from a function named `seqenceStats` generally? I would guess something like GC content.

R/dominantTaxa.R
microsud: You make good points here and we need to rethink this. My idea is that,  

- `dominantTaxa` should be `perSampleDominantTaxa` and return per sample the dominant taxa, like below
image

- `addPerSampleDominantTaxa` should add the output of `perSampleDominantTaxa`  to `colData`
- While `summarizeDominantTaxa` / ` countDominantTaxa` returns a summary of dominant taxa based on its dominance in all samples 
image

or by a group of samples.

image

I think this is a more logical approach. Do you agree?

PS: looking at the outputs above we also need to think of how best to deal with NAs especially when working with taxonomic ranks from family to species level...

R/summaries.R
microsud: Agreed. I will rename this to `summaries`

R/dominantTaxa.R
FelixErnst: This is a good summary. So `dominantTaxa` should not return a `DataFrame`, but a named `character` vector. (Currently `dominantTaxa` and `addDominantTaxa` return the same result, aka. as a modified SE. So this needs to be fixed.)

I think you can add an `na.rm` argument. This is expected by the user, because thats exactly how the argument is named for other functions. `agglomerateBy*` uses it as well.

R/summaries.R
microsud: Yes. The `perTaxaSummary` will at least be a separate function together with add* to `rowData`. At the moment, I am still not sure what additional information can be provided by `perSampleSummary` that `perCellQCMetrics` from _scuttle_ doesn't provide. 

The `summarizeSE` returns the following output as a list of `tibbles`.
image
My first thought was that these are denoised sequencing stats. It should be more like `sequencingStats()` and not `sequenceStats()` but still it may not be ideal :(

edit: also I notice here the last three values for `$features` are redundant and I will remove them.

FelixErnst: `sequencingStats` implies a specific technique was used. Would this function then not be usable for array data?

`inputStats` would then be more appropriate, but it is of course logical, that the stats are computed on the input. This leaves just `stats`, which is another way of saying `summary`. With this I have brought it back to my first comment: Maybe just using the base R generic `summary` is suitable best.

microsud: Okay. So I need to do it like this?
```
setGeneric("summary",
           signature = c("x"),
           function(x, ...)
             standardGeneric("summary")
)

#' @rdname summaries-basic
#' @export
setMethod("summary",
          signature = c(x = "SummarizedExperiment"),
          function(x, abund_values = "counts"){
            .check_rel_neg(x, abund_values)
            sample.summary <- .get_summary_col_data(x, abund_values)
            feature.summary <- .get_summary_rowl_data(x, abund_values)
            return(list("samples" = sample.summary, "features" = feature.summary))
          }
)
```

FelixErnst: The first part you don't need, since the generic already exists. 

I am not sure whether you really need to add the following roxygen tag to the chunk preceeding `setMethod`:
```
#' @importFrom base summary
```
... since the base package is always present. Try and see if roxygen or a check complains.

microsud: Removed `setGeneric` and added `#' @importFrom base summary`. I get this error

```
Error in matchSignature(signature, fdef, where) : 
  there are named arguments (‘x’) in the method signature that are missing from the generic signature, for function ‘summary’
Error in unregister_namespace(package) : 
  mia is not a registered namespace.
```

FelixErnst:
```r
getGeneric("summary")
standardGeneric for "summary" defined from package "base"

function (object, ...) 
standardGeneric("summary")
<environment: 0x0000019d241e6628>
Methods may be defined for arguments: object
Use  showMethods("summary")  for currently available ones.
```

The first input has to be name `object` and the dispatch also has to be set to the argument `object`. Like this:

```r
setMethod("summary", signature = c(object = "SummarizedExperiment"),
           function(object, ...){
 # some code here
})
```

microsud: Now working atleast roxygen not showing error.

microsud:
Changes: 
`domintantTaxa()` is replaced by `perSampleDominantTaxa()` and corresponding add* function. Major change includes replacing agglomeration done manually in `domintantTaxa()` with `agglomerateByRanks()` in `perSampleDominantTaxa()`.  
`summarizeSE()` is now `summary()`.  
Tests have been changed accordingly.  
 Edit: and `summarizeDominantTaxa` is now `countDominantTaxa`

FelixErnst:
Some of the test are reporting an error(s). Can you have a look?

R/dominantTaxa.R
```diff
@@ -0,0 +1,133 @@
...
+           function(x,
+                    abund_values = "counts",
+                    rank = NULL, ...)
+               standardGeneric("perSampleDominantTaxa"))
+
+#' @rdname perSampleDominantTaxa
+#' @importFrom IRanges relist
+#' @export
+setMethod("perSampleDominantTaxa", signature = c(x = "SummarizedExperiment"),
+          function(x,
```
FelixErnst: Can you fix the indentation?

FelixErnst (changes requested):
Looks almost done. Can you fix the tests, the indentation and have a look at the internal function names? Some are not quite descriptive enough.

microsud:
made necessary changes.

FelixErnst:
Have a look at the changes I made. If you agree with these, give the go and I will merge.

microsud:
Okay from me :)

PR #106, opened 2021-06-03
Subset by prevalent/rare taxa

### LINKED ISSUE #59: Rarity indices
The microbiome pkg has a set of functions to quantify "rarity" and subset the data to rare groups. The higher the rarity, the higher the diversity. These could be therefore added to estimateDiversity but conceptually they focus on rarity and are typically not included among standard diversity indices, in this sense they may deserve their own function. 

The set of functions in microbiome pkg are as follows:

Function to identify rare taxa (complement to core taxa):
`rare_members.R`

Function to take subset of the phyloseq object that only includes rare taxa (complement to microbiome::core):
`rare.R`  

Function to calculate rarity index (could be estimateRarity in mia?)
`rarity.R`

The helper functions for specific indices:
- `log_modulo_skewness.R`  
- Could be combined: `low_abundance.R`  & `rare_abundance.R`

TuomasBorman:
Hi,

here are `subsetByPrevalentTaxa` and `subsetByRareTaxa`. Names can be easily changed if needed. I don't know which is better these or `subsetToPrevalentTaxa` and `subsetToRareTaxa`.

Equivalents:
- `microbiome::core_members` = `getPrevalentTaxa`
- `microbiome::core` = `subsetByPrevalentTaxa`
- `microbiome::rare_members` = `getRareTaxa`
- `microbiome::rare` = `subsetByRareTaxa`

Issue: https://github.com/microbiome/mia/issues/59

-Tuomas


R/getPrevalence.R
```diff
@@ -311,6 +333,63 @@ setMethod("getRareTaxa", signature = c(x = "SummarizedExperiment"),
...
+            # Agglomerates the object by rank. Taxa that do not have information
+            # at specific rank are excluded.
+            x <- agglomerateByRank(x, rank = rank, na.rm = TRUE)
+            # Changes rank to NULL, so that agglomeration is not done in next steps
+            rank <- NULL
+        }
+        # Gets the prevalent taxa
+        prevalent_taxa <- getPrevalentTaxa(x, ...)
+        # Subsets the object based on prevalent taxa
+        x <- x[prevalent_taxa]
```
FelixErnst: This works just by accident. Since an SE is a two dimensional object, you should specify the dimension explicitly.

```r
x[prevalent_taxa,]
```

btw: What happens if rownames are empty? There needs to be an alternative path for that case. This also needs to be fixed for the other three functions:

```r
> data(esophagus)
> getPrevalentTaxa(esophagus)
 [1] "59_8_22" "59_8_12" "65_3_22" "65_7_12" "65_2_17" "65_7_4"  "65_1_17" "65_8_7"  "59_5_2"  "65_6_2"  "65_4_10" "9_7_25"  "9_4_5"  
[14] "65_1_8"  "65_2_5"  "59_7_6"  "59_9_31" "59_9_18" "65_9_13" "65_3_18" "65_8_25" "59_9_17" "9_4_6"   "59_4_25" "59_5_19" "65_9_9" 
[27] "59_2_6" 
> rownames(esophagus) <- NULL
> getPrevalentTaxa(esophagus)
NULL
```

R/getPrevalence.R
```diff
@@ -311,6 +333,63 @@ setMethod("getRareTaxa", signature = c(x = "SummarizedExperiment"),
...
+            # Agglomerates the object by rank. Taxa that do not have information
+            # at specific rank are excluded.
+            x <- agglomerateByRank(x, rank = rank, na.rm = TRUE)
+            # Changes rank to NULL, so that agglomeration is not done in next steps
+            rank <- NULL
+        }
+        # Gets the rare taxa
+        rare_taxa <- getRareTaxa(x, ...)
+        # Subsets the object based on rare taxa
+        x <- x[rare_taxa]
```
FelixErnst: See above

FelixErnst (changes requested):
Hi Tuomas @tvborm 

Thanks for the PR.

See the comment for required changes. Please also include a matching test case.
Thanks!

R/getPrevalence.R
```diff
@@ -109,12 +115,20 @@
...
-#' head(taxa)
+#' prevalent <- getPrevalentTaxa(GlobalPatterns,
+#'                               rank = "Phylum",
+#'                               detection = 10,
+#'                               prevalence = 50/100,
+#'                               as_relative = FALSE)
+#' head(prevalent)
+#' 
+#' # Gets a subset of object that includes prevalent taxa
+#' x_prevalent <- subsetByPrevalentTaxa(GlobalPatterns,
```
FelixErnst: In these cases you could work with `altExp` for storing the data. That would actually good practice since the data is kept in one object and makes subsetting and other task a bit easier.

antagomir:
By the way, I remember that we previously had discussion that such subsetting functions would not be preferred since the subsetting could be done easily when the function to access taxa / sample names are available (for core, rare etc). But it seems that we are now all happy with adding this subsetting functionality?

FelixErnst:
> By the way, I remember that we previously had discussion that such subsetting functions would not be preferred since the subsetting could be done easily when the function to access taxa / sample names are available (for core, rare etc). But it seems that we are now all happy with adding this subsetting functionality?

Basically yes, you convinced me 😄 

I guess I didn't realize the use cases in which rownames are not set. In that case a function can guarantee a valid result, whereas for a user this might be not trivial without extensive examples. However, the implementation for this case has to be done, yet (@tvborm).

The easiest way to do it, would be to check for `is.null(rownames(x))` and set them to a sequential vector. Since this is shared by all function a common internal function should be used.

R/getPrevalence.R
```diff
@@ -168,6 +200,11 @@ setMethod("getPrevalence", signature = c(x = "ANY"),
             prev <- x > detection
         }
         prev <- rowSums(prev)
+        # # If prev doesn't include names / if rownames(x) was NULL
+        # if( is.null(names(prev)) ){
+        #     # Adds indices as names
+        #     names(prev) <- factor(1:length(prev))
+        # }
```
FelixErnst: Commented out?

R/getPrevalence.R
```diff
@@ -300,17 +346,81 @@ setMethod("getRareTaxa", signature = c(x = "SummarizedExperiment"),
...
-            # Gets rownames if agglomeration is not done
-            taxa <- rownames(x)
+        } else {
+            # If rownames is not NULL
+            if( !is.null(rownames(x)) ){
+                # Gets rownames 
+                taxa <- rownames(x)
+            } else {
+                # Gets indices of taxa
+                taxa <- 1:nrow(tse)
```
FelixErnst: This will be returned as warning by `BiocCheck`. Never ever use `1:x`, always use `seq_along` or `seq_len`

R/getPrevalence.R
TuomasBorman: That's leftover, I remove it

R/getPrevalence.R
FelixErnst: I think the check for `is.null(rownames(x))` should be done before the call to `getPrevalentTaxa`. That way the result is always valid and the if tree doesn't get so big. (`rownames==NULL` is pseudocode and does not work)

The rational behind it: If you can avoid putting garbage in, do it. Cleanup afterwards is always messy and bound to fail.

R/getPrevalence.R
```diff
@@ -300,17 +346,81 @@ setMethod("getRareTaxa", signature = c(x = "SummarizedExperiment"),
...
+setGeneric("subsetByPrevalentTaxa", signature = "x",
+           function(x, ...)
+               standardGeneric("subsetByPrevalentTaxa"))
+
+#' @rdname getPrevalence
+#' @export
+setMethod("subsetByPrevalentTaxa", signature = c(x = "SummarizedExperiment"),
+    function(x, rank = NULL, ...){
+        # If rank is not NULL
+        if( !is.null(rank) ){
```
FelixErnst: The check for missing rownames is not present, yet. Again, I would try putting it just before `getPrevalentTaxa`

R/getPrevalence.R
TuomasBorman: Changed

TuomasBorman: Changed

R/getPrevalence.R
TuomasBorman: Does it need check for` rownames(x)`? This calls `getPrevalentTaxa`. `getPrevalentTaxa` returns taxa names or indices depending whether `rownames(x)` is NULL or not. Because of that subsetting should work

antagomir:
> > By the way, I remember that we previously had discussion that such subsetting functions would not be preferred since the subsetting could be done easily when the function to access taxa / sample names are available (for core, rare etc). But it seems that we are now all happy with adding this subsetting functionality?
> 
> Basically yes, you convinced me smile
> 
> I guess I didn't realize the use cases in which rownames are not set. In that case a function can guarantee a valid result, whereas for a user this might be not trivial without extensive examples. However, the implementation for this case has to be done, yet (@tvborm).
> 
> The easiest way to do it, would be to check for `is.null(rownames(x))` and set them to a sequential vector. Since this is shared by all function a common internal function should be used.

Great!

R/getPrevalence.R
FelixErnst: That might be the case, but is not really a good structure. One of the mantras "fail early, fail often" can also be extended to "check early, check often"

In addition, I am not so sure, whether it really works. If `x` in `subsetByPrevalentTaxa` does not have rownames, the subsetting works "only" maybe. Even if you go in to the subsetting with "indices", as you have called them, they are actually a character vector and not a numeric vector (rownames are always `character`). Therefore the subsetting of `x` w/o rownames may only work, because the subsetting function maybe tries to convert the character vector to an integer one, but I am not so sure about it.

If you check the rownames in `subsetByPrevalentTaxa`, the rownames are set and the returned indices are actually the rownames of the object your are trying to subset. All in all, this is a much more robust solution.

And since you have to check for rownames in four functions, write an internal function for this check, so that you only have to write it once.

R/getPrevalence.R
```diff
@@ -258,8 +290,17 @@ setGeneric("getPrevalentTaxa", signature = "x",
...
+    indices  <- which(taxa)
+    # If vector contains names
+    if( !is.null(names(indices)) ){
+        # Gets the names
+        taxa <- names(indices)
+    } else {
+        # Otherwise indices are returned
+        taxa <- indices
+    }
     taxa
```
TuomasBorman: Regarding the subsetting. I'm still thinking what kind of value check of `rownames(x)` in `subsetByPrevalentTaxa` brings and also how it should be then constructed. 

This part is from `getPrevalentTaxa`. So, should this part be constructed differently? `getPrevalentTaxa` and `getRareTaxa` actually return numeric vector, when `rownames` is `NULL`. 

```
> x <- GlobalPatterns
> y <- x
> rownames(y) <- NULL
> is.numeric(getPrevalentTaxa(x))
[1] FALSE
> is.numeric(getPrevalentTaxa(y))
[1] TRUE
```

FelixErnst: That is certainly a bit weird, since both results are generated from `names(which())` once with an unnamed input and once with a named input. I first thought this was due to changes from R 4.0 to R 4.1, but apparently it is not. I have to check with release though.

FelixErnst: What is the result of `getPrevalentTaxa(y)`? It should be `NULL` since the names are not set.

TuomasBorman: This part does it (rows 294-304)

```
    # Gets indices of most prevalent taxa
    indices  <- which(taxa)
    # If vector contains names
    if( !is.null(names(indices)) ){
        # Gets the names
        taxa <- names(indices)
    } else {
        # Otherwise indices are returned
        taxa <- indices
    }
    taxa
```

TuomasBorman: `getPrevalentTaxa(y)` returns currently indices / numeric vector, but I can change that

FelixErnst: Yeah I didn't pull the commits properly. Let me check again

FelixErnst: So I think this needs some in depth refactoring. WIth all the names indices and such, it has gotten quite confusing. Give me a few minutes to come up with something

TuomasBorman: Ok. Now when I thought this thing again,  I think that I understand what was your point. (add rownames --> get the most prevalent rownames --> subset x --> remove rownames)

However, I'm not sure which is better. Currently, only difference is that `getPrevalentTaxa` returns indices instead of `NULL`. User does not get any information from `getPrevalentTaxa(x) = NULL`, but indices might give some additional value.  Are those indices misleading? That is another thing

Both options are ok to me, and I think that restructuring does not need that much of work

FelixErnst: So this was a bit tricky, but I think the clear seperation for returning indices and names is helpful. Have a look at the changes and let me know, what you think about the changes.

TuomasBorman: Yep, now it looks very good!

PR #107, opened 2021-06-10
implemented mapTaxonomy

### LINKED ISSUE #68: Map levels
We could add a function that would map between different taxonomic levels, for instance from genera to species, or from species to genera etc. (either way, from lower to higher levels and vice versa). This can be handy in many cases.

This would be a rough equivalent to [microbiome::map_levels.R](https://github.com/microbiome/microbiome/blob/master/R/map_levels.R).

FelixErnst:


antagomir (approved):

antagomir:
Just looks good to me.

PR #109, opened 2021-07-03
Added rowTree check to makePhyFromTSE

### LINKED ISSUE #108: tse to pseq works only when tse is not agglomerated
Hi,

if we first agglomerate data in a tse object to e.g. Phylum level and then try to convert to pseq, we get an error `Error in validObject(.Object): invalid class "phyloseq" object: 
Component taxa/OTU names do not match.`. The function only works when data is not agglomerated before. 

Example where tse is any tse object with e.g. a count assay:

```
tse_phylum <- agglomerateByRank(tse, rank = "Phylum")
pseq_phylum <- makePhyloseqFromTreeSummarizedExperiment(tse_phylum)
```

I had a look at the function: If I take out the function of just the first part:

```
setMethod("makePhyloseqFromTreeSummarizedExperiment",
     signature = c(x = "SummarizedExperiment"),
  function(x, abund_values = "counts")
```

and use this function on the agglomerated tse object, then it works fine. Unfortunately, I could not go further than that as I need to read on about the methods that we use in the mia package to understand what is going on...

TuomasBorman:
Hi,

now it is checked that `rowTree` matches with `rownames` in `makePhyloseqFromTreeSummarizedExperiment`.

ISSUE: https://github.com/microbiome/mia/issues/108

Branch `agglomerateTree_TRUE` is related to this issue, so it can be also deleted. 

-Tuomas

### DIFF
--- a/R/makephyloseqFromTreeSummarizedExperiment.R
+++ b/R/makephyloseqFromTreeSummarizedExperiment.R
@@ -111,12 +111,18 @@ setMethod("makePhyloseqFromTreeSummarizedExperiment",
     function(x, ...){
         
         # If rowTree exists, checks if the rowTree match with rownames:
-        # every taxa is found from tip labels and just once
-        if( !is.null(rowTree(x)) && any(!(rownames(x) %in% rowTree(x)$tip)) && 
-                     length(rowTree(x)$tip) == length(rownames(x)) ){
-            stop("rowTree does not match with rownames. 'x' can not be converted
-                 to a phyloseq object. Check rowTree(x)$tip and rownames(x).",
-                 call. = FALSE)
+        # tips labels are found from rownames
+        if( !is.null(rowTree(x)) && any(!( rowTree(x)$tip) %in% rownames(x)) ){
+            # Tries to prune the tree. If that does not work, gives an error. 
+            x <- tryCatch(
+                {addTaxonomyTree(x)},
+                error = function(cond){
+                    stop("rowTree does not match with rownames. 'x' can not be converted
+                         to a phyloseq object. Check rowTree(x)$tip and rownames(x).",
+                         call. = FALSE)
+                }
+            )
+            warning("rowTree is agglomerated to match rownames.")
         }
 
         # Gets otu_table object from the function above this, if tse contains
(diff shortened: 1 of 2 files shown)


R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -109,6 +109,21 @@ setMethod("makePhyloseqFromTreeSummarizedExperiment",
 setMethod("makePhyloseqFromTreeSummarizedExperiment",
           signature = c(x = "TreeSummarizedExperiment"),
     function(x, ...){
+        
+        # If rowTree exists, checks if the rowTree match with rownames:
+        # tips labels are found from rownames
+        if( !is.null(rowTree(x)) && any(!( rowTree(x)$tip) %in% rownames(x)) ){
+            # Tries to prune the tree. If that does not work, gives an error. 
+            x <- tryCatch(
+                {addTaxonomyTree(x)},
```
FelixErnst: Instead of pruning, this generates a totally new tree.

I think you have to extract the tree and use `ape::keep.tips` with the information provided by `rowLinks`.

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -156,7 +170,7 @@ setMethod("makePhyloseqFromTreeSummarizedExperiment",
                 args[["refseq"]] <- refseq
             }
         }
-
+
```
FelixErnst, line 164: Please remove unnecessary empty lines

FelixErnst (changes requested):
Hi Tuomas @tvborm,

thanks for the PR. Please have a look at the comments.

Felix

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Removed

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Now it works differently, indeed tree is different with `keep.tip `compared to `addTaxonomyTree`. I also removed `tryCatch`, I don't know if that's needed anymore

Issue #64, opened 2021-03-01
divergence 


antagomir:
Divergence is close to beta diversity as a concept and sometimes used to quantify community heterogeneity. For instance, two different groups of samples may be different in terms of community heterogeneity (divergence). But this has to be balanced with group sizes and against some reference (selected reference sample or community median for instance). The microbiome pkg has one simple implementation for [divergence](https://github.com/microbiome/microbiome/blob/master/R/divergence.R). We could discuss whether there is place for such function in the `mia` pkg.

Can be also used to measure divergence over time but that could be a separate issue to `miaTime`.

FelixErnst:
As far as I understand it, this should do the following:
- calculate a distance measurement select by `method` specifically in a 1:many basis
- return a numeric vector.

Is that correct?

If it is a yes, wouldn't this do the same thing? (mat = abundance matrix, x = the abundance value of a single sample)
```
mat <- cbind(matrix(x,ncol=1),mat)
dist <- calculateDistance(mat)
as.matrix(dist)[,1]
```

FelixErnst:
I think, if we want to implement this functionality, the name has to be more on point.

antagomir:
Yes, this is the point except that there could be additional arguments to tune sample sizes and possible other relevant factors (how the reference sample is chosen etc) because these may influence comparability of the divergence index between different groups. This index is correlated with sample size, for instance and therefore cannot be directly compared between groups with different sample size.

FelixErnst:
In the example given about, the assumption is made, that the reference information are provided as `x`, which can be used to connect to different sample sets. The number of samples wouldn't matter in that case, just the same number of features and by implication the same identity of features would be required.

Is that correct?

antagomir:
Yes, same features should be required so that the comparisons are meaningful.

The function itself could do subsampling for the samples to a predefined sample size (function argument).

antagomir:
This feature is less popular / less widely used and therefore not necessarily top priority.

antagomir:
Merged via #110

PR #110, opened 2021-07-03
Divergence

TuomasBorman:
Hi,

here is the implementation of divergence index. I thought that estimateDiversity would be the best place to implement it. 

Issue: https://github.com/microbiome/mia/issues/64

-Tuomas


antagomir:
This is conceptually different from alpha diversity measures as it quantifies heterogeneity within a sample collection, not within a single sample like alpha diversity measures. Therefore it is conceptually closer to beta diversity. I suggest to create a separate function, `estimateDivergence`.

R/estimateDivergence.R
```diff
@@ -0,0 +1,147 @@
...
+                    reference = "median", FUN = vegan::vegdist, method = "bray", 
+                    ...)
+             standardGeneric("estimateDivergence"))
+
+#' @rdname estimateDivergence
+#' @export
+setMethod("estimateDivergence", signature = c(x="SummarizedExperiment"),
+    function(x, abund_values = "counts", name = "divergence", 
+             reference = "median", FUN = vegan::vegdist, method = "bray", 
+             ..., BPPARAM = SerialParam()){
```
FelixErnst: `BPPARAM` is not used. Please remove the argument or use it. If you remove it, please also remove the documentation.

R/estimateDivergence.R
```diff
@@ -0,0 +1,152 @@
...
+            stop(reference_stop_msg, call. = FALSE)
+        } else {
+            if(is.numeric(reference) && length(reference) != nrow(x)){
+                stop(reference_stop_msg, call. = FALSE)
+            }
+            if(is.character(reference) && 
+               (length(reference) != 1L || !(reference %in% c("median","mean")))){
+                stop(reference_stop_msg, call. = FALSE)
+            }
+        }
```
FelixErnst: I simplified this a bit. Please check, that it still works as intended by you.

FelixErnst (changes requested):
Hi Tuomas @tvborm,

thanks for the PR. Looks good!

Have a look at the comments.

Felix

R/estimateDivergence.R
TuomasBorman: I used `estimateDiversity` as a template, so it was left there by mistake

Removed

R/estimateDivergence.R
TuomasBorman: I made some changes because boolean values went through without an error. Now it should allow only two type types of input: 

- numeric and length of `nrow`
- "median" or "mean"

PR #122, opened 2021-08-02
Added rank to agglomerated rownames

### LINKED ISSUE #120: Naming and subsetting in agglomerations (and elsewhere?)
The naming in the output of `agglomerateByRank` is not necessarily as consistent as it could be.

Problem: the examples below show differences in `rownames()` for agglomerated `TreeSE` objects, and the name format varies by agglomeration arguments. This is important because rownames can be used subsetting, for instance by using functions such as `getPrevalentTaxa` (see below). Such operations run smoothly only when the same naming formats are being used.

Possible solution: in some cases, it is difficult/impossible to avoid adding the rank name (like "Genus::Dialister"); should we consider providing rownames always in this format? In addition, there could be a wrapper that allows an easy incorporation of shorter name formats as additional columns in `rowData`. Another option is to keep everything "as is" and trust that the user can clean up names whenever needed. But this will cause extra overheads for many users.

```
library(mia)
data(GlobalPatterns)

tseA <- agglomerateByRank(GlobalPatterns, rank="Genus")
print(rownames(tseA[1:3,]))

```
[1] "Class:Thermoprotei" "Genus:Sulfolobus"   "Class:Sd-NA"  

```
tseB <- agglomerateByRank(GlobalPatterns, rank="Genus", na.rm=TRUE)
print(rownames(tseB[1:3,]))

```
[1] "Sulfolobus"     "Cenarchaeum"    "Nitrosopumilus"

```
tseC <- agglomerateByRank(GlobalPatterns, rank="Genus", agglomerateTree=TRUE)
print(rownames(tseC[1:3,]))

```
[1] "Class:Thermoprotei" "Genus:Sulfolobus"   "Class:Sd-NA"

For instance, getPrevalentTaxa returns the names in the following format and one motivation has been that this could then be used for easy subsetting.

```
  prevalent <- getPrevalentTaxa(tse,                  
                                   rank = "Phylum",
                                   detection = 10,
                                   prevalence = 50/100,
                                   as_relative = FALSE)
     head(prevalent)

```  
[1] "Crenarchaeota"  "Euryarchaeota"  "Actinobacteria" "Spirochaetes"  
[5] "Proteobacteria" "Fusobacteria"

TuomasBorman:
Hi,

now rank is added to rownames every time when data is agglomerated, e.g., "Thermoprotei" --> "Class:Thermoprotei"

Issue: https://github.com/microbiome/mia/issues/120

-Tuomas

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.1.8
+Version: 1.1.9
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
              email = "felix.gm.ernst@outlook.com",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.1.9
+Version: 1.1.8
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
              email = "felix.gm.ernst@outlook.com",
--- a/R/agglomerate.R
+++ b/R/agglomerate.R
@@ -87,6 +87,19 @@
 #' ## print the available taxonomic ranks. Shows only 1 rank available
 #' ## not useful for agglomerateByRank
 #' taxonomyRanks(enterotype)
+#' 
+#' \donttest{
+#' # Prefixes, like "Family:", can be removed from rownames, e.g, with 
+#' # stringr package
+#' library(stringr)
+#' # Get available taxonomy ranks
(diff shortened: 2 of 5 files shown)


antagomir (commented):
I would then also include an @example of shortening the names.

FelixErnst:
I think this is a solution for a non-issue (see comment on the issue). IMO a sharping of the example might be better.

antagomir:
ok

FelixErnst:
@TuomasBorman do you want to close this?

TuomasBorman:
This PR includes example of adding prefixes. If that is not necessary (everything is ok to me), then this can be closed

antagomir:
Let's merge and close.

PR #133, opened 2021-08-27
Fix bug in loadFromQIIME2()

hpages:
`loadFromQIIME2()` now checks that the sample ids in the supplied sample metadata file are compatible with those in the supplied feature table file, and align them if necessary.

This fixes https://bioconductor.org/checkResults/3.14/bioc-LATEST/mia/nebbiolo2-checksrc.html

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.1.11
+Version: 1.1.12
 Authors@R:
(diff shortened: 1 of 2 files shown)


antagomir:
Awesome, thanks!

yiluheihei:
Do we also need to check the the rownmaes of feature table to ensure it is the same with rowData (taxa data).

antagomir:
This has to be checked. Perhaps even unit tests to ensure correct function. 

Would you have a chance to look into it?

PR #135, opened 2021-08-30
Import functions fix

### LINKED ISSUE #132: Import functions need to be fixed because of SE bugfix
Hi,

bugfix of `SummarizedExperiment` https://github.com/Bioconductor/SummarizedExperiment/commit/1e1da5b79a035a0f597dedcba42d2cec7a2406c3

Before, `SE` constructor just replaced assay's colnames with coldatas rownames, assay's rownames with rowdata's rownames. But it did not order the data accordingly. Now, **`SE` constructor requires that sample and feature names are in same order as they are in rowdata and coldata.** 

Because of this, we probably need to fix some of our functions. For example,  `loadFromQIIME2` takes only file paths as an input, and user does not have possibility to adjust the order of names in assay, colData, and rowData. 

Should we include a functionality that orders the columns of assay based on coldata's rownames, and rows of assay based on rowdata's rownames?

-Tuomas

TuomasBorman:
Hi,

there was still two functions were I wasn't sure if order of names is always correct (`loadFromMothur` and `loadFromBiom`). I added checks for those functions. 

Issue: https://github.com/microbiome/mia/issues/132

In addition to that, I made some other changes. 

-Tuomas

### DIFF
--- a/R/makeTreeSummarizedExperimentFromPhyloseq.R
+++ b/R/makeTreeSummarizedExperimentFromPhyloseq.R
@@ -38,13 +38,16 @@ makeTreeSummarizedExperimentFromPhyloseq <- function(obj) {
     }
     #
     assays <- SimpleList(counts = obj@otu_table@.Data)
-    rowData <- S4Vectors:::make_zero_col_DataFrame(nrow(assays$counts))
-    colData <- S4Vectors:::make_zero_col_DataFrame(ncol(assays$counts))
+    
     if(!is.null(obj@tax_table@.Data)){
         rowData <- DataFrame(data.frame(obj@tax_table@.Data))
+    } else{
+        rowData <- S4Vectors:::make_zero_col_DataFrame(nrow(assays$counts))
     }
     if(!is.null(obj@sam_data)){
         colData <- DataFrame(data.frame(obj@sam_data))
+    } else{
+        colData <- S4Vectors:::make_zero_col_DataFrame(ncol(assays$counts))
     }
     if(!is.null(obj@phy_tree)){
         rowTree <- obj@phy_tree
@@ -57,7 +60,7 @@ makeTreeSummarizedExperimentFromPhyloseq <- function(obj) {
         referenceSeq <- NULL
     }
     TreeSummarizedExperiment(assays = assays,
-                             rowData = obj@tax_table@.Data,
+                             rowData = rowData,
                              colData = colData,
                              rowTree = rowTree,
                              referenceSeq = referenceSeq)
--- a/R/makeTreeSummarizedExperimentFromPhyloseq.R
+++ b/R/makeTreeSummarizedExperimentFromPhyloseq.R
@@ -43,11 +43,13 @@ makeTreeSummarizedExperimentFromPhyloseq <- function(obj) {
         rowData <- DataFrame(data.frame(obj@tax_table@.Data))
     } else{
         rowData <- S4Vectors:::make_zero_col_DataFrame(nrow(assays$counts))
+        rownames(rowData) <- rownames(assays$counts)
     }
     if(!is.null(obj@sam_data)){
         colData <- DataFrame(data.frame(obj@sam_data))
     } else{
         colData <- S4Vectors:::make_zero_col_DataFrame(ncol(assays$counts))
+        rownames(colData) <- colnames(assays$counts)
     }
     if(!is.null(obj@phy_tree)){
         rowTree <- obj@phy_tree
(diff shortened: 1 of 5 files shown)


R/makeTreeSummarizedExperimentFromPhyloseq.R
```diff
@@ -57,7 +62,7 @@ makeTreeSummarizedExperimentFromPhyloseq <- function(obj) {
         referenceSeq <- NULL
     }
     TreeSummarizedExperiment(assays = assays,
-                             rowData = obj@tax_table@.Data,
+                             rowData = rowData,
```
TuomasBorman, line 65: a bug

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -80,10 +80,10 @@ setMethod("makePhyloseqFromTreeSummarizedExperiment",
         args[["otu_table"]] <- otu_table
 
         # If rowData includes information
-        if(!( length(rowData(x)[,taxonomyRanks(x)]) == 0 ||
-              is.null((rowData(x)[,taxonomyRanks(x)])) )){
+        if(!( length(rowData(x)) == 0 ||
+              is.null((rowData(x))) )){
             # Gets the taxonomic data from rowData, and converts it to tax_table
-            tax_table <- as.matrix(rowData(x)[,taxonomyRanks(x)])
+            tax_table <- as.matrix(rowData(x))
```
TuomasBorman: I think phyloseq's taxonomy table is not that flexible, but still I think that we shouldn't exclude anything

FelixErnst: you have to because, a matrix can hold only one type. If you have characters and numbers in the `rowData` it will be converted to character anyway and you cannot use it as taxonomic information.

The subsetting should be used, because you don't expect anything else from a phyloseq object

FelixErnst (changes requested):
See the comment. Only with subsetting it makes sense

R/makephyloseqFromTreeSummarizedExperiment.R
TuomasBorman: Ok, in one data I had some additional information which was not ranks, so I thought that this does too much, I will revert this change

TuomasBorman: Done

TuomasBorman:
Problem was that `shared` file does not contain sample names --> name mismatch

However, those are just dummy data

TuomasBorman:
I removed the name check also from Biom function because it should have all the names in correct order

I don't have that much of experience of these different formats

R/makeTreeSummarizedExperimentFromPhyloseq.R
FelixErnst: Not a bug

FelixErnst: How does taxonomic information otherwise pass from the phylosed to TSE object? And the direct slot access wasn't my first choice, but is faster than using the accessors, because they do stuff slowly.

FelixErnst (changes requested):

R/makeTreeSummarizedExperimentFromPhyloseq.R
TuomasBorman: Ok, I thought that rowData = NULL leads to an error

I will fix that

FelixErnst: hmm. you might be right for that aspect. Can you combine the data if present?

TuomasBorman: I remembered wrongly, `colData = NULL` leads to an error, but `rowData = NULL` not

```
> library(mia)
> data("GlobalPatterns")
> tse <- GlobalPatterns
> TreeSummarizedExperiment(assays = assays(tse), rowData = NULL)
class: TreeSummarizedExperiment 
dim: 19216 26 
metadata(0):
assays(1): counts
rownames(19216): 549322 522457 ... 200359 271582
rowData names(0):
colnames(26): CL3 CC1 ... Even2 Even3
colData names(0):
reducedDimNames(0):
mainExpName: NULL
altExpNames(0):
rowLinks: NULL
rowTree: NULL
colLinks: NULL
colTree: NULL
> TreeSummarizedExperiment(assays = assays(tse), colData = NULL)
Error in `rownames<-`(`*tmp*`, value = .get_colnames_from_assays(assays)) : 
  invalid rownames length
```

We could keep it as it is in master, but in this PR data is combined if it is present. Otherwise, empty DataFrame is created. 
This slot access thing is used in that part, 

(Actually that data gathering if present was already there but it was overwritten in this `TreeSE` constructor part)

image

FelixErnst (approved):

FelixErnst:
Please update from master and bump the version.

PR #138, opened 2021-09-17
Atlas1006

antagomir:


R/calculateDMM.R
```diff
@@ -68,14 +68,12 @@
...
 #'
-#'
-#' #
 #' dmn <- calculateDMN(se)
 #' dmn[[1L]]
 #'
 #' # since this take a bit of resources to calculate for k > 1, the data is
 #' # loaded
-#' \dontrun{
+#' \donttest{
```
FelixErnst, line 76: Beaware, all these change will make testing take quite a bit longer.

FelixErnst (changes requested):
I am not sure whether I understand the changes correctly. Do they have anything to do with the name of the PR?

R/calculateDMM.R
antagomir: This was changed because this causes a NOTE in BiocCheck and change was recommended. Hence I would keep to change but we can also change back if necessary.

antagomir:
> I am not sure whether I understand the changes correctly. Do they have anything to do with the name of the PR?

Good question :+1: 

This started by adding a new data set (atlas1006) which would be useful for some development purposes in `miaTime` but it turned out that `devtools::check()` and `BiocCheck()` gave various notes and warnings, so I decided to fix those on the same go. But we can split these to two different PRs if this helps.

Although now I notice that there have been many commits before Sep 16. Not sure how. Perhaps there already was an older branch with the same name and I had missed.

If this works now we can keep, or I can create a new PR.

antagomir:
Ok I splitted this into multiple smaller PRs.

FelixErnst:
> This started by adding a new data set (atlas1006) which would be useful for some development purposes in `miaTime` [...]

I think you should put the data in MiaTime, if you need it for examples and such. `mia` wouldn't be the right place for it.

antagomir:
We can do that but this data set is not primarily a time series data set. It is mainly a cross-sectional population study (1006 individuals) where a smallish subset of around 76 individuals have a few time points. This can be useful also for other demo tasks, although the value is limited because it is based on phylogenetic microarrays which are not commonly used besides a few labs. They are an interestnig complementary technology, generating similar data, though.

PR #141, opened 2021-09-18
Atlasdata

antagomir:


FelixErnst (commented):
It is called hitchip in one place, and atlas in another. Is it the same? Is it different? Whouldn't it be best put into `microbiomeDataSets`?

Or do you need it for a function as example data, for which another dataset does not work?

antagomir:
It is already in `microbiomeDataSets` but download always takes time, and we would like to use this as a fast demodata in `miaTime` where I already have earlier examples with the same data. Speed and ability to avoid constant downloads in function examples is the reason. Using exactly this data set is useful because of the earlier material from microbiome project that has already been tuned for this one and we could easily migrate.

Yes, potentially confusing to have two names around now. But I was planning to provide an alias or change the name also elsewhere. The reason is that hitchip1006 is more specific, there are many other "atlas" data sets available from various teams and various omics.

FelixErnst:
How about subsetting to the examples with the time data and putting into `miaTIme`? That whay you can cite the reference and don't need to jump packages for the examples.

I think you should describe the differences and the relation in the NEWS in more detail

antagomir:
There are use cases where we need the combination of time series and the larger cross-sectional background collection as prior information, so subsetting does not help. But the data can be moved to miaTime and called from there in case some other packages need this for demo purposes.

PR #143, opened 2021-09-18
Biocfixes

antagomir:


FelixErnst:
You are sure about the change from `dontrun` to `donttest`? I mean bioconductor is not dogmatic about this, and it just causes a NOTE and not a WARNING.

antagomir:
> You are sure about the change from `dontrun` to `donttest`? I mean bioconductor is not dogmatic about this, and it just causes a NOTE and not a WARNING.

It is optional. Notes are just disturbing on the screen but I can live with it easily if there are reasons to ignore, and we jointly decide to do so. No strong opinions here. I am not sure if dontrun has so notable advantages that it would deserve to be kept. But if it is preferable, we can stick to it.

FelixErnst:
I would like to avoid `dontrun`, but then the examples are run and do take ages to run. For me this is a waste of time and resources, especially, if I think about the bioconductor test machines. And if we dont test what is the point in running the examples in the first place?

antagomir:
Fair enough. Let us save some CO2 and keep dontrun.

Issue #112, opened 2021-07-13
adding rarefaction (e.g. to transformCounts)


HenrikEckermann:
Hi,

I noticed that mia has already several options to transform counts. I think it would be a good addition to add to the `transformCounts` function to transform counts to rarefied counts. I could have a look at the end of the week if I am able to help with this.

antagomir:
Hi - this is useful but first let us make sure that there are no existing solutions in the (T)SE ecosystem. This is a common task and if we did not implement it for `mia` yet, it may be readily available in the SingleCellExperiment ecosystem. In that case, we could instead 1) add an example to OMA and 2) consider if we need alias of some modified capabilities for microbiome data (which I am not sure of). @FelixErnst or @microsud may have some comments here.

antagomir:
I am not aware of a ready-made implementation. Should be rather straightforward to implement and phyloseq version could provide some insipration.

microsud:
Thanks! Yes I am testing this now using `phyloseq::rarefy_even_depth`. Two things I notice:
1. Returning subsampled assay to `TSE`? 
    If from N samples in `TSE`, 1 sample is removed because it has reads less than user specified **min samples size**, then the output assay has N-1 samples i.e. colnames. Adding this to the `TSE` as new "subsampled" assay returns the TSE with N-1 samples as all assays are subsetted to have same ncol(x). Thus, the original number of samples are not preserved.
2.  Returning subsampled as `altExp` to TSE? 
     In such as case we still have error because it expects the same number of columns.
```
Error in .set_internal_character(x, e, value, getfun = int_colData, setfun = `int_colData<-`,  : 
  invalid 'value' in 'altExp(<TreeSummarizedExperiment>, type="character") <- value':
  'value' should have number of columns equal to 'ncol(x)'
```

The question is: should we use the first option and warn the user that a particular sample is removed from the entire `TSE` and add details in `OMA` or should we return subsampled data inside a `MultiAssayExperiment` container? 

Another thing, this is not the same as transforming the data because rows and columns can change and therefore cannot go with `transformCounts`. I think we need a separate function `getSubsample()`.

antagomir:
This is a broad issue that also touches, for instace, transformations such as `ilr` (not yet implemented) that would return N-1 samples. 

[...]

PR #147, opened 2021-09-26
Get subsample

microsud:
This function is basically a `phyloseq::rarefy_even_depth` for TSE/SE objects. A minor difference being the removal of features (OTUs) by default instead of an optional `trimOTUs` in `phyloseq::rarefy_even_depth`. This is because there is hardly any use of keeping the features (OTUs) that are not present in any of the samples after sub-sampling. I choose the name because subsample is more generic and can be applicable to non-microbiome data as well. Suggestions are welcome regarding naming.  In this implementation, the tree is not pruned to keep features that are remaining. Should we return pruned tree?


R/getSubsample.R
```diff
@@ -0,0 +1,178 @@
...
+            # get samples with less than min number of reads
+            if(min(colSums2(assay(x, abund_values))) < min_size){
+              rmsams <- colnames(x)[colSums2(assay(x, abund_values)) < min_size]
+              if(verbose){
+                message(length(rmsams), " samples removed ",
+                        "because they contained fewer reads than `min_size`.")
+              }
+              # remove sample(s)
+              newtse <- x[, !colnames(x) %in% rmsams]
+            }
```
TuomasBorman: `seed`, `replace`, `name`, and `verbose` are missing input checks

R/getSubsample.R
```diff
@@ -0,0 +1,178 @@
...
+              # Print to screen this value
+              message("`set.seed(", seed, ")` was used to initialize repeatable random subsampling.")
+              message("Please record this for your records so others can reproduce. \n ... \n")
+            }
+            set.seed(seed)
+            # Make sure min_size is of length 1.
+            if(length(min_size) > 1){
+              warning("`min_size` had more than one value. ", 
+                      "Using only the first. \n ... \n")
+              min_size <- min_size[1]
```
TuomasBorman: Should this be just an error, because it is not sure what user has meant?

R/getSubsample.R
```diff
@@ -0,0 +1,178 @@
...
+              newtse <- x[, !colnames(x) %in% rmsams]
+            }
+            newassay <- apply(assay(newtse, abund_values), 2, 
+                              .subsample_assay,
+                              min_size=min_size, replace=replace)
+            rownames(newassay) <- rownames(newtse)
+            # remove features not present in any samples after subsampling
+            message(paste(length(which(rowSums2(newassay) == 0)), "features", 
+                          "removed becasue they are not present in all samples", 
+                          "after subsampling.\n"))
```
TuomasBorman: typo: becasue

R/getSubsample.R
TuomasBorman: `newtse` is created inside `if `--> when there are no samples that have libraries lower than threshold, `newtse` is not created --> error in line 127

TuomasBorman (changes requested):
Hi! 

Thanks, I found some things that are worth to check

-Tuomas

R/getSubsample.R
microsud: Thank for catching this.

R/getSubsample.R
microsud: okay will change this.

R/getSubsample.R
microsud: These have default values but yes better to have these checks.

R/getSubsample.R
```diff
@@ -0,0 +1,201 @@
...
+#' @return a \code{SummarizedExperiment} or \code{TreeSummarizedExperiment} object 
+#'   with subsampled data that is different from input 
+#'   \code{SummarizedExperiment} or \code{TreeSummarizedExperiment}
+#' 
+#' @author Sudarshan A. Shetty 
+#' 
+#' @name getSubsample
+#'  
+#' @examples
+#' # When samples in TSE are less than specified min_size, they will be removed.
```
antagomir: i suggest to use abbreviation TreeSE systematically?

antagomir (commented):
In my opinion, getSubsample is a suboptimal name because it is easily confused with subsetting the samples (rather than reads).

How about `subsampleReads`? This would still be of form verbObject. 

We could also have an alias, for instance `rarifyReads`? To make this more explicit for those who are familiar with that term.

R/getSubsample.R
microsud: Would `subsampleCounts` be an option? I remember in some previous discussion we had avoided `reads` as it may not be that ideal? `rarifyReads`/`rarifyCounts` as alias is good for microbiome researchers.

antagomir: Ok with `subsampleCounts` and `rarifyCounts`

microsud:
If there are no other comments, should I merge this PR?

antagomir:
Ok to me but might be good to wait for @FelixErnst view

FelixErnst:
See my comment on the issue. I don't like the concept, but it might be necessary.

For what workflow dou you need it? Can it be kept internal?

microsud:
Yes highly debated aspect. At the moment we are supporting traditional diversity measures and workflow in `mia` and OMA. For this, we many need to have support for subsampling. Else we direct users to use `DivNet` instead of `mia` for diversity estimations.

FelixErnst (changes requested):
Hi @microsud 

Please fix the indentation, add a warning by default about the dangerous nature of the function and output the subsampled counts as TSE with the original counts as a `altExp`.

I think going full MAE is not necessary here, because samples are not dropped, are they?

Felix

microsud:
Okay. Where the `min_size` is above the total reads in some samples, these will be dropped. Having `altExp` when they are not dropped and new `TreeSE`, when they are dropped, can be confusing. Therefore I think only one option i.e. new giving `TreeSE` is convenient.

antagomir:
Samples are dropped when their read count is below the rarification depth. This is relatively common thing that happens in rarification that a few samples become discarded this way. But they often have remarkably lower (10x, 100x) read counts than other samples which may also indicate other quality issues. 

The option is to either return a new TreeSE, or a MAE object that includes both old and new data.

microsud:
> The option is to either return a new TreeSE, or a MAE object that includes both old and new data.

New `TreeSE` is better. `MAE` can be left up to the user. We can demonstrate this in the `OMA` 
```
library(mia)
library(MultiAssayExperiment)
data("GlobalPatterns")
tse <- GlobalPatterns
tse.subsampled <- subsampleCounts(tse, min_size = 60000, name = "subsampled")
mae <- MultiAssayExperiment::MultiAssayExperiment(c("originalTreeSE" = tse,
                                                    "subsampledTreeSE" = tse.subsampled))
mae
```
Output
```
A MultiAssayExperiment object of 2 listed
 experiments with user-defined names and respective classes.
 Containing an ExperimentList class object of length 2:
 [1] originalTreeSE: TreeSummarizedExperiment with 19216 rows and 26 columns
 [2] subsampledTreeSE: TreeSummarizedExperiment with 12459 rows and 25 columns
Functionality:
 experiments() - obtain the ExperimentList instance
 colData() - the primary/phenotype DataFrame
 sampleMap() - the sample coordination DataFrame
 `$`, `[`, `[[` - extract colData columns, subset, or experiment
 *Format() - convert into a long or wide DataFrame
 assays() - convert ExperimentList to a SimpleList of matrices
 exportClass() - save all data to files
```
The user then for e.g. calculate diversity like this.
```
estimateDiversity(experiments(mae)$subsampledTreeSE, 
                  abund_values = "subsampled",
                  index = "shannon")
```
Gives the user the freedom to use specific `TreeSE` for specific analysis.

_Originally posted by @microsud in https://github.com/microbiome/mia/issues/112#issuecomment-928900367_

FelixErnst:
I think there should be a switch to allow the user to choose, what happens. Depending on that either return a TSE or MAE

microsud:
Okay then should modified TreeSE (incase sample is removed) be the default with option `return_mae=FALSE`? 
Including this switch will also lead to the addition of `MultiAssayExperiment` to imports. I guess this should not be an issue.

antagomir:
`MAE` can be added to imports I think.

The argument could be also `return_type="TreeSummarizedExperiment"` or something like that, in case we like to include something else in the future besides TreeSE and MAE?

R/subsampleCounts.R
```diff
@@ -0,0 +1,257 @@
+#' Subsample Counts in a \code{SummarizedExperiment} or \code{TreeSummarizedExperiment} object
+#' 
+#' \code{subsampleCounts} will randomly subsample counts in SE/TreeSE to return a SE/TreeSE in
+#' which each sample has same number of total observations/counts/reads.
```
TuomasBorman: "...return a SE/TreeSE,,," --> output types are MAE and TreeSE

R/subsampleCounts.R
```diff
@@ -0,0 +1,257 @@
...
+           seed = runif(1, 0, .Machine$integer.max), replace = TRUE, 
+           name = "subsampled", verbose = TRUE, 
+           return_type=c("TreeSummarizedExperiment", "MultiAssayExperiment"), ...){
+    warning("Subsampling/Rarefying may undermine downstream analyses",
+            "\nand have unintended consequences. Therefore, make sure",
+            "\nthis normalization is appropriate for your data. \n ",
+            call. = FALSE)
+    # Input check
+    .check_assay_present(abund_values, x)
+
```
TuomasBorman: Only counts can be used. I tested this with `relabundance` and ended up to an error. 

Should we help user here, because this is quite clear case; only integers will work?

E.g. in addition to that existing test, 
```
if( any(assay(x, abund_values) %% 1 != 0) ){
  warning("assay contains non-integer values. Only counts table is applicable, or something...")
} 
```

R/subsampleCounts.R
```diff
@@ -0,0 +1,257 @@
...
+            "\nthis normalization is appropriate for your data. \n ",
+            call. = FALSE)
+    # Input check
+    .check_assay_present(abund_values, x)
+    
+    # check return_type
+    if(!return_type[1] %in% c("TreeSummarizedExperiment", "MultiAssayExperiment")){
+      message("`return_type` must be either `TreeSummarizedExperiment`",
+              "or `MultiAssayExperiment`")
+    }
```
TuomasBorman: This leads to a warning by default
```
2: In if (return_type == "TreeSummarizedExperiment") { :
  the condition has length > 1 and only the first element will be used
```
So I am wondering if this should this be done differently

E.g., only one value would be listed in `setGenerig/setMethod`, e.g., `return_type = TreeSummarizedExperiment`

R/subsampleCounts.R
```diff
@@ -0,0 +1,257 @@
...
+    # Make sure min_size is of length 1.
+    if(length(min_size) > 1){
+      stop("`min_size` had more than one value. ", 
+           "Specifiy a sinlge integer value. \n ... \n")
+      min_size <- min_size[1]	
+      }
+    if(min_size <= 0){
+      stop("min_size less than or equal to zero. ", 
+           "Need positive sample size to work.")
+      }
```
TuomasBorman: TRUE goes through, so is.numeric(min_size) should be added

E.g.,

```
if( !is.numeric(min_size) || as.integer(min_size) != top && min_size <= 0 ){
  stop("min_size needs to be a positive integer value.")
}
```

R/subsampleCounts.R
```diff
@@ -0,0 +1,257 @@
...
+        }
+    newassay <- apply(assay(newtse, abund_values), 2, 
+                      .subsample_assay,
+                      min_size=min_size, replace=replace)
+    rownames(newassay) <- rownames(newtse)
+    # remove features not present in any samples after subsampling
+    message(paste(length(which(rowSums2(newassay) == 0)), "features", 
+                  "removed because they are not present in all samples", 
+                  "after subsampling.\n"))
+    #keepfeatures <- rownames(newassay[(which(rowSums2(newassay) != 0)),])
```
TuomasBorman: Is this commented line still needed?

TuomasBorman (commented):
It works well

I found just just few minor things that might be worth to discuss

R/subsampleCounts.R
microsud: Something like this would work ?

```
  
    if(return_type[1] == "TreeSummarizedExperiment"){
      newtse
      } else if(return_type[1] == "MultiAssayExperiment"){
        .return_mae(x, newtse, name)
      }

```

TuomasBorman: Yes, I think this would work also

R/subsampleCounts.R
microsud: Yes, agree that we can help the user. I also mention this in the man page for `abund_values` section.

R/subsampleCounts.R
microsud: removing this.

R/subsampleCounts.R
FelixErnst: Usually you check this via `match.arg`. I don't see a point why it shouldn't apply here.

Side question: What is the difference between the return types? If it is a conveniance thing, remove it. 

edit:
it seems it is from looking at lines 190-194. Please remove it. Just return the same class as the input type and don't wrap a conversion function in this function. It is simply not on the label aka. the name of the function.

edit2: 
If you need to convert, convert to the class type which gives you the functionality you need and not the next class above it. So I guess this would be SCE and not TSE in this case.

R/subsampleCounts.R
```diff
@@ -0,0 +1,261 @@
...
+      # Print to screen this value
+      message("`set.seed(", seed, ")` was used to initialize repeatable random subsampling.")
+      message("Please record this for your records so others can reproduce. \n ... \n")
+      }
+    if(!.is_numeric_string(seed)){
+      stop("`seed` has to be an numeric value See `?set.seed`\n")
+      } 
+    if(!.is_a_bool(verbose)){
+      stop("`seed` has to be an integer value See `?set.seed`\n")
+      }
```
FelixErnst: ?

FelixErnst (changes requested):
Hi Sudarshan,

thanks for the PR.

See the comments for the requested changes 😄 

Can fix the indentation as well? Thanks.

Best
Felix

R/subsampleCounts.R
FelixErnst: I have a faint idea, that we discussed some of this in an issue. Can you link the issue to the PR, if that is the case?

microsud: #112 here we discussed this and the consensus was to either provide the `TSE` (which is the input class) or modified `TSE` within an `MAE` because rownames and colnames can differ after subsampling. I am unable to understand the rationale for returning to a different class like `SCE`. If input `TSE` has a tree, will SCE allow `rowTree` and `referenceSeq` slots?

FelixErnst: > ... the `TSE` (which is the input class) 

does it need to be? You dispatch to `SE` in line 112 so if that works, why not return the same class as the input?

If you want to have different results, based on the input class, you should the specific methods. Possibly they each wrap each other.

Back to the current state: If it is not necessary to have an MAE, than don't return it. The users can convert the object themselves. However, if you need MAE functionality always, than return always an MAE. If you need it in some cases, return it those cases and document those cases in  the man page.

But the user doesn't need to choose the output type via an argument. That can be done like this as mentioned above:

```r
subsamples_se <- subsampleCounts(se)
mae <- MultiAssayExperiment::MultiAssayExperiment(c("original" = se,"subsampled" = subsamples_se ))
```
Here the user chooses actively via the normal constructor.

microsud: This was also my initial view in  #112 
> Yes something like this...
> 
> ```
> library(mia)
> library(MultiAssayExperiment)
> data("GlobalPatterns")
> tse <- GlobalPatterns
> tse.subsampled <- subsampleCounts(tse, min_size = 60000, name = "subsampled")
> mae <- MultiAssayExperiment::MultiAssayExperiment(c("originalTreeSE" = tse,
>                                                     "subsampledTreeSE" = tse.subsampled))
> mae
> ```
> Gives the user the freedom to use specific `TreeSE` for specific analysis.

I will simplify this function then it is a straightforward implementation without the need for `altExp` or `MAE`, just a `modified-TSE` . 
I can change `signature = c(x = "SummarizedExperiment")` to `signature = c(x = "TreeSummarizedExperiment")`

FelixErnst: Did I talk in circles on the issue thread? Sorry for that

FelixErnst: > I will simplify this function then it is a straightforward implementation without the need for `altExp` or `MAE`, just a `modified-TSE` . I can change `signature = c(x = "SummarizedExperiment")` to `signature = c(x = "TreeSummarizedExperiment")`

I would keep the former. You are not subsampling the tree are you? So technically there is need for a TSE as input

microsud: hahaha we have too many PRs and discussions going on... good sign for development :)

microsud: True. I will keep `SummarizedExperiment`

microsud:
Is the indentation and implementation okay to merge?

R/subsampleCounts.R
```diff
@@ -0,0 +1,203 @@
...
+    function(x, abund_values = "counts", min_size = min(colSums2(assay(x))),
+           seed = runif(1, 0, .Machine$integer.max), replace = TRUE, 
+           name = "subsampled", verbose = TRUE, ...){
+      
+      warning("Subsampling/Rarefying may undermine downstream analyses",
+              "\nand have unintended consequences. Therefore, make sure",
+              "\nthis normalization is appropriate for your data. \n ",
+              call. = FALSE)
+      
+      .check_assay_present(abund_values, x)
```
TuomasBorman: I think the indentation is still off. If I have understood correctly, [indentation is always multiplication of 4 spaces](https://www.bioconductor.org/developers/how-to/coding-style/). 

So, line 90 is correct, it has 4 spaces. 
For example, line 94 is incorrect, it has 6 spaces. It should have 8 spaces.

R/subsampleCounts.R
```diff
@@ -0,0 +1,203 @@
...
+        stop("'name' must be a non-empty single character value and be ",
+             "different from `abund_values`.",
+             call. = FALSE)
+        }
+      set.seed(seed)
+      # Make sure min_size is of length 1.
+      if(length(min_size) > 1){
+        stop("`min_size` had more than one value. ", 
+             "Specifiy a sinlge integer value. \n ... \n")
+        min_size <- min_size[1]
```
TuomasBorman: typo

mia.Rproj
```diff
@@ -0,0 +1,17 @@
...
+UseSpacesForTab: Yes
+NumSpacesForTab: 2
+Encoding: UTF-8
+
+RnwWeave: Sweave
+LaTeX: pdfLaTeX
+
+BuildType: Package
+PackageUseDevtools: Yes
+PackageInstallArgs: --no-multiarch --with-keep.source
```
TuomasBorman: We don't have this project file currently in `mia` repo. Should this be deleted?

TuomasBorman (approved):
After this, I think it's good

R/subsampleCounts.R
microsud: Yes I see. My RStudio setting was set to default and it was changing these indentations. Now manually changed these.

R/subsampleCounts.R
microsud: fixed

mia.Rproj
microsud: Yes my gitignore file did not recognize this. I use github desktop for commits. 
Now I get this error 
image
Do you know what this is related to ?

TuomasBorman: It's related to my branch, but I'm not sure what causes that. It might be that the branch name is not supported, it includes wrong characters or something. But if this branch is still correctly working, then there should not be any problems

microsud: Oh, I see. I think it may be the '-->' you use in the name.

DESCRIPTION
```diff
@@ -75,6 +75,7 @@ Suggests:
     reldist,
     ade4,
     microbiomeDataSets,
+    MultiAssayExperiment,
```
FelixErnst, line 78: I think this is not needed anymore, is it?

R/subsampleCounts.R
```diff
@@ -0,0 +1,198 @@
...
+        # resample without replacement
+        obsvec <- mapply(rep_len, x = seq_along(x), length.out = x)
+        obsvec <- unlist(obsvec, use.names = FALSE)
+        # use `sample` for subsampling. Hope that obsvec doesn't overflow.
+        prob <- NULL
+    }
+    suppressWarnings(subsample <- sample(obsvec,
+                                         min_size,
+                                         replace = replace,
+                                         prob = prob))
```
FelixErnst, line 191: I simplified this so that only one function call has to be maintained. Can you have a look, that it works correltly? (The tests go through, but you never know)

FelixErnst (approved):

R/subsampleCounts.R
microsud: thanks!

DESCRIPTION
microsud: In vignette, we show how this function modifies input object and how the user can use `MAE`  store the two different objects.

FelixErnst:
fyi: this PR will not be in release 3.14 of mia, because it was not squashed and merged. It will be the earliest on bioc 3.15.

To many commit to pick...

PR #149, opened 2021-10-05
Added getExperimentCrossCorrelation function

### LINKED ISSUE #65: associate
The microbiome pkg has function [associate](https://github.com/microbiome/microbiome/blob/master/R/associate.R) to cross-correlate (or associate) taxonomic and other features for instance for [heatmap visualizations](https://microbiome.github.io/tutorials/Heatmap.html) of such associations. 

We could discuss whether such functionality is needed in `mia` (either based on the microbiome::associate, or entirely redesigned based on some other external packages). There are many different techniques in the R universe for such tasks but in the context of `SE` data, which is specifically suited for multi-omics / multi-source analyses, an easy access to a standard set of such associative techniques could be justified and facilitate exploration & visualization.

TuomasBorman:
Hi,

here is the implementation of `getExperimentCrossCorrelation` function. This is still a draft, because there might be some bigger things that might be worth to discuss before using more time on this. 

Basically, I have moved [`microbiome::associate`](https://rdrr.io/github/microbiome/microbiome/man/associate.html) to `mia`, and made some changes, e.g., internal functions.

We discuss that this would get `MultiAssayExperiment` as an input, but I think that we could also allow `TreeSE` + `altExp`, because some times it's better to use it than `MAE`.

Issue: https://github.com/microbiome/mia/issues/65

-Tuomas


R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
+#' Calculate cross-correlation
+#' 
+#' @param x A
+#'   \code{\link[MultiAssayExperiment:MultiAssayExperiment-class]{MultiAssayExperiment}} or
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'   object.
+#' @param experiment1 experiment 1 (MAE)
+#' @param experiment2 experiment 2 (MAE and TreeSE)
```
antagomir: x and y would be more standard input names (although less explicit as well). We could still consider using those names?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param abund_values2 A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param method association method ('pearson', or 'spearman'
+#' for continuous; categorical for discrete)
+#' @param mode Specify output format ('table' or 'matrix')
+#' @param p_adj_method p-value multiple testing correction method. 
+#' @param p_adj_threshold q-value threshold to include features
```
antagomir: adjusted p-values are not always q-values; therefore change "q-value" to "adjusted p-value"

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param method association method ('pearson', or 'spearman'
+#' for continuous; categorical for discrete)
+#' @param mode Specify output format ('table' or 'matrix')
+#' @param p_adj_method p-value multiple testing correction method. 
+#' @param p_adj_threshold q-value threshold to include features 
+#' @param n_signif mininum number of significant correlations for each 
+#' element
+#' @param cth correlation threshold to include features
```
antagomir: Could `cor_threshold`?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#'   transformed.
+#' @param method association method ('pearson', or 'spearman'
+#' for continuous; categorical for discrete)
+#' @param mode Specify output format ('table' or 'matrix')
+#' @param p_adj_method p-value multiple testing correction method. 
+#' @param p_adj_threshold q-value threshold to include features 
+#' @param n_signif mininum number of significant correlations for each 
+#' element
+#' @param cth correlation threshold to include features 
+#' @param order order the results
```
antagomir: just thinking if `sort` would be better in the end? In R sort refers to sorting the values, and order retrieves ranks of those values.

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#' @param filter_self_correlations Filter out correlations between
```
antagomir: Could there be description in @details what this means. Obviously, it only applies for the cases when experiment1==experiment2 ?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#' @param abund_values1 A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param abund_values2 A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param method association method ('pearson', or 'spearman'
+#' for continuous; categorical for discrete)
+#' @param mode Specify output format ('table' or 'matrix')
+#' @param p_adj_method p-value multiple testing correction method.
```
antagomir: Should be mentioned that this is anything from function `p.adjust` ?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#' identical items.
+#' 
+#' @details
+#' Calculate cross-correlation. ADD INFORMATION ABOUT PARAMETERS###################
+#'
+#' @references
+#' Add references here.
+#'
+#' @return 
+#' Matrix or table.
```
antagomir: I am here starting to think if we like to allow two different outputs. Another, potentially more transparent way would be to reserve this function for calculating the association matrix, and then another manipulation function that can be used to fetch differnt things from that matrix as needed (unless this is trivial based on already existing tools): melting the matrix into the long format with the appropriate field names; including filtering of the diagonal (self correlations) if necessary; or only picking up lower or upper triangular elements of the matrix. This would make the process two lines instead of one line but the separation of these steps could be useful in other ways.

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,533 @@
...
+#' @param experiment1 experiment 1 (MAE)
+#' @param experiment2 experiment 2 (MAE and TreeSE)
+#' @param abund_values1 A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param abund_values2 A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   transformed.
+#' @param method association method ('pearson', or 'spearman'
+#' for continuous; categorical for discrete)
```
antagomir: Could we allow the user to provide their own function for calculating the pairwise associations as an input argument? This would provide a very flexible way for the users to experiment with different correlation techniques while taking advantage of these manipulation tools.

antagomir (changes requested):
Thanks, and sorry for the messy legacy code..

Function name: `getExperimentCrossCorrelation`; how about making this shorter i.e. `getCrossCorrelation`? In addition, I would propose to name it `getCrossAssociation` or `getExperimentAssociations` because not all pairwise association measures are "correlations", and we may potentially like to allow users now or later use methods such as mutual information or other such stuff.

I like the idea of supporting both `altExp` as well as `MAE`.

R/getExperimentCrossCorrelation.R
TuomasBorman: Current lines 337-340

I think that could be possible

It needs some planning, because e.g. `cor.test` returns a list of values like `estimate` and `p.value` --> **other functions might have different output format and different names for correlation and p-value**


One option would be to store all the results to **table format**, then it would be possible to just use `rbind` to store next result (--> it wouldn't matter how many elements there are in the output of `correlation test function`... however, output should be a list, but usually they are in this kind of test, I guess...)

TuomasBorman: This is just for an example, but it could be like this

Adjusted p-values are calculated based on column name `p.value`. Because of the partial matching, it should match with most of the function outputs, but I am not sure, if we can ensure that. 

Additionally, I am not sure if we can take only correlation values and p-values from the function output, because some functions might have different names for them (here correlation is estimate.cor). 

We could ask user to provide them, but that would be too complicated. So, I think this kind of approach is the only way

```
> head(getExperimentCrossCorrelation(mae, method = "pearson"))
                             X                 Y        statistic.t parameter.df            p.value       estimate.cor null.value.correlation alternative
1             Actinomycetaceae Cer(d18:1/16:0).1   2.05436407129486            8 0.0740049639683336  0.587671313657527                      0   two.sided
2                   Aerococcus Cer(d18:1/16:0).1   -1.0482911483265            8  0.325139342281964 -0.347525863123677                      0   two.sided
3                    Aeromonas Cer(d18:1/16:0).1    1.3264925854383            8   0.22129190179129  0.424609009781129                      0   two.sided
4                  Akkermansia Cer(d18:1/16:0).1   1.09884358789145            8  0.303802565922371  0.362131266207358                      0   two.sided
5 Alcaligenes.faecalis.et.rel. Cer(d18:1/16:0).1    1.3116950294421            8  0.226012364044537  0.420714607178806                      0   two.sided
6           Allistipes.et.rel. Cer(d18:1/16:0).1 -0.660903568484737            8  0.527244154960545 -0.227535616145621                      0   two.sided
                                method
1 Pearson's product-moment correlation
2 Pearson's product-moment correlation
3 Pearson's product-moment correlation
4 Pearson's product-moment correlation
5 Pearson's product-moment correlation
6 Pearson's product-moment correlation
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                data.name
1                  c("sample-1" = 0, "sample-2" = 2, "sample-3" = 9, "sample-4" = 4, "sample-5" = 0, "sample-6" = 3, "sample-7" = 10, "sample-8" = 24, "sample-9" = 9, "sample-10" = 19) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
2                  c("sample-1" = 6, "sample-2" = 6, "sample-3" = 2, "sample-4" = 0, "sample-5" = 13, "sample-6" = 0, "sample-7" = 13, "sample-8" = 21, "sample-9" = 9, "sample-10" = 0) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
3                c("sample-1" = 0, "sample-2" = 16, "sample-3" = 11, "sample-4" = 18, "sample-5" = 0, "sample-6" = 0, "sample-7" = 7, "sample-8" = 0, "sample-9" = 10, "sample-10" = 11) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
4 c("sample-1" = 224, "sample-2" = 186, "sample-3" = 224, "sample-4" = 179, "sample-5" = 241, "sample-6" = 416, "sample-7" = 357, "sample-8" = 276, "sample-9" = 804, "sample-10" = 334) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
5                    c("sample-1" = 0, "sample-2" = 0, "sample-3" = 12, "sample-4" = 0, "sample-5" = 0, "sample-6" = 0, "sample-7" = 0, "sample-8" = 0, "sample-9" = 0, "sample-10" = 0) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
6    c("sample-1" = 169, "sample-2" = 107, "sample-3" = 194, "sample-4" = 216, "sample-5" = 277, "sample-6" = 312, "sample-7" = 99, "sample-8" = 123, "sample-9" = 44, "sample-10" = 86) and c("sample-1" = -0.552841968657781, "sample-2" = -0.431798275933005, "sample-3" = -0.180456064458131, "sample-4" = -0.283996656365201, "sample-5" = -0.481486060122112, "sample-6" = -0.301029995663981, "sample-7" = -0.387216143280265, "sample-8" = -0.309803919971486, "sample-9" = -0.229147988357856, "sample-10" = -0.173925197299174)
            conf.int1         conf.int2     p.adj
1 -0.0665966245540044 0.888529768771108 0.6283948
2  -0.801725471784413 0.361116951458973 0.7245048
3  -0.279827873406205 0.831845559196061 0.6556797
4  -0.346499465897828 0.807615321626787 0.7022288
5  -0.284192216903108  0.83037926750731 0.6647422
6  -0.749750809950309 0.469327700910242 0.8486179
```

TuomasBorman: I think that if we allow user to use own function, then only way to go is to just return "raw" associations as a table format.
 
Then user needs to do p-value adjustment and filtering, and additionally user needs to do matrix if it is wanted. We could provide examples though

Or then we could create a filtering function that gets information about columns as an input

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+        if( !( is.character(experiment1) && experiment1 %in% names(experiments(x)) || 
+            is.numeric(experiment1) && experiment1 <= length(experiments(x)) ) ){
+          stop("'experiment1' must be numeric or character value specifying", 
+               " experiment in experiment(x).", call. = FALSE)
+        }
+        if( !( is.character(experiment2) && experiment2 %in% names(experiments(x)) || 
+            is.numeric(experiment2) && experiment2 <= length(experiments(x)) ) ){
+          stop("'experiment2' must be numeric or character value specifying", 
+               " experiment in experiment(x).", call. = FALSE)
+        }
```
FelixErnst: Can you improve readability of the checks? Options would be formating or using a separate check functions.

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+        # If experiments are not the same, disable filter_self_correlations
+        if( experiment1 != experiment2 ){
+          filter_self_correlations <- FALSE
```
FelixErnst: You probably can also export a function called `getExperimentInterCorrelation`

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+#' @export
+setMethod("getExperimentCrossCorrelation", signature = c(x = "MultiAssayExperiment"),
+    function(x,
+             experiment1 = 1,
+             experiment2 = 2,
+             abund_values1 = "counts",
+             abund_values2 = "counts",
+             method = c("categorical", "pearson","spearman"),
+             mode = "table",
+             p_adj_method = "fdr",
```
FelixErnst: Can you include all options for the methods? This way you could check for validity of the `p_adj_method` value

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+        ############################ INPUT CHECK END ###########################
+        # Fetch assays to correlate
+        assay1 <- assay(tse1, abund_values1)
+        assay2 <- assay(tse2, abund_values1)
+        # Transposes tables to right format
+        assay1 <- t(assay1)
+        assay2 <- t(assay2)
+        # Converts tables to data frame
+        assay1 <- as.data.frame(assay1)
+        assay2 <- assay2
```
FelixErnst: ?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+################################ HELP FUNCTIONS ################################
+############################## .cor_test_data_type #############################
+.cor_test_data_type <- function(assay, method){
+  # Different available methods
+  numeric_methods <- c("spearman", "pearson")
+  categorical_methods <- c("categorical")
+  # Check if method match with values, otherwise give warning.
+  # For numeric methods, get only numeric values. For categorical methods, get only factors.
+  if (method %in% numeric_methods) {
+    inds <- vapply(assay, is.numeric, TRUE)
```
FelixErnst: `inds` the variable name is not really descriptive

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+    # If there are no numeric values, give an error
+    if( all(!inds) ){
+      stop("Assay, specified by 'abund_values', of 'experiment1' does not include",
+           " numeric values. Choose categorical method for 'method'.",
+           call. = FALSE)
+    }
+    else if (any(!inds)) {
+      warning("Considering only numeric annotations for \n
```
FelixErnst: This should be an error. I don't see an option, where a matrix (or matrix-like object) can support multiply types of values. 

```r
is.numeric(mat)
```

should be enough.

Do you have some examples of situation, where this granular check makes sense?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+    }
+  }
+  # Names of features that are numeric or factor (based on previous step)
+  names <- names(which(inds))
+  # Subset assay to get only numeric or factor values
+  if (!is.vector(assay)) {
+    assay <- suppressWarnings(as.matrix(assay[, inds], ncol=length(inds)))
+  } else {
+    assay <- as.matrix(assay[inds], ncol=length(inds))
+  }
```
FelixErnst: This should not happen. Either all in or all out, because the dimension could change in ways, which no one can anticipate

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+          warning(paste("Not enough observations (",
+                        minimum_number_of_observations, "required); \n   
+                        (", 
+                        sum(!is.na(xi)), ") \n \n 
+                        - skipping correlation estimation"))
+          res <- c(NA, NA)
+        }
+        # Return res which will be stored in jc object
+        res
+      })
```
FelixErnst: Why not do a apply(apply)) setup?

This is I think quite memory inefficient. First creating the matrix, then using a for loop without the requirement of using the value calculated for j-1.

This needs some refactoring.

Side note: What is the reason for `minimum_number_of_observations` being set to `8` in a fixed way?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+        xvec <- xvec[keep]
+        yvec <- yvec[keep]
+        
+        # Number of data-annotation samples for
+        # calculating the correlations
+        n <- sum(keep)
+        # Calculate cross-correlation using Goorma and Kruskal tau
+        correlations[varname, lev] <- .calculate_gktau(xvec, yvec) 
+      }
+    }
```
FelixErnst: See above. The for loops are not necessary. Can you benchmark with a vectorized version?

It would also be a good idea to make the functions accessible for unit testing

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+  }
+  
+  # If there are p_values that are not NA, adjust them
+  if (!all(is.na(p_values))) {
+    # Corrected p-values
+    p_values_adjusted <- array(NA, dim=dim(p_values))
+    p_values_adjusted <- matrix(p.adjust(p_values, method=p_adj_method), nrow=nrow(p_values))
```
FelixErnst: `p.adjust` can work with `NA`s. Why not use that functionality?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+                                cor_threshold, 
+                                sort,
+                                assay1, assay2, filter_self_correlations){
+  # Filter
+  if (!is.null(p_adj_threshold) || !is.null(cor_threshold)) {
+    
+    # Replace NAs with extreme values for filtering purposes
+    p_values_adjusted[is.na(p_values_adjusted)] <- 1
+    p_values[is.na(p_values_adjusted)] <- 1
+    correlations[is.na(correlations)] <- 0
```
FelixErnst: Why not just keep the NA and use them for filtering? This is not needed in my opinion

Also why calc the indices not once like this

```r
f <- is.na(p_values_adjusted)
p_values_adjusted[f] <- 1
p_values[f] <- 1
```

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+    
+    # Filter by adjusted p-values and correlations
+    inds1.q <- inds2.q <- inds1.c <- inds2.c <- NULL
+    # Which features have more adjusted p-values under the threshold than the 
+    # 'n_signif' specifies
+    if (!is.null(p_adj_threshold)) {
+      inds1.q <- apply(p_values_adjusted, 1, function(x) {
+        sum(x < p_adj_threshold) >= n_signif
```
FelixErnst: use the NA.rm functionality and add `sum(is.na(x))` instead of changing the value above.

Also you only need to calc that value only once.

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+    dimnames(p_values_adjusted) <- dimnames(p_values)
+  }
+  
+  return(list(correlations = correlations, 
+              p_values = p_values, 
+              p_values_adjusted = p_values_adjusted))
+}
+
+############################## .correlation_filter #############################
+.correlation_filter <- function(correlations,
```
FelixErnst: This function is quite long. I am not sure what it does and what steps it takes. Can you split this function into subfunctions and describe the function in a comment above?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+      inds2 <- inds2.q & inds2.c
+    } else if (is.null(p_adj_threshold) && !is.null(cor_threshold)) {
+      inds1 <- inds1.c
+      inds2 <- inds2.c
+    } else if (!is.null(p_adj_threshold) && is.null(cor_threshold)) {
+      inds1 <- inds1.q
+      inds2 <- inds2.q
+    }
+    
+    # If both features have more TRUEs than n_signif specifies /
```
FelixErnst: what does this mean? Are these the "valid" results? or not?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+                                      p_adj_threshold,
+                                      n_signif, 
+                                      cor_threshold,
+                                      sort,
+                                      assay1, assay2, filter_self_correlations)
+        # Matrix or table?
+        if (mode == "table") {
+          result <- .correlation_matrix_to_table(result)
+        }
+        return(result)
```
FelixErnst: I think you should split this into two functions:
1. getExperimentCrossCorrelation
2. testForExperimentCrossCorrelation

because the functions has no way of returning the correlation value itself.

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,577 @@
...
+             abund_values1 = "counts",
+             abund_values2 = "counts",
+             method = "spearman",
+             mode = "table",
+             p_adj_method = "fdr",
+             p_adj_threshold = 0.05,
+             n_signif = 0,
+             cor_threshold = NULL,
+             sort = FALSE,
+             filter_self_correlations = FALSE){
```
FelixErnst: use `...` for most of the arguments. No point in having to maintain all the args twice, when the are not used in the wrapper.

FelixErnst (changes requested):
Hi @TuomasBorman 

thanks for the PR.

See the comments for suggestions. I think a bit more conceptual work might be good, to iron out the usefullness of the functions. Especially the siloed approach (not having the option to return the correlation values itself) lacks extendability and is probably not the leap you can take.

In addition, the basic function for a `SummarizedExperiment` seems to be missing, but the method for `TreeSummarizedExperiment` is available without using any functionality added by `SCE` or `TSE`.

Furthermore, the `run*` equivalents, whichs tore the result somehow in the input object als missing. The result should be storable as `reducedDims`, which would allow plotting functions to immediately be usable. For `MAE` this might be more tricky, but a I think you could return the correlation table as column at least.

edit:

Also the use of `altExp`is a bit rough around the edges, since the use is only base on the fact of having second experiment or not. The following options are all valid:
1. assay1 and assay2 from one SE
2. assay1 and assay2 from two SE (This includes altExp() because this is the same case)
3. and 4. Both options but based with an MAE

I think you should, without writting any code, figure out a way, how this can be easily, understandibly mapped with as few arguments as possible. The functionallity can then fall into place as needed, because wrappers-galore and S4 dispatch can take over. Was this understandable?

R/getExperimentCrossCorrelation.R
antagomir: Here it might be necessary to consider asymmetric cases where association(a, b) != association(b, a) - this can occur with some common similarity measures, for instance KL divergence that is sometimes used is asymmetric (Jensen-Shannon is its symmetric form). 

The way to do this to explicitly specify that argument x and y correspond to rows and cols of the association table, respectively, and the function will always return upper triangular matrix, for instance. So that it is consistent.

Or there can be an argument to choose lower or upper triangular matrix. It can have one of these as the default value. Then association(a, b, upper=TRUE) should be equivalent to association(b, a, upper=FALSE). But this might be more confusing.

R/getExperimentCrossCorrelation.R
TuomasBorman: No, I don't have any examples. Like you said, I think there can be only one type of data in one matrix. I simplified the test

R/getExperimentCrossCorrelation.R
TuomasBorman: This was feature from `microbiome::associate`

This get the subset of this correlation table --> subset includes only those features that have more significant correlations ( > `n_signif`)

I'm not sure if this is useful, because there can be very strong correlation between individual `feature1` and individual `feature2`, but it can be subsetted off, because of this threshold. 

However, with this it is possible to find features that correlates more holistically with other feature set

I am not sure if we should implement this `n_signif`. This part could be just "get only those features that have at least one significant correlation"

Any ideas @antagomir ?

antagomir: I agree, just filtering with one / none basis should be sufficient. But the ability to filter out the rows and cols with no significant findings is often helpful for visualization.

R/getExperimentCrossCorrelation.R
TuomasBorman: I done this with apply+apply

I think `minimum_number_of_observations` was quite arbitrary, so I removed it

FelixErnst: Now you can simplify this even futher like this:

```r
if(non-categorical){
  FUN1 <- function()
  FUN2 <- function()
} else {
  FUN1 <- function()
  FUN2 <- function()
}
apply(x,2,FUN1, FUN_internal = FUN2)
```

In this case FUN1 has to support the FUN_internal argument.

In addition you can try, not splitting the matrix with apply, but iterating over the matrices with indices, like this:

```r
i <- seq_len(ncol(assay1))
j <- seq_len(ncol(assay2))
lapply(i,function(){
lapply(j,function(){
})
})
``` 

Since subsetting is faster in R, than splitting the object, this approach can be faster some situations. The only way to find out is benchmarking.

FelixErnst:
Additional note: Please fix the indentation. We always use 4 spaces as indentation in the other files.

R/getExperimentCrossCorrelation.R
FelixErnst: So you mean because, we always use colwise-correlation in `getExperimentCrossCorrelation`, this would not be as clear for `getExperimentInterCorrelation`?

That could be an argument I guess. Maybe a name like `getSamplesInterCorrelation` would solve the issue.

antagomir: Correlations can be only calculated over samples because only there the tables are matched (same samples, different features). In principle, correlations over features could be done as well if the features were matched. But the real cases are possibly rare, so it is correlations between each pair of features, over their matched samples. 

Therefore I am not sure if if should be `getFeatureInterCorrelations` instead.

However only some of the association measures are correlations.

How about `getFeatureAssociations`? Do we need to specify if the associations are within one assay, or between assays? If the assay/altExp/MAE could be specified through the arguments, and one could also choose to do this with x vs. x (within a single assay) or x vs. y (inter correlations between different assays/altexp/mae)? It might be conceptuallly clear if we do not limit to the case where the two associated assays have to be different, user could also choose to do this for one assay against itself? Or does this add complexity rather than reduce it?

R/getExperimentCrossCorrelation.R
TuomasBorman:
```
mae <- microbiomeDataSets::peerj32()
# New implementation
start.time <- Sys.time()
# Calculate cross-correlations
result <- getExperimentCrossCorrelation(mae, method = "pearson",p_adj_threshold = 0,  cor_threshold = 0,  filter_self_correlations = FALSE, mode = "matrix")
end.time <- Sys.time()
time.taken <- end.time - start.time
time.taken
```

with my laptop old and new implementations  give quite similar results, around 15 s

But there is quite big difference: this new implementation gives table as default

TuomasBorman:
I think this function is starting slowly to get its final shape, though still there are some things to improve before polishing 

**Biggest changes and things to discuss**

1. NAMING: these are now named as `getExperimentCrossCorrelation` and `testExperimentCrossCorrelation`. Should they be named e.g. `getFeatureAssociations` and `testFeatureAssociations`?

2. Because it is necessary to loop through every feature-pair, that is the bottle neck. I tried to find faster solution, but I couldn't. 

3. Currently we have two functions. `getExperimentCrossCorrelation` calculates only correlation values by default.`testExperimentCrossCorrelation` and `getExperimentCrossCorrelation(test_significance = TRUE)` 	calculate also p-values

4. Should we call `experiment1` -> `x `and `experiment2` --> `y`. Then name of `MAE` object needs to be changed --> `mae`?

5. Input can be `MAE`, `two TreeSEs` or `TreeSE with altExp`

antagomir:
Great work - this is an important but tricky function indeed.

1. Naming: perhaps it is easiest to stick to the current names. They are also rather intuitive.
2. We can add accelerated solutions later, if identified; one thing to try is to add random subsampling of the samples that are used to count correlations; by using 10% of the samples (or another fraction drawn at random) can potentially speed up the calculations without not affecting the results remarkable; also, for symmetric measures, one should not count a~b and b~a since these are known to be identical by definition (just to check).
3. ok 
4. good to change to x, y
5. Why TreeSE and not SE? I do not think there is anything strictly specific to TreeSE here, and SE would be more general hence preferable?

TuomasBorman:
5. My mistake. There are two options: 
- `MAE`
- `SE + SE` or `SE + character specifying experiment in altExp`

NAMESPACE
```diff
@@ -212,6 +216,7 @@ importFrom(dplyr,n)
 importFrom(dplyr,rename)
 importFrom(dplyr,select)
 importFrom(dplyr,tally)
+importFrom(reshape2,acast)
```
FelixErnst: Can replicate the function with dplyr and such?
1. rehsape2 is quite old 
2. I don't want to pull in more denpendencies than need

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,653 @@
...
+    function(x, y = NULL, ...){
+        ############################## INPUT CHECK #############################
+        if( !(class(y) == "SummarizedExperiment" || 
+              class(y) == "TreeSummarizedExperiment" ||
+              (is.character(y) && y %in% names(altExps(x)))  || 
+              (is.numeric(y) && y <= length(altExps(x))) ||
+              is.null(y) ) ){
+            stop("'y' must be SE or TreeSE object, or numeric or character value specifying", 
+                 " experiment in altExps(x) or it must be NULL.", call. = FALSE)
+        }
```
FelixErnst: I don't have the faintest ides, what this check does.

Can you split it into several seperate checks?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,649 @@
...
+    # Are these two features independent or not?
+    if ( !test_significance ){
+        return(tau)
+    } 
+    # Do the Pearson's chi-squared test
+    temp <- chisq.test(x, y)
+    # Take the p-value
+    p_value <- temp$p.value
+    # Result is combination of tau and p-value
+    list(estimate = tau, p.value = p_value)
```
FelixErnst: I simplified this a bit, but it still returns two different value types. Can you stremline this, that it always return the same type?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,649 @@
...
+# Output: List of matrices (cor, p-values and adjusted p-values / matrix (cor)
+#' @importFrom reshape2 acast
+.correlation_table_to_matrix <- function(result){
+    # Correlation matrix is done from cor column, Var1 to rows and Var2 to columns
+    cor <- reshape2::acast(result, Var1 ~ Var2, value.var = "cor")
+    # Create a result list
+    result_list <- list(cor = cor)
+    # If p_values exist, then create a matrix and add to the result list
+    if( !is.null(result$pval) ){
+        pval <- reshape2::acast(result, Var1 ~ Var2, value.var = "pval")
```
FelixErnst: Can you remove the dependency by using dplyr tools?

FelixErnst (changes requested):
I simplified things a bit, but I still have some things, which need to be addressed.

R/getExperimentCrossCorrelation.R
TuomasBorman: Now it uses tools from dplyr and tidyr

R/getExperimentCrossCorrelation.R
TuomasBorman: done

NAMESPACE
TuomasBorman: --> tidyr::pivot_wider

R/getExperimentCrossCorrelation.R
TuomasBorman: Now it returns always a list

FelixErnst:
Can you fix the build error, bump the version and modify the NEWS?

R/getExperimentCrossCorrelation.R
```diff
@@ -0,0 +1,708 @@
...
+    if( !( is.character(experiment1) && experiment1 %in% names(experiments(x)) || 
+           is.numeric(experiment1) && experiment1 <= length(experiments(x)) ) ){
+        stop("'experiment1' must be numeric or character value specifying", 
+             " experiment in experiment(x).", call. = FALSE)
+    }
+    if( !( is.character(experiment2) && experiment2 %in% names(experiments(x)) || 
+           is.numeric(experiment2) && experiment2 <= length(experiments(x)) ) ){
+        stop("'experiment2' must be numeric or character value specifying", 
+             " experiment in experiment(x).", call. = FALSE)
+    }
```
FelixErnst: The error "You don't have two experiments" is missing. I think you have to include that case.

FelixErnst:
Looks good, final version bump, fix for the input and check and you can merge. Please squash and merge, because that makes my life considerably more easy, when patching the RELEASE_3_14 branch.

FelixErnst (approved):

R/getExperimentCrossCorrelation.R
TuomasBorman: You mean test if `experiment1 == experiment2`, or?  With that we would lost some nice functionality

Or do you mean if we should test if these are `NULL`? I think this already does that

TuomasBorman: Or do you mean:

```
if(is.numeric(experiment2) && experiment2 > length(experiment(x)){
stop(paste0("MAE do not include", experiment2, " experiments"))
}
```

FelixErnst: experiment1 = 1
experiment2 = 2

but suppose there is only one experiment in the `MAE`? Then this is an invalid input, but the error will be "'experiment2' must be numeric or character value specifying experiment in experiment(x).". For that case the error message isn't really helpful.

And since we are discussing this: Both ifs starting in line 236 and 241 are the same. It would be best to put them in a function and just call the function with the variable name and the variable value. The additional test mentioned above would be seperate of that.

TuomasBorman: Like this?

```
> testForExperimentCrossCorrelation(mae, experiment1 = 3)
 Error: MAE object do not include 3 experiments. 
> testForExperimentCrossCorrelation(mae, experiment1 = "aaa")
 Error: 'experiment1' must be numeric or character value specifying experiment in experiment(x). 
```

I also made a function for this test

TuomasBorman:
```
> testForExperimentCrossCorrelation(mae, experiment1 = 3)
 Error: 'experiment1' is greater than the number of experiments in MAE object. 
```

antagomir:
Awesome, well done. This was a tricky one but can prove to be very useful.

It might deserve some example use case to OMA now.

PR #150, opened 2021-10-07
Added estimateOverlap

### LINKED ISSUE #60: Overlap between samples
The [microbiome::overlap](https://github.com/microbiome/microbiome/blob/master/R/overlap.R) function returns the overlap index between communities, used in Bashan, A., Gibson, T., Friedman, J. et al. Universality of human microbial dynamics. Nature 534, 259–262 (2016) https://doi.org/10.1038/nature18301 - could be potentially added in mia as well. In a sense this is a beta diversity index but not a typical one.

TuomasBorman:
Hi,

here is the implementation of [`microbiome::overlap`](https://github.com/microbiome/microbiome/blob/master/R/overlap.R). 

**I compared `microbiome::overlap` with [`Gao's overlap`](https://github.com/Russel88/DOC/blob/master/R/DOC.do.R)**

- 	Gao returns a list of overlap and **rJSD (root Jensen-Shannon divergence)** (and a dataframe with Overlap and rJSD as vectors) / microbiome only overlap
- 	**Overlap is similar**, microbiome has only option for relabundance, Gao does not restrict this
- 	Gao does not have detection threshold

**Jensen-Shannon divergence**

- 	We have this already implemented --> `calculateJSD`
- 	rJSD in matrix format `sqrt(as.matrix(calculateJSD(tse)))`
- 	However, this method does not have detection threshold, should there be one?
- 	(Man page does not have a formula, I can add)
- 	**Is this commonly used as an overlap estimate? If so, should estimateOverlap have new parameter `index`(?) that have options overlap, JSD, and rJSD? `estimateOverlap(x, index = c("overlap", "JSD", rJSD"), detection = 0)`**

Issue: https://github.com/microbiome/mia/issues/60

-Tuomas


antagomir:
Can we view "overlap" as one type of dissimilarity matrix? If there are no relevant differences, we could implement this in the same way than `calculateJSD`, `calculateUnifrac`..  in that case the name could also be `calculateOverlap`.. ?

FelixErnst (changes requested):
Hi Tuomas,

thanks for the PR.

Can you seperate the renaming things from the implementation. I don't think there is a link between the two things, is there?

Also remember, that you have to support legacy code now. This renaming could potentially break someones code. Just add an `abund_values = exprs_values` as an argument and use the abund_values argument internally. Modify the man pages accordingly.

Best
Felix

R/calculateOverlap.R
```diff
@@ -0,0 +1,109 @@
...
+
+.calculate_overlap <- function (x, y, detection) {
+  # Take those taxa that have abundance over threshold
+  inds <- which(x > detection & y > detection)
+  x <- x[inds]
+  y <- y[inds]
+  # Overlap is the average of the sums of the values in each sample
+  overlap <- (sum(x) + sum(y))/2
+  return(overlap)
+}
```
FelixErnst: Can you fix the indentation, bump the version, amend the NEWS and merge?

FelixErnst (approved):

FelixErnst (changes requested):
Another thing: Shouldn't the function be called `estimateOverlap` as the PR title suggests?

@antagomir

FelixErnst:
Any news on this regarding the function name?

antagomir:
> Another thing: Shouldn't the function be called `estimateOverlap` as the PR title suggests?

My earlier comment: "Can we view "overlap" as one type of dissimilarity matrix? If there are no relevant differences, we could implement this in the same way than `calculateJSD`, `calculateUnifrac`.. in that case the name could also be `calculateOverlap`.. ?" -> the title suggested `estimateOverlap` but perhaps we can change this now to make it consistent with the other dissimilarity measures?

FelixErnst:
test it, bump it, NEWS add it (Think about that one Daft Punk song, when reading)

TuomasBorman:
Lol, I had to play it

TuomasBorman:
I added `runOverlap`. It stores result to `reducedDim`.

PR #156, opened 2021-10-14
mergeCols fix

### LINKED ISSUE #154: mergeCols breaks meltAssay
Hi,

after using mergeCols, the meltAssay function fails:

```
data(GlobalPatterns)
GlobalPatterns
merged <- mergeCols(GlobalPatterns,colData(GlobalPatterns)$SampleType)
merged

meltAssay(GlobalPatterns)
# works
meltAssay(merged)
# Error in as.vector(x) : no method for coercing this S4 class to a vector
```

Thanks!

TuomasBorman:
Hi,

Problem: `mergeCols` returned `SE` with `SE`s in assays. 

Solution: 
1) `sumCountsAcrossCells` is [deprecated](https://rdrr.io/github/LTLA/scuttle/man/sumCountsAcrossCells.html) --> `summarizeAssayByGroup` is now used
2) Both of them return `SE`, and there is no option to get matrix as returned value
3) Now `summarizeAssayByGroup`  is called and matrices are extracted from returned `SE`s

I also added information that values are summed up / agglomerated, so some values might lead to meaningless values.

Issue: https://github.com/microbiome/mia/issues/154

-Tuomas


antagomir (approved):

antagomir (approved):

antagomir:
I noticed that the samples of the merged object are named by sample names from the original data.

The merged values are, however, not same as the original samples and this can be misleading. They would be more intuitively named by the corresponding levels of the grouping variables (the sum is per each level). 

**Proposal** name the samples by the grouping variable levels in the merged object.
 

```
library(mia)
# Computing relative abundance
data(GlobalPatterns,package="mia")
tse <- GlobalPatterns

# Merge columns
tse2 <- mergeCols(tse, colData(tse)$SampleType)

# Calculate relabundances for the merged data;
# first we need to extract the treeSE object from tse2!
tse3 <- transformSamples(tse2, abund_values="counts", method="relabundance")

head(assay(tse3, "counts"))

```

TuomasBorman:
Yep, that is more intuitive, I changed that

R/merge.R
```diff
@@ -192,16 +196,22 @@ setGeneric("mergeCols",
...
+    assays <- S4Vectors::SimpleList(lapply(assays, FUN = function(mat, ...){
+        temp <- scuttle::summarizeAssayByGroup(mat,
+                                               ids = f,
+                                               subset.row = NULL,
+                                               subset.col = NULL,
+                                               ...)
+        # "sum" includes agglomerated (summed up) data
+        mat <- assay(temp, "sum")
+        return(mat)
+    }))
```
FelixErnst: Just call call `scuttle::summarizeAssayByGroup` directly. I think the lapply is not necessary anymore. Just add the assays from the result to `x` in line 212

FelixErnst (changes requested):
Hi Tuomas,

thanks for PR.

See the comment.

Best
Felix

FelixErnst:
Can you fix `.merge_rows` as well? Should be the same kind of situation there

edit: Just keep the check from `.calculate_merge_row_values` and check the assays before. Rename the function to `.check_assays_for_merge` or something more descriptive.

R/merge.R
TuomasBorman: image

I think that is not possible. 

- Input can be `matrix` or `SE`
- Output is `SE` that includes `assays` that are calculated from `matrix` or `assay(SE, abund_values)`. --> Only one matrix would be then agglomerated

https://rdrr.io/github/LTLA/scuttle/man/summarizeAssayByGroup.html

FelixErnst: Here is me being wrong 😄 

Can you have a look at `.merge_rows` so that it matches this implementation?

TuomasBorman: Now I have also modified `.merge.rows`

R/merge.R
```diff
@@ -147,17 +147,26 @@ setGeneric("mergeCols",
     archetype <- .norm_archetype(f, archetype)
     # merge assays
     assays <- assays(x)
-    assays <- S4Vectors::SimpleList(lapply(names(assays), .calculate_merge_row_values, 
-                                           x = x, f = f, ...))
+    assays <- S4Vectors::SimpleList(lapply(assays, FUN = function(mat, ...){
+        .check_assays_for_merge
```
FelixErnst: ? I guess the function call is incomplete

R/merge.R
```diff
@@ -143,17 +147,26 @@ setGeneric("mergeCols",
...
+        temp <- scuttle::sumCountsAcrossFeatures(mat,
```
FelixErnst: Shouldn't this be `summarizeAssayByGroup`?

FelixErnst:
This is as annoying as hell. The scuttle functions have to different interfaces...

R/merge.R
```diff
@@ -151,8 +152,8 @@ setGeneric("mergeCols",
     FUN <- function(mat, f = f, ...){
         temp <- scuttle::summarizeAssayByGroup(mat, 
                                                ids = f, 
-                                               subset_row = NULL, 
-                                               subset_col = NULL, 
+                                               subset.row = NULL, 
+                                               subset.col = NULL, 
                                                statistics = "sum",
                                                ...)
```
TuomasBorman: I think this `summarizeAssayByGroup` does this agglomeration for only column-wise, so there shoulld be transpose step?

mat --> t(mat)

R/merge.R
```diff
@@ -149,17 +149,12 @@ setGeneric("mergeCols",
     # merge assays
     assays <- assays(x)
     mapply(.check_assays_for_merge, names(assays), assays)
```
TuomasBorman: And I think this check would be good for` .merge_cols` also

R/merge.R
FelixErnst: No this is overhead. But I would have assumed, that there would have been a switch from row to col-wise summing up...

Does the function name `summarizeAssayByGroup` suggest that it can do only summarization by cols aka. cells/samples? No it does not and this is really annoying. And now the interface to `sumCountsAcrossFeatures` is slightly different (no statistics arg...) and the names are also not symmetrically. 🤬

TuomasBorman: Yep, this is little bit weird and not that user-friendly

TuomasBorman:
I think these changes look good now, and everything seems to be working

PR #161, opened 2021-11-03
getExperimentCrossCorrelation improvements

TuomasBorman:
Hi,

here are some ideas that might improve the function

1. When there are many ties and exact p-value or correlation cannot be calculated correlation functions give warnings. Because there are so many warnings and they are printed by default, that slows down the computation. 

**--> should warnings be suppressed? If so, should we still give a one warning to user about these warnings?**

2. **I think it is better that filtering is disabled by default**, i.e., p_adj_threshold = NULL.

3. Leo suggested that **the name of testForExperimentCrossCorrelation could be testExperimentCrossCorrelation**. I agree that.

4.  **I modified messages.** 

-Tuomas


antagomir:
1. Hmm yes this is a tricky one.. how about having show warnings TRUE by default but by setting if FALSE in all examples..? Then the users will not ignore warnings by accident so easily. I would not thinkg that collecting many warnings in a single warning is the way to go.
2-4: all ok

TuomasBorman:
I added `show_warnings` which is `TRUE` by default. I thought that it is best to have `suppressWarnings` as low level as possible so that other warnings are not missed. 

It could also be added with one line like `suppressWarnings( .calculate_correlations(...) ) `in the main function

FelixErnst (approved):

FelixErnst:
bump it, update the NEWS, merge it.

PR #165, opened 2021-11-05
sort return values in getTopTaxa and getUniqueTaxa

### LINKED ISSUE #163: getUniqueTaxa: return sorted?
Hi :)

Wouldn't it be convenient to sort the return value of getUniqueTaxa()? I see no advantage of keeping it unsorted? sort() would also remove the NA that can happen, which would be good too maybe?

https://github.com/microbiome/mia/blob/4a09bbed14c283dbcccf09ea21893c1c8f933f22/R/summaries.R#L135

Thanks!
Bela

TuomasBorman:
Hi,

I added a functionality that removes NAs and sort taxa in alphabetical order. 

Issue: https://github.com/microbiome/mia/issues/163

-Tuomas


R/summaries.R
```diff
@@ -291,3 +307,18 @@ setMethod("summary", signature = c(object = "SummarizedExperiment"),
              call. = FALSE)
     }
 }
+
+# Remove NAs and order in alphabetical order
+.remove_NAs_and_sort <- function(names, sort = FALSE, ...){
+    # Check sort
+    if( !(sort == TRUE || sort == FALSE) ){
```
FelixErnst: Please use `.is_a_bool` function

antagomir (commented):
Just thinking that alphabetic sorting might be slightly misleading for `getTopTaxa`. The ordering should reflect the selection criteria, the most "top" on the top of the list. Is it possible to define that, and perhaps provide both alphabetic and "top score" as the sorting criterion?

TuomasBorman:
If I read the code correctly, `getTopTaxa` returns taxa in "top score" order by default

FelixErnst:
bump it, update the NEWS, merge it

PR #171, opened 2021-11-13
Export readQZA

### LINKED ISSUE #169: Improve loadFromQIIME2
Hi,

user might have sometimes files in different formats. For example, `loadFromQIIME2` supports only qza format (sample meta in tsv). 

In this case, if user has everything but assay in qza format, he/she cannot import files with our methods, because `.read_qza` is just an internal function. (assay is always needed in importing)

I think we have two options:
1. We can export `.read_qza` (`readQZA`) so that user can import individual files and then create `SE` from them.
2. Allow other file types (csv and tsv) also. I think in this case we need to make a presumption that first column includes rownames, because the format of the file can be anything.

Option 1 is might be the bets in my opinion

-Tuomas

TuomasBorman:
Hi,

now `readQZA` is exported.

Issue: https://github.com/microbiome/mia/issues/169

-Tuomas


antagomir (approved):
Seems good to me. 

Perhaps a full example on constructing `(Tree)SE` object should be somehow linked in @seealso or added to OMA.

TuomasBorman:
That is quite short example, so I added it to @examples. Though, some kind of example could be added also to OMA

FelixErnst (approved):

FelixErnst:
The example SE shown give the folowing description:

```
se <- SummarizedExperiment(assays = list(assay), rowData = rowdata, colData = coldata)
se
#> class: SummarizedExperiment 
#> dim: 770 34 
#> metadata(0):
#> assays(1): ''
#> rownames(770): 4b5eeb300368260019c1fbc7a3c718fc
#>   fe30ff0f71a38a39cf1717ec2be3a2fc ... 98d250a339a635f20e26397dafc6ced3
#>   1830c14ead81ad012f1db0e12f8ab6a4
#> rowData names(8): Kingdom Phylum ... Species Confidence
#> colnames(34): L1S105 L1S140 ... L6S68 L6S93
#> colData names(8): barcode.sequence body.site ...
#>   reported.antibiotic.usage days.since.experiment.start

```

Are the rownames correct? They look weird to me.

https://microbiome.github.io/mia/reference/loadFromQIIME2.html

TuomasBorman:
I extracted the biom file and opened it, and it includes those rownames. Also, I opened taxonomy.tsv alone, and it includes those same rownames, so I guess they are correct

image

antagomir:
I thinks is because some systems give unique random ids for the detected ASVs. These are then subsequently assigned to actual named taxonomic groups. I have bumped into this multiple time. Should be correct.

PR #178, opened 2021-11-19
Metaphlan import function

### LINKED ISSUE #170: Metaphlan3 results loading
Hi,

I have built a Summarized Experiment from Metaphlan3 results. However, I have not seen any previous example of this so I would like to know if the approach used is correct or would you recommend a different strategy to load this metagenomic results and work with them with Mia package:

```
library(dplyr)
library(tidyr)
library("TreeSummarizedExperiment")
library("mia")
library("miaViz")
library("SingleCellExperiment")

# Select rows from metaphlan3 that reach species level (based on the taxonomy/data that AsnicarF provides in "CuratedMetagenomicData" package to create a TSE: he provides only taxa at species level with relative abundances recalculated on this new table so that each sample has a total abundance of 100 (each sample = 100) E.g: a<-tse$`2021-10-14.AsnicarF_2017.relative_abundance`@assays@data$relative_abundance)
filteredM3 <- Metaphlan3[grepl("s__", Metaphlan3$clade_name),]

# Create the taxonomy table
Tax_table <- separate(data = filteredM3, col = clade_name, into = c("Domain", "Phylum", "Class", "Order", "Family", "Genus", "Species"), sep = "\\|")
Tax_table <- Tax_table[,1:7]
Tax_table<- as.matrix(Tax_table)

# Create the counts table
counts_table <- select(filteredM3, -c("clade_name"))
counts_table <- as.matrix(counts_table)

# Check total abundance in each sample
apply(counts_table, 2, sum)
# Generate new relative abundances by dividing each value by sample
counts_table_rel_ab <- decostand(counts_table, MARGIN = 2, method = "total")

counts  <- counts_table_rel_ab   # Abundance table (e.g. ASV data; to assay data)
tax     <- Tax_table    # Taxonomy table (to rowData)
samples <- metadata # Sample data (to colData)

se <- SummarizedExperiment(assays = list(counts = counts),
                                     rowData = tax,
                                     colData = samples)
                                     #rowTree = row_tree,
                                     #rowNodeLab = row_lab,
                                     #colTree = col_tree)

```

Thank you,

Asier

TuomasBorman:
Hi,

here is an implementation of `loadFromMetaphlan`.

`clade_name` seems to be general, so we can use that. 

However, I'm not sure if there is other variations of this file type. For example, my example file included `NCBI_tax_id`. 

It does not matter if that is not present, but **if there are other columns apart from `clade_name` and `NCBI_tax_id`, they all are currently interpreted as sample names, which will cause problems**.

Can there be, e.g., `confidence` column?

Additionally, **I moved `.parse_q2_taxonomy` to utils**, because it can be used in other functions also. Because it had some QIIME2 specific functionality, I took them off and add them directly to the `loadFromQIIME2`'s internal function that called `.parse_q2_taxonomy`

Issue: https://github.com/microbiome/mia/issues/170

-Tuomas


R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,103 @@
...
+#' \dontrun{
+#' # File path
+#' file_path <- "/data/metaphlan_result.txt"
+#' # Import data
+#' se <- loadFromMetaphlan(file)
+#' se
+#' }
+#' 
+
+loadFromMetaphlan <- function(file, ...){
```
microsud: Metaphlan can give two types of output. I per sample and another merged file with all samples. This function I think uses only merged output as input. We need to specify this.

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,103 @@
...
+    if (!file.exists(file)) {
+        stop(file, " does not exist", call. = FALSE)
+    }
+    ############################## Input check end #############################
+    # Read the table
+    table <- read.table(file, header = TRUE)
+    # Subset so that only those rows are included that include all taxonomic levels
+    table <- .get_rows_that_include_lowest_level(table)
+    # Get the data that belongs to rowData
+    rowdata_columns <- c("clade_name", "NCBI_tax_id")
```
microsud: Better to include arguments for these in function. The names can be variable. Is this for version metaphlan3 or 2 ?
image
Or position based selection cols numbers one and two.

microsud (commented):
I just checking around to see if I can share a file with standard output from metaphlan. Will let you know if I get it.

R/loadFromMetaphlan.R
antagomir: There may also be differences between Metaphlan versions 1, 2, 3. Not sure, though, if all these need to tackled or only some.

antagomir:
Also remember to add example to OMA in the end.

microsud:
Thanks, Toumas! I have recently started using Metaphlan3 and have few suggestions.
A couple of things looking forward. **Metaphlan3** has a text on the first line of the file which is the name of the database used. i.e., #mpa_v30_CHOCOPhlAn_201901. We need to handle this internally.
```
read.table(file, comment.char = '#', header = TRUE) # maybe like this?
```

Secondly, we also need to allow colData and rowTree inputs as optional arguments (default=NULL). The [curatedMetagenomicData](https://waldronlab.io/curatedMetagenomicData/) which uses `TreeSE`  for outputs from **Metaphlan3** example has several slots filled. The tree can be found [here](https://github.com/biobakery/MetaPhlAn/blob/master/metaphlan/utils/mpa_v30_CHOCOPhlAn_201901_species_tree.nwk). We should not provide this tree internally but for developing this function we need to use this. 

image
  
The other argument we need is whether to return the _lowest rank_ or all the ranks as is in the input. The default can be _lowest rank_ but in some cases, the user may need the format as in input file (like using for Lefse).

microsud:
Haha 
> A couple of things looking forward. **Metaphlan3** has a text on the first line of the file which is the name of the database used. i.e., #mpa_v30_CHOCOPhlAn_201901. We need to handle this internally.
> 
> ```
> read.table(file, comment.char = '#', header = TRUE) # maybe like this?
> ```

Cool!

TuomasBorman:
> Haha
> 
> > A couple of things looking forward. **Metaphlan3** has a text on the first line of the file which is the name of the database used. i.e., #mpa_v30_CHOCOPhlAn_201901. We need to handle this internally.
> > ```
> > read.table(file, comment.char = '#', header = TRUE) # maybe like this?
> > ```
> 
> Cool!

I was little bit faster :D

TuomasBorman:
> The other argument we need is whether to return the lowest rank or all the ranks as is in the input. The default can be lowest rank but in some cases, the user may need the format as in input file (like using for Lefse).

But is the information of higher ranks already included? We can get, e.g., Genus rank information by agglomerating from Species. 

image

If those all ranks were included, then there would be same information multiple times, and we would lost some functionality of `SE`. Or did I misunderstood something?

microsud:
Yes I am also not sure how SE will deal with this. Maybe we skip this option for now and come back if there is a genuine need for keeping all ranks.

antagomir:
If it can be solved at once and case closed, this would be better.

There can be some differences in precalculated rank-specific tables and the ones that are obtained by summing up the lowest level rank on your own. This can be due to small differences in the agglomeration details. Sometimes the tables match perfectly but not always. Typically they are rather close. I think both are relevant. If the user wants to rely on the precalculated tables, this should be possible. If the user wants to have their own control over all agllomeration choices, then the lowest level rank option is good. Could we provide it as an option to read in all rank-specific tables if the user wants to do so? This does exclude the possibility that the user can later also make their own agglomerations and compare for instance. This can be useful in fact. We also have this issue popping up with other methods than Metaphlan3.

The `SE` approach is using `altExp` for this. Look for instance the `splitByRanks` function, which generates all rank-specific tables from the lowest-level rank plus taxonomic table.

TuomasBorman:
Is this better?

- Lowest rank (e.g., Species) is in `TreeSE` object
- Higher ranks are in `altExp` as `SE` objects. They are named according to their rank.
- Only the lowest rank gets the `colData` and `rowTree` if those are provided

antagomir:
Sounds fine to me for now, experience will tell if something has to be adjusted more.

TuomasBorman:
That fail is caused when building vignette

```
lines 280-283
esophagus
esophagus <- makeTreeSummarizedExperimentFromPhyloseq(esophagus)
esophagus
```

```
Error: processing vignette 'mia.Rmd' failed with diagnostics:
invalid class "LinkDataFrame" object: 
    'dim(x)' must work
--- failed re-building ‘mia.Rmd’

```
I'm not sure what causes it, because this is new error and it's not related to this PR. Also, `makeTreeSummarizedExperimentFromPhyloseq` is working

antagomir:
If nothing else helps, you can try to print diagnostic messages just before the failure. Those should be visible in the GA run.

TuomasBorman:
> If nothing else helps, you can try to print diagnostic messages just before the failure. Those should be visible in the GA run.

This error occurs inside `TreeSummarizedExperiment` constructor, it seems that it's related just to this check

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,188 @@
...
+#' altExp(tse, "Phylum")
+#' }
+#' 
+
+loadFromMetaphlan <- function(metaphlan, sample_meta = NULL, phy_tree = NULL, ...){
+    ################################ Input check ################################
+    if(!.is_non_empty_string(metaphlan)){
+        stop("'metaphlan' must be a single character value.",
+             call. = FALSE)
+    }
```
FelixErnst: Lets call the argument file, because thats what is, isn't it?

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,188 @@
...
+        stop("'sample_meta' must be a single character value or NULL.",
+             call. = FALSE)
+    }
+    if(!is.null(phy_tree) && !.is_non_empty_string(phy_tree)){
+        stop("'phy_tree' must be a single character value or NULL.",
+             call. = FALSE)
+    }
+    ############################## Input check end #############################
+    # Create a list SE objects from the Metaphlan data. All represent own taxonomic rank
+    se_objects <- .read_metaphlan_into_se_objects(metaphlan, ...)
```
FelixErnst: I would structure it differently. This has to much nesting.

How about this:
```r
.read_metaphlan(metaphlan, ...)
.check_metaphlan_data(...)
.parse_metaphlan_data(...)
.create_se_from_metaphlan(...)
```

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,188 @@
...
+    # Get the object with lowest rank
+    tse <- se_objects[[ length(se_objects) ]]
```
FelixErnst, line 96: I didn't follow the discussion. Why not just put out the most detail result possible. The rest can be done with agglomeration.

For unavaialble information just set `NA`.

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,188 @@
...
+    # Get the order
+    indices <- match(ranks, names(tables))
+    # Remove NAs which occurs if rank is not included
+    indices <- indices[!is.na(indices)]
+    # Order tables 
+    tables <- tables[indices]
+    # Create multiple SE objects at different rank from the data
+    se_objects <- lapply(tables, .create_se_from_metaphlan, ...)
+    return(se_objects)
+}
```
FelixErnst: I think you just need to create one. The rest can be done by the user.

FelixErnst (changes requested):
Looks good, but I have a few comments.

R/loadFromMetaphlan.R
TuomasBorman: @antagomir 

> There can be some differences in precalculated rank-specific tables and the ones that are obtained by summing up the lowest level rank on your own.

FelixErnst: Very weird, but OK if this is expected

antagomir: In some cases users do want to use precalculated tables for different ranks from data provider, and there can be some differences in how the are obtained from the lowest rank. Users may also like to compare these options.

In my view we need to have it clearly available as an option to use precalculated tables for different ranks if the user wants to do so. The alternative is to show how to import every rank table separately but I am not sure if this is any better than providing a quick importer.

R/loadFromMetaphlan.R
TuomasBorman: metaphlan --> file

R/loadFromMetaphlan.R
FelixErnst: Bumping this. it is shown as outdated but it isn't

TuomasBorman: Now I restructured it

Also, I added `abund_values` for selecting the name of assay (hidden with` ...` )

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,226 @@
...
+#' \code{\link[=makeTreeSummarizedExperimentFromPhyloseq]{makeTreeSummarizedExperimentFromPhyloseq}}
+#' \code{\link[=makeSummarizedExperimentFromBiom]{makeSummarizedExperimentFromBiom}}
+#' \code{\link[=makeTreeSummarizedExperimentFromDADA2]{makeTreeSummarizedExperimentFromDADA2}}
+#' \code{\link[=loadFromQIIME2]{loadFromQIIME2}}
+#' \code{\link[=loadFromMothur]{loadFromMothur}}
+#'
+#' @export
+#' @author Leo Lahti and Tuomas Borman. Contact: \url{microbiome.github.io}
+#' 
+#' @references
```
antagomir, line 46: Add reference at least to the original [MetAPhlAn3 paper](https://elifesciences.org/articles/65088#content).

R/loadFromMetaphlan.R
```diff
@@ -0,0 +1,226 @@
...
+#'
+#' @examples
+#' \dontrun{
+#' # File path
+#' file_path <- "/data/metaphlan_result.txt"
```
antagomir: If the file is not large, it could be included as example data in `inst/extdata/` and called with `system.file()`. Then the example becomes reproducible and you can remove `dontrun`

antagomir: @KatariinaParnanen has promised to provide another open demo data set for this

antagomir (commented):
All ok to me if all checks pass.

R/loadFromMetaphlan.R
TuomasBorman: I added the example file. It's not too big, it's 8 KB while there are files 2-90 KB in extdata

FelixErnst (approved):
Add entry to NEWS and then:

bump it, **squash it**, merge it

PR #180, opened 2021-11-25
taxonomicRanks --> taxonomyRanks

TuomasBorman:
Hi,

this changes all `taxonomicRanks` to `taxonomyRanks`

-Tuomas

### DIFF
--- a/R/taxonomy.R
+++ b/R/taxonomy.R
@@ -32,7 +32,7 @@
 #'   object
 #'
 #' @param rank a single character defining a taxonomic rank. Must be a value of
-#'   \code{taxonomicRanks()} function.
+#'   \code{taxonomyRanks()} function.
 #'
 #' @param empty.fields a \code{character} value defining, which values should be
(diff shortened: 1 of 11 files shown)


R/taxonomy.R
```diff
@@ -135,7 +135,6 @@ setGeneric("taxonomyRanks", signature = c("x"),
              standardGeneric("taxonomyRanks"))
 
 #' @rdname taxonomy-methods
-#' @aliases taxonomicRanks
 #'
```
TuomasBorman, line 138: It said that `taxonomyRanks` had an alias called `taxonomicRanks`, but it wasn't working. Should we have an alias?

antagomir: In this case I think it would be more clear to have one unique name. Should be checked to OMA as well.

TuomasBorman: I checked OMA, there is used only `taxonomyRanks`, not `taxonomicRanks` so that's correct

FelixErnst (approved):

PR #183, opened 2021-12-14
Build fix

TuomasBorman:
Hi,

there is a following error 

image

This does not occur locally, so it is quite weird

`esophagus` data does not include `rowData` or `colData`, so that might cause the issue

-Tuomas


TuomasBorman:
Now it seems to work. I am not sure what is the correct way to fix this. 

The error occurs in `makeTreeSummarizedExperimentFromPhyloseq`

- It occurs only in github action, and I'm not sure why
- If we fix the function, it feels like fixing a problem that we don't have

antagomir (approved):

FelixErnst (changes requested):
Did you try with devel? Since the current build errors with pkgdown misconfig, I am not sure the issue persist. Without a clear explanation there is no reason the merge this PR, is there?

I mean the renaming is fine, but switching to GlobalPatterns for the example doesn't fix anything.

Can you revert that section and edit the title of the PR to reflect the changes in the PR more clearly? Thanks

PR #184, opened 2021-12-14
Subsample counts fix

TuomasBorman:
Hi,

I found a bug.

`subsampleCounts` led to an error, when there were no samples after subsampling. 

```
library(mia)
data(GlobalPatterns)

tse.subsampled <- subsampleCounts(GlobalPatterns, 
                                  min_size = 6000000, 
                                  name = "subsampled",
                                  replace = TRUE,
                                  seed = 1938)

```

-Tuomas

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.3.9
+Version: 1.3.10
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
              email = "felix.gm.ernst@outlook.com",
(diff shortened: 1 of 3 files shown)


TuomasBorman:
Now this returns `NULL` and message, if there are no samples

antagomir (approved):
Seems like a good solution.

microsud (approved):
Good catch

FelixErnst:
The solution was a bit to much window dressing for my taste. Again the fix was 15 lines away from the cause. This is now rectified and it errors in this case.

Reporting `NULL` would mean a difference in input and output type, which is not expected I would argue, therefore the error. An another option would be just report an `SE` without any columns and let the use figure it out for themselves.

FelixErnst (approved):

antagomir (approved):

PR #187, opened 2021-12-22
Implementation draft of rowClust

### LINKED ISSUE #186: rowData Clustering
Hi,

Adding a function for clustering rowData (of a TreeSe object, e.g. taxonomic, metabolomic.. data) could be useful in analysis. For instance, performing hierarchical clustering on rowData based on correlation, then summarizing clusters by their mean abundance or sum. This could simplify analysis as it reduces the multiple statistical testing in many occasions. The following are examples of co-abundance groups "CAGs" using clustering methods for analysis:
https://www.nature.com/articles/nature11319
https://pubmed.ncbi.nlm.nih.gov/31370880/

ChouaibB:
Hi,

This is a draft implementation of the discussion at #186 
The draft function is at "R/rowClust.R"
And a corresponding simple test at "tests/testthat/test-rowClust.Rmd"


R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+                                          #    "euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski")
+                     method = "ward.D2", # c("single", "complete", "average", "mcquitty", "ward.D", "ward.D2", "centroid","median")
+                     name = "clust_id",
+                     k = NULL, # number of clusters
+                     h = NULL, # height of cut-off
+                     ... # arguments passed to stats::dist (mainly for the p arg of "minkowski")
+                     ) {
+  # Getting Assay data
+  if(abund_values %in% assayNames(tse))
+    x <- assay(tse, abund_values)
```
antagomir: indentations should multiples of 4

antagomir: I propose to run 

```
devtools::run_examples()
devtools::check()
devtools::test()
BiocCheck::BiocCheck()
```

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+  else
+    stop(paste0("Assay ",abund_values," not Found!"))
+  
+  # Calculating distance
+  if (metric %in% c("pearson", "kendall", "spearman"))
```
antagomir: I think `metric` is called `method` in `dist`. 

How about naming methods as `clustering_method` and `dissimilarity_method`?

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+    x <- as.dist(cor(t(x), method = metric))
+  else if (metric=="cosine")
+    x <- as.dist(1 - coop::cosine(t(x)))
```
antagomir: `coop` would be a new dependency. We need to consider carefully if we can add new dependencies. 

Could we instead provide an option that the user provides their own dissimilarity function as function argument, if they like to use something that is not available through the `dist` function?

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+  else if (metric %in% c("euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski"))
+    x <- stats::dist(x, method = metric, ...)
+  else
+    stop("metric not recognized! (look at methods at help(cor) or help(dist))")
+  
+  # Clustering
+  if (method %in% c("single", "complete", "average", "mcquitty", "ward.D", "ward.D2", "centroid","median"))
+    hc <- fastcluster::hclust(x, method = method)
```
antagomir: fastcluster would be a new dependency, should add `@importFrom fastcluster hclust` to roxygen headers.

tests/testthat/test-rowClust.Rmd
```diff
@@ -0,0 +1,34 @@
+# Loading example data
+
+```{r, message=FALSE, warning=FALSE}
+library(mia)
+data("GlobalPatterns")
+tse <- GlobalPatterns
+tse <- agglomerateByRank(tse, rank="Class")
+source("../../R/rowClust.R")
```
antagomir: This source line is just for testing? Should be removed from ready made version,

tests/testthat/test-rowClust.Rmd
```diff
@@ -0,0 +1,34 @@
...
+
+```{r}
+tse_CAG_sum <- mergeRows(tse, rowData(tse)$clust_id)
+head(assay(tse_CAG_sum))
+```
+
+```{r}
+tse_CAG_av <- mergeRows(tse, rowData(tse)$clust_id, average=TRUE)
+head(assay(tse_CAG_av))
+```
```
antagomir: Good solution. But what is `average=TRUE`? I did not find this from `help(mergeRows)`.

By default, I would merge rows by summing up the abundances.

antagomir (changes requested):

antagomir:
@TuomasBorman could also provide feedback.

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+                     metric = "spearman", # c("pearson", "kendall", "spearman" || 
+                                          #    "cosine" ||
+                                          #    "euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski")
+                     method = "ward.D2", # c("single", "complete", "average", "mcquitty", "ward.D", "ward.D2", "centroid","median")
+                     name = "clust_id",
+                     k = NULL, # number of clusters
+                     h = NULL, # height of cut-off
+                     ... # arguments passed to stats::dist (mainly for the p arg of "minkowski")
+                     ) {
+  # Getting Assay data
```
microsud: Good strat! Soem suggestions: Please check some other functions in `mia` on how checks are done for inputs.       
https://github.com/microbiome/mia/blob/master/R/estimateDivergence.R  
```
################### Input check ###############
        # Check abund_values
        .check_assay_present(abund_values, x)
        # Check name
        if(!.is_non_empty_character(name) || length(name) != 1L){
            stop("'name' must be a non-empty character value.",
                 call. = FALSE)
        }
```

`k` and `h` must be either numeric or NULL, if NULL or not a single numeric value it should return an error

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+    x <- assay(tse, abund_values)
+  else
+    stop(paste0("Assay ",abund_values," not Found!"))
+  
+  # Calculating distance
+  if (metric %in% c("pearson", "kendall", "spearman"))
+    x <- as.dist(cor(t(x), method = metric))
+  else if (metric=="cosine")
+    x <- as.dist(1 - coop::cosine(t(x)))
+  else if (metric %in% c("euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski"))
```
microsud: In general, in `mia` we use helper functions so that tests and checking bugs are easier. 
You can create helper functions for these options and also use `match.arg()` and  `switch()` for example used here https://github.com/microbiome/mia/blob/master/R/summaries.R

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+  else
+    stop("metric not recognized! (look at methods at help(cor) or help(dist))")
+  
+  # Clustering
+  if (method %in% c("single", "complete", "average", "mcquitty", "ward.D", "ward.D2", "centroid","median"))
+    hc <- fastcluster::hclust(x, method = method)
+  else
+    stop("Clustering method unknown! (look at help(hclust))")
+  
+  if (is.null(k) & is.null(h))
```
microsud: also check for numeric

microsud (changes requested):

R/rowClust.R
ChouaibB: sure

tests/testthat/test-rowClust.Rmd
ChouaibB: yes just for testing

tests/testthat/test-rowClust.Rmd
ChouaibB: It's the arguments passed from `mergeRows` to `sumCountsAcrossFeatures`; help(sumCountsAcrossFeatures).
Ok, should the merging be shown at examples (docs/OMA) or shall it be a separate function?

R/rowClust.R
ChouaibB: Thanks, I will take a look and fix that.

tests/testthat/test-rowClust.Rmd
antagomir: Right. Let us show the merging now as an example in OMA as well as in the manpage of `rowClust` (because it is often needed in that context).

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
+# REQUIRED PACKAGES: "fastcluster", "stats", "coop"
+rowClust <- function(tse,
+                     abund_values = "counts",
```
TuomasBorman: please, use 'x' instead of 'tse'

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+                     name = "clust_id",
+                     k = NULL, # number of clusters
+                     h = NULL, # height of cut-off
+                     ... # arguments passed to stats::dist (mainly for the p arg of "minkowski")
+                     ) {
+  # Getting Assay data
+  if(abund_values %in% assayNames(tse))
+    x <- assay(tse, abund_values)
+  else
+    stop(paste0("Assay ",abund_values," not Found!"))
```
TuomasBorman: Add `.call = FALSE`. This removes function name from error message / makes the message more clear (At least if this error would be inside internal function)

Like this:
```
stop("'rank' must be an single character value.",
                     call. = FALSE)
```

R/rowClust.R
```diff
@@ -0,0 +1,41 @@
...
+  # Calculating distance
+  if (metric %in% c("pearson", "kendall", "spearman"))
+    x <- as.dist(cor(t(x), method = metric))
+  else if (metric=="cosine")
+    x <- as.dist(1 - coop::cosine(t(x)))
+  else if (metric %in% c("euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski"))
+    x <- stats::dist(x, method = metric, ...)
+  else
+    stop("metric not recognized! (look at methods at help(cor) or help(dist))")
+
```
TuomasBorman: Add  `call. = FALSE`

TuomasBorman (changes requested):
Looks nice.  I think that after you have got this functionality working you could check other functions. You could convert this into generic version (`setGeneric() `, `setMethod() `...) and add other required info

R/rowClust.R
ChouaibB: yes

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
+#' Clusters rowData using \code{fastcluster::hclust}
```
antagomir: I think that the title row should not contain markup (not sure). If this goes through all checks then it is ok. 

You could next also run this (on command line) to check there are no errors:

```
/path/R-4.1.2/bin/R CMD build ../../  
/path/R-4.1.2/bin/R CMD check mia_x.y.z.tar.gz 
/path/R-4.1.2/bin/R CMD BiocCheck mia_x.y.z.tar.gz
/path/R-4.1.2/bin/R CMD INSTALL mia_x.y.z.tar.gz 

```

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#'
+#' @details
```
antagomir: The section after the title row should be @description and it does not need header. Check the other functions for an example. The @details tag comes later, if needed.

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' 
+#' @param transposed whether the data should be transposed.
+#' 
+#' @param dissimilarity_FUN function to compute the dissimilarity matrix
+#' 
+#' @param k number of clusters (same as in \code{\link[hclust]{hclust}}).
+#' 
+#' @param h cut-off height (same as \code{\link[hclust]{hclust}})
+#'
+#' @return \code{x} with additional \code{\link{rowData}} named \code{*name*}
```
antagomir: In general, good to indicate which type the argument should be (logical / numeric / character..). Check the other functions for examples.

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' @param clustering_method clustering method to be used, same as methods 
+#' available at \code{\link[hclust]{hclust}}.
+#' 
+#' @param transposed whether the data should be transposed.
+#' 
+#' @param dissimilarity_FUN function to compute the dissimilarity matrix
+#' 
+#' @param k number of clusters (same as in \code{\link[hclust]{hclust}}).
+#' 
+#' @param h cut-off height (same as \code{\link[hclust]{hclust}})
```
antagomir: Could we just pass these parameters to hclust via the "..." mechanism, and refer to hclust documentation?

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#'
+#' @return \code{x} with additional \code{\link{rowData}} named \code{*name*}
+#'
+#' 
+#' @name rowClust
```
antagomir: Perhaps not needed?

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' 
+#' @export
+#'
+#' @examples
+#' # Load example data
+#' library(mia)
+#' data("GlobalPatterns")
+#' tse <- GlobalPatterns
+#' # Agglomerate to Class rank
```
antagomir: mention here that this is done to speed up the example?

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' tse <- agglomerateByRank(tse, rank="Class")
+#' # Cluster using spearman correlation as a dissimilarity and ward.D2
```
antagomir: `fastcluster::hclust` default method is "complete"; I propose to keep the same defaults than in `hclust` but in the examples show how to do `ward.D2` (preferred imo).

antagomir: Spearman is written with capital S

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' @param abund_values name of the assay containing the data.
+#' 
+#' @param name name of the new column at rowData where to place cluster ids.
+#' 
+#' @param clustering_method clustering method to be used, same as methods 
+#' available at \code{\link[hclust]{hclust}}.
+#' 
+#' @param transposed whether the data should be transposed.
+#' 
+#' @param dissimilarity_FUN function to compute the dissimilarity matrix
```
antagomir: Mention here what distance function is used by default.

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' # Agglomerate to Class rank
+#' tse <- agglomerateByRank(tse, rank="Class")
+#' # Cluster using spearman correlation as a dissimilarity and ward.D2 
+#' # for clustering; with results having 6 clusters
+#' tse <- rowClust(tse, k = 6)
+#' # Using cosine dissimilarity, average clustering and 0.6 as the height
+#' # of cut-off
+#' tse <- rowClust(tse, 
+#'     dissimilarity_FUN=function(x) 1 - coop::cosine(x),
+#'     clustering_method = "average",
```
antagomir: I am not fan of "average" method in this context. Show "complete" or "ward.D2" instead

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' tse <- GlobalPatterns
+#' # Agglomerate to Class rank
+#' tse <- agglomerateByRank(tse, rank="Class")
+#' # Cluster using spearman correlation as a dissimilarity and ward.D2 
+#' # for clustering; with results having 6 clusters
+#' tse <- rowClust(tse, k = 6)
+#' # Using cosine dissimilarity, average clustering and 0.6 as the height
+#' # of cut-off
+#' tse <- rowClust(tse, 
+#'     dissimilarity_FUN=function(x) 1 - coop::cosine(x),
```
antagomir: Is this our recommended method? Which distance methods other papers have used? We can give interesting usage examples like this but we should be clear on what we think is a good default choice.

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+#' # Aggregate clusters as a sum of each cluster values
+#' # (CAG: Co-Abundance Groups)
+#' tse_CAG <- mergeRows(tse, rowData(tse)$clust_id)
+#'
+NULL
+
+#' @importFrom fastcluster hclust
+#' @importFrom stats cutree as.dist
+
+#' @rdname rowClust
```
antagomir, line 78: I think we do not need this one?

R/rowClust.R
```diff
@@ -0,0 +1,112 @@
...
+        # Calculating dissimilarity
+        if(!.is_function(dissimilarity_FUN))
+            stop("'dissimilarity_FUN' must be a function.",
+                 call. = FALSE)
+        mat <- do.call(dissimilarity_FUN, list(mat))
+        
+        # Clustering
+        method <- match.arg(clustering_method,
+                            c("single", "complete", "average", "mcquitty",
+                              "ward.D", "ward.D2", "centroid","median"))
```
antagomir: do we need this? you can use "..." to pass any argument value allowed by the target function (i.e. hclust). I am not sure how important the matching is here. Perhaps it is.

antagomir (changes requested):
Seems good. Make sure unit tests are included and R CMD tests run, too (see the comment).

R/rowClust.R
ChouaibB: https://www.nature.com/articles/nature11319 has used `Hierarchical Ward-linkage clustering based on the
Spearman correlation`.
and the second https://pubmed.ncbi.nlm.nih.gov/31370880/ `Approximate Nearest Neighbor heuristic for near-exhaustive average linkage clustering`, however very big dataset computed on a cloud.

antagomir: The first one OK for now

R/rowClust.R
antagomir: I still seem to get some warnings with `devtools::check()` and with the above script. Confirm when cleaned up.

R/rowClust.R
antagomir: I still seem to get some warnings with devtools::check() and with the above script. Confirm when cleaned up.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#' 
+#' @param transposed \code{logical} whether the data should be transposed.
+#' 
+#' @param dissimilarity_FUN \code{stats::cor} function to compute the dissimilarity matrix
+#' 
+#' @param ... other arguments passed onto \code{\link[cutree]{cutree}}
+#'
+#' @return \code{x} with additional \code{\link{rowData}} named \code{*name*}
+#'
+#' @export
```
antagomir: I think we typically do not use spaces between these lines. Check other functions, ensure uniform design,

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#' @param abund_values a \code{character} vector, name of the assay containing the data.
+#' 
+#' @param name a \code{character} vector, name of the new column at \code{rowData} where to place cluster ids.
+#' 
+#' @param clustering_method a \code{character} vector representing the clustering 
+#' method to be used, same as methods available at \code{\link[hclust]{hclust}}.
+#' 
+#' @param transposed \code{logical} whether the data should be transposed.
+#' 
+#' @param dissimilarity_FUN \code{stats::cor} function to compute the dissimilarity matrix
```
antagomir: Is stats::cor the default, or is it the only choice here. Clarify the text.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#'   object.
+#'   
+#' @param abund_values a \code{character} vector, name of the assay containing the data.
+#' 
+#' @param name a \code{character} vector, name of the new column at \code{rowData} where to place cluster ids.
+#' 
+#' @param clustering_method a \code{character} vector representing the clustering 
+#' method to be used, same as methods available at \code{\link[hclust]{hclust}}.
+#' 
+#' @param transposed \code{logical} whether the data should be transposed.
```
antagomir: Is this necessary? The assays are _always_ having "taxa x samples" format, if the SE object has been constructed correctly.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#' # Load example data
+#' library(mia)
+#' data("GlobalPatterns")
+#' tse <- GlobalPatterns
+#' # Agglomerate to Class rank; for speeding up the example.
+#' tse <- agglomerateByRank(tse, rank="Class")
+#' # Cluster using Spearman correlation as a dissimilarity and ward.D2 
+#' # for clustering; with results having 6 clusters
+#' tse <- rowClust(tse, clustering_method = "ward.D2",
+#'                 dissimilarity_FUN = function(x) stats::cor(x, method="spearman"),
```
antagomir: Important: `stats::cor` is similarity method (increases with similarity) and not dissimilarity method (decreases with similarity). The common metrics, like Euclidean distance, are dissimilarity methods. The treatment should be uniform, and here the corresponding dissimilarity function should be `1 - stats::cor`.  Fix.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#'                 k = 6)
+#' # Aggregate clusters as a sum of each cluster values
+#' # (CAG: Co-Abundance Groups)
+#' tse_CAG <- mergeRows(tse, rowData(tse)$clust_id)
+#'
+NULL
+
+#' @importFrom fastcluster hclust
```
antagomir: accoding to automated checks, this seems to conflict `importFrom stats hclust`. That is used in `getExperimentCrossCorrelation` by @TuomasBorman - I think we could use fastcluster implementation there as well. If Tuomas agrees, this can be changed as a *separate* PR.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#' @importFrom stats cutree as.dist
+
+#' @export
+setGeneric("rowClust",signature = c("x"),
+    function(x, abund_values = "counts", name = "clust_id",
+             dissimilarity_FUN = stats::cor,
```
antagomir: See above. This is not dissimilarity. Should be 1 - cor.

R/rowClust.R
```diff
@@ -0,0 +1,83 @@
...
+#' # for clustering; with results having 6 clusters
+#' tse <- rowClust(tse, clustering_method = "ward.D2",
+#'                 dissimilarity_FUN = function(x) stats::cor(x, method="spearman"),
+#'                 k = 6)
+#' # Aggregate clusters as a sum of each cluster values
+#' # (CAG: Co-Abundance Groups)
+#' tse_CAG <- mergeRows(tse, rowData(tse)$clust_id)
+#'
+NULL
+
```
antagomir, line 75: We should have @references section in roxygen, to provide original literature references. At least the two mentioned in the original issue-

antagomir (changes requested):
Looks good, almost there!

R/rowClust.R
ChouaibB: I have checked existing functions and it seems there is a space between each description (or when using @)?

R/rowClust.R
ChouaibB: From the initial commit how I tested the calculation of the dissimilarity matrix, correlation and cosine needed the assay transposed. However, wen using stats::dist() the assay didn't need to be transposed. ([link](https://github.com/microbiome/mia/commit/c4eb8e0e14771ba529281caf7c8b016349581f72)) (line 20-25).
Therefore, I thought that the user, could define how the assay should be handled (e.g. dissimilarity_FUN=function(x) cor(t(x))) , or the function would transpose it? (error will be thrown in a dim mismatch case) 
What do you think is the best?

R/rowClust.R
antagomir: Right, so it  is. Then all is good.

R/rowClust.R
antagomir: Ah - that's a tricky one. 

I feel that it would be safer if we require that the user-provided function is always assuming the same format (features x samples). Then the user is responsible for respecting these. Having an additional `transpose` argument is probably more confusing, and transpose is easy to add to user-defined function as well, if necessary.

antagomir:
Also "resolve" the comments while you go and are done with them.

R/rowClust.R
TuomasBorman: Hmmm... does this `fastcluster` give us any additional value?  `stats` is default package, so by using `stats::hclust` we do not create any extra dependencies

But if that `factcluster` is better and faster, then I'm fine with that, and we can use it also in `getExperimentCrossCorrelation`

antagomir: It is better and faster [according to their own benchmarks](https://www.jstatsoft.org/article/view/v053i09).

At least miaverse will get better in terms of CO2 emission contributions if we optimize these things..

If these two considerations are crucial enough to justify more dependencies is another question. The same solution should be used in all functions, no matter which option we end up choosing.

I see 2-5 fold speedups with some of our own demo data sets. See e.g. below. For most practical use cases this probably does not make huge difference because the computing is not prohibitively slow in either case. Hence, I have no strong opinions but perhaps a slight preference to the computationally more efficient solutions if this can be easily added.


```
library(microbiomeDataSets)
tse <- GrieneisenTSData()
x <- assay(tse, "counts")
d <- dist(x)
ts <- system.time(hcs <- stats::hclust(d))
tf <- system.time(hcf <- fastcluster::hclust(d))
print(ts)
print(tf)


```

TuomasBorman: Drops make an ocean, I guess...

I prefer to keep this as simple as possible and to avoid dependencies that are not vital, but speed is also thing that we must consider. 

In `getExperimentCrossCorrelation`, other things create the bottle neck, and this clustering part is not something that requires this kind of optimization. But here I think it is more beneficial to have faster algorithm...

However, this is very easy to change: just `stats::hclust()` --> `fastcluster::hclust()`

--> It's ok with me if we start using `fastcluster`

antagomir:
Hmm also true that one can give fastcluster::hclust as a function argument.

Wouldn't that be a solution: use hclust by default to avoid dependencies, and show in examples how to speed up with fastclust?

R/rowClust.R
antagomir: Hmm also true that one can give fastcluster::hclust as a function argument.

Wouldn't that be a solution: use hclust by default to avoid dependencies, and show in examples how to speed up with fastcluster?

TuomasBorman: I like that idea 

- There could be an internal function that has argument `clustering_method = stats::hclust`
     - By using that kind of approach, we could hide argument with `...` so that there are not too many options
     - Internal function should have input check that ensures that method is `stats::hclust` or `fastcluster::hclust`, because we cannot be sure that all methods return results in same format

antagomir: This is ok, let's do.

The clustering should use CLR-transformed data in order to reduce compositionality bias. This brings it closer to SparCC and SPIEC-EASI, which are taxonomic co-occurrence network reconstuction methods. We may need to compare to these later, but clr + hclust should be good enough for now, and I have seen it occasionally being used in literature.

ChouaibB: Should this kind of transformation (clr) be optional? Would the method also be meant for other type of data? (e.g functional, metabolomic)

antagomir: Yes, of course. True. But make sure to have this in the examples.

R/rowClust.R
```diff
@@ -0,0 +1,114 @@
...
+#' @return \code{x} with additional column at \code{rowData} named \code{*name*}
+#'
+#' @export
+#'
+#' @references
+#' Claesson, M., Jeffery, I., Conde, S. et al. Gut microbiota composition
+#'  correlates with diet and health in the elderly. Nature 488, 178–184 (2012)
+#' \url{https://doi.org/10.1038/nature11319}
+#'
+#' @references
```
antagomir: One @references tag is enough. Remove line 34

R/rowClust.R
```diff
@@ -0,0 +1,114 @@
+#' Clusters rowData using hierarchical clustering
```
antagomir: This does not cluster rowData. This clusters the rows of assay data. 

Suggestion: "Cluster assay rows" (no need to mention hclust here on the title row)

antagomir (commented):
Seems good!

R/rowClust.R
ChouaibB: > I like that idea
> 
>     * There could be an internal function that has argument `clustering_method = stats::hclust`
>       
>       * By using that kind of approach, we could hide argument with `...` so that there are not too many options
>       * Internal function should have input check that ensures that method is `stats::hclust` or `fastcluster::hclust`, because we cannot be sure that all methods return results in same format

Regarding the implementation of it. I have tried to compare the functions (provided and the options allowed), then the comparison/checking got lengthy and messy. However, at the end of it, since both  `stats::hclust` and `fastcluster::hclust` return an object of class `hclust`, I used that as the checking criterion for accepting the clustering function provided.

antagomir: Is it clear now? I think we just use stats::hclust as the default, but show an example with fastcluster::hclust. OK?

ChouaibB: Sure, I'll add a fastcluster::hclust example.

R/rowClust.R
```diff
@@ -0,0 +1,123 @@
...
+#'                 dissimilarity_FUN = function(x) 1 - stats::cor(t(x),method="spearman"),
+#'                 k = 6)
+#' }
+#' # Removing clr assay, since the next aggregation of values would be 
+#' # meaningless on clr transform.
+#' assay(tse, "clr") <- NULL
+#' # Aggregate clusters as a sum of each cluster values
+#' # (CAG: Co-Abundance Groups)
+#' tse_CAG <- mergeRows(tse, rowData(tse)$clust_id)
+#'
```
TuomasBorman: Looks nice! There could be one basic example like `rowClust(tse,abund_values = "clr",  k = 6)`

R/rowClust.R
```diff
@@ -0,0 +1,123 @@
...
+#' \url{https://doi.org/10.1186/s40168-019-0722-6}
+#'
+#' @name rowClust
+#' @export
+#' 
+#' @examples
+#' # Load example data
+#' library(mia)
+#' data("GlobalPatterns")
+#' tse <- GlobalPatterns
```
TuomasBorman: This `library(mia) `might cause that error in GH action

Additionally, there are some notes so check them also

FelixErnst (changes requested):
I don't think this function would be beneficial to the package.

As I mentioned in #186 `rowClust` is not a good function name since the same clustering should also be available for cols. (The SE object cetainly favours samples in cols and features in rows, but this is not required). 

In addition the dissimilarity functions available in `mia` are not usable with this function adding a challenge and confusion for the user.

To get to bottom of this task, I propose that you prepare a workflow which shows which type of analysis you want run. From that steps which can be standardized can be identified. Would this work for you?

antagomir:
The function name could be changed, and the procedure made to work for both rows and cols.

The original motivation was about clustering taxa into co-abundant groups (CAGs). The desired workflow will identify these CAGs by clustering, and then merge the taxa within each CAG. This will then provide reduced dimensionality and reduce multiple testing. Not the only nor statistically pure solution, but relatively common and also a possible benchmark for developing improved solutions. One common approach is to do CLR transformation, then cluster the taxa with hierarchical clustering with a user-defined cutoff (e.g. r > 0.8). 

This can be generalized to clustering samples as well, and this could be useful sometimes, at least in principle. Not so common.

Most dissimilarity functions in *mia* are ecological dissimilarities that are mainly defined between samples, not taxa, and many of them do not naturally fit taxonomic abundances. For instance, Bray-Curtis is typically used with compositional data in microbiome context (although not strictly limited to that). Unifrac is phylogeny-aware ecological distance. Etc. 

One could allow the use of any dissimilarity measure from *mia* for either margin but I am not sure if this is desirable because the natural dissimilarities for taxa and samples can be quite different.

FelixErnst:
But why call the function `rowClust`, when the real function name should be `getCAGs`?

the OSCA book has a nice section on clustering. I would stick to those example as much as possible, before a new function needs to be implemented: http://bioconductor.org/books/3.14/OSCA.advanced/clustering-redux.html

FelixErnst:
In addition, why write a wrapper function, when the implementation for getting the CAGs is not implemented?

I read this PR as a very fancy version of `rowData() <-` since the function in its current state doesn't do anything else then:

```r
mat <- dissimilarity_FUN(assay(x, "whatever"))
clusters <- clustering_FUN(mat)
clusters <- cutree(clusters , ...)
rowData(x) <- clusters 
```

The easiest way to get what you want, an a useable implementation for getting CAG, would be to implement the `getCAGs` function so that the following example works:

```r
rowData(x)$CAG <- getCAGs(x)
x_cag <- mergeRows(x, rowData(x)$CAG)
```

antagomir:
The current PR implementation already operates as suggested in the last example above, the difference is that the function name is `rowClust` instead of `getCAGs` but that can be easily changed. I reopen this until the discussion is resolved.

FelixErnst:
The current implementation is more like an `addCAG`.

You know I like to keep it simple, so my advice is: start with `getCAG`, get that function so smooth and usable that all kinds of wrapper functions are a no brainer at that point.

antagomir:
Indeed good to review the relevant [OSCA section](http://bioconductor.org/books/3.14/OSCA.advanced/clustering-redux.html) in more detail before finalizing the implementation

If those options are not sufficient, then the action points would be: 
- Add function `getCAG` (gets the factor indicating the cluster i.e. CAG for each assay row; same as current rowClust but just output the factor, not assign it to colData)
- Add function `addCAG` that adds the `getCAG` output to colData

ChouaibB:
Looking at the [OSCA book](https://bioconductor.org/books/3.14/OSCA.advanced/clustering-redux.html), it relies on `scran::clusterCells`; which just transpose the assay and pass it on to `bluster::clusterRows` [source](https://github.com/MarioniLab/scran/blob/master/R/clusterCells.R):
image

However, `bluster::clusterRows` is very practical in sense that it does allow several clustering algorithms given the desired parameters in one pass.

Yet, the question again is clustering taxonomic data to obtain the CAGs (clustering row data with features (taxa) being rows, not samples or cells).
Doing the CAGs with `bluster::clusterRows` is ofc possible. Yet, one down side is using custom define dissimilarity functions. Since the method would take the original matrix and compute the distance within e.g. `HclustParam` that accepts only a string parameter `metric` (but not a function) that is passed to `stats::dist`. As a workaround for it, the following code seemed to work (passing rather a distance matrix and no distance metric to be performed within `HclustParam`):
```
library(bluster)
library(mia)
data("GlobalPatterns")
tse <- GlobalPatterns
tse <- agglomerateByRank(tse, rank="Class")
tse <- transformCounts(tse, "counts", "clr", pseudocount = 1)
D <- as.dist(1-cor(t(assay(tse, "clr")), method = "spearman"))
CAGs <- clusterRows(D, HclustParam(metric = NULL, method="ward.D2",
                                   cut.params = list(k=3)))
CAGs
```

All opinions are welcome. Would such an example be enough and informative to put at vignette and/or OMA? or still implement the `getCAG` in a similar fashion?

antagomir:
According to `?bluster::clusterRows`, the argument x (which gets value D in your last example) is: 
"x: A numeric matrix-like object where rows represent observations and columns represent variables." This means that your dissimilarity matrix D (taxa x taxa) is clustered based on the correlation profiles (the rows of correlation matrix D), not based on abundance profiles across samples (rows of assay(x, "clr")). Clustering based on correlation profiles is possible and sometimes done in some contexts, but it is not the typical way to obtain CAGs. Hence I would say that this way to circumvent to issue does not lead to the anticipated outcome.

Our equivalent to `scran::clusterCells` would be something like `mia::clusterTaxa`. The shortcoming with `stats::dist` (used by `clusterCells` as well as `clusterRows`) is that it provides very limited number of distances, and they are not particularly suited for clustering taxa abundance profiles, or ecological studies in general. 

**Possible solution** is provided by `vegan::vegdist`: It provides more distances than `stats::dist`, they are readily suited for estimating ecological similarity between samples, and the manpage says also that "_Function ‘designdist’ can be used for defining your own dissimilarity index._". The implementation could be rather similar than in `scran::clusterCells` but we could replace `stats::dist` with `vegan::vegdist` and customize the implementation accordingly. 

This could provide a clean way to implement the general `clusterBy` method that we have discussed. Clustering by columns (samples) would be naturally addressed by the readily available distances in `vegan::vegdist`, and in addition the `designdist` function mentioned by the `vegan::vegdist` manpage could be used for more customized distances to cluster by rows (taxa) if the readily available options are not sufficient (to check if something like 1-cor(x,y) could be defined by the user?). 

This would be highly general and potentially pretty useful method that can be used to retrieve so-called CAGs as well as many other things.

FelixErnst:
I think taking a page of playbook of bluster is really good. MAybe the whole clustering business should be a new package integrating as much as possible with bluster?

antagomir:
> I think taking a page of playbook of bluster is really good. MAybe the whole clustering business should be a new package integrating as much as possible with bluster?

It is one option. We will need to have a closer look at this before concluding. We could have a look with Chouaib soon.

FelixErnst:
Any news on this?

antagomir:
We are looking into the available options at the moment. Hopefully we can have a concrete plan during Feb but let's see.

antagomir:
I tend to think we need a dedicated package for clustering. But implementing this will take time. Meanwhile, we could have this merged as a temporary solution?

antagomir:
@BananaCancer you could scroll through the discussions here and see if you get any further insipiration from this; and then you could close this PR. Thanks @ChouaibB for initiating, now @BananaCancer has done some additional new work on the clustering side that was initially motivated by the discussion that arose from this PR.

Basil-Courbayre-Dussau:
I believe every feature in this PR now exists either in bluster or mia except for the choice of the clustering method: this code offers the possibility to choose the hclust function (from `stats` or `fastcluster` for example) but `bluster` doesn't offer this feature. That's mainly because this is focused on hclust while `clusterRows` offers multiple algorithms but doesn't offer the choice for the function to use. 
Should this be addressed? If so, that would be something for bluster -> we should ask Aaron.
@antagomir @TuomasBorman

antagomir:
Choosing the source of hclustering method is not necessarily that critical (unless easy to add technically). 

But the availability of `fastcluster` would be great to have in bluster as it can lead to drastic performance gains in particular with larger data sets: http://danifold.net/fastcluster.html?section=1#performance

I would open an issue in bluster, suggesting to add fastcluster as a new clustering method/option.

Then this current issue can be closed if it does not include new material.

PR #188, opened 2021-12-22
getExperimentCrossCorrelation bugfix

TuomasBorman:
Hi,

I found a bug from `getExperimentCrossCorrelation`. 

Problem:
- When features were named similarly, `matrix` mode did not work. 
- For example when there were multiple "unknown" samples, `dplyr::select` selected all the "unknown" features --> **list of correlations/p-values in "unknown" cells**

Fix:
- Now when `mode == "matrix"` is selected, features get unique names in the beginning
- They are named back to original format at the end

Example that caused problems
```
library(mia)

data("esophagus")

tse <- esophagus

rownames(tse)[1:10] <- rep("Unknown", 10)

# SE allows same names
rownames(tse)

corr <- testExperimentCrossCorrelation(tse, mode = "matrix")

# List of correlations!
corr$cor["Unknown", 1]

cor_table <- testExperimentCrossCorrelation(tse, mode = "table")

# Not same problem with mode == table: "Unknown" samples * all samples
sum(cor_table$Var1 == "Unknown") == 10*58
```

-Tuomas


antagomir (approved):
Great!

FelixErnst (changes requested):
I am not sure I follow:
- What do you mean with "similarly"?
- Do you mean equally as shown in your example?
- Why is there a difference between the modes? Should't both modes be affected by this?

If the answer to the last question is no, then the fix does not need to start at the beginning, but just where the difference between the modes is (lines 372-387). If the answer is yes, you fixed the issue for one mode and chose to not fix it for the other one. I guess that is not the intention, is it?

TuomasBorman:
> I am not sure I follow:
> 
> * What do you mean with "similarly"?
> * Do you mean equally as shown in your example?
> * Why is there a difference between the modes? Should't both modes be affected by this?
> 
> If the answer to the last question is no, then the fix does not need to start at the beginning, but just where the difference between the modes is (lines 372-387). If the answer is yes, you fixed the issue for one mode and chose to not fix it for the other one. I guess that is not the intention, is it?

1. & 2.  Yes, equally. I have a one dataset with multiple "Unknown" taxa. 
3. No. This is caused by `dplyr::select` in function that converts table into matrices. This seemed to be robust way to do this to ensure that right name is assigned to right place. But I get your point; the solution needs to be where the problem is located

PR #195, opened 2021-12-30
getExperimentCrossCorrelation, user's own function supported

### LINKED ISSUE #192: getExperimentCrossCorrelation for sample pairs?
The current implementation of `getExperimentCrossCorrelation` calculates `n x m` correlation matrix between two input experiments.

There are many use cases, where we need to calculate correlation between sample pairs, when the columns of the two input experiments are paired. In that case, the calculation would return a correlation vector of length `n` instead, assuming that both experiments have `n` columns that are matched (by order, by name, or both). This is a subset of the full correlation matrix between two experiments, but remarkably faster to compute when only the paired correlations are needed.

Is it feasible to implement this as an additional argument in the `getExperimentCrossCorrelation` function? For instance, `paired=TRUE` or a similar argument. Or would it be better to implement an entirely new function? We would have an immediate use case in *miaTime*, if this can be added in one way or another.

TuomasBorman:
Hi,

I added support for user's own function in `getExperimentCrossCorrelation`.

There are still some things that I want to check, but I started to wonder if we want to create a more general function for this? 

Function that 

- Takes user's function as an input
   -    Must take matrix as an input
   -    Must return numeric values
- Applies it to all the feature-pairs

I think this should not be too complicated task because we already have working framework

`getExperimentCrossCorrelation` with own function

```
devtools::load_all()

data("GlobalPatterns")

tse <- GlobalPatterns
tse <- tse[1:10,]
getExperimentCrossCorrelation(tse, method = "canberra",
                                        mode = "table", 
                                        show_warnings = T,
                                        corr_FUN = stats::dist)

getExperimentCrossCorrelation(tse, method = "bray",,
                              show_warnings = FALSE,
                              mode = "matrix",
                              corr_FUN = vegan::vegdist)

```

-Tuomas


TuomasBorman:
Now, correlation calculation can also be done sample-wise, I added also `paired` option


```
# Paired samples
getExperimentCrossCorrelation(tse, method = "bray",
                              direction = "col",
                              paired = T,
                              show_warnings = FALSE,
                              corr_FUN = vegan::vegdist)
                              
# Unpaired samples                       
getExperimentCrossCorrelation(tse, method = "bray",
                              direction = "col",
                              paired = F,
                              show_warnings = FALSE,
                              corr_FUN = vegan::vegdist)

# features                          
getExperimentCrossCorrelation(tse, method = "bray",
                              direction = "row",
                              paired = F,
                              show_warnings = FALSE,
                              corr_FUN = vegan::vegdist)                         
                              
```

This is just a draft. The name of `direction `might be better, I think. There are probably some bugs

TuomasBorman:
And as we discussed, we might want new wrapper for paired samples

And maybe more general function for dissimilarities --> `getExperimentDissimilarities`

antagomir:
Looks good.

`corr_FUN` is now a misleading name if you apply it to distances (like vegan::vegdist). Because the sign in correlation and distance/dissimilarity is different. If this would allow both similarity and dissimilarity methods, a more suitable name could be `association_FUN`? Then also the function name should be something like `getExperimentCrossAssociation`. I think this can be feasible because the function is all about pairwise association, and I am not sure if there is a need, from the perspective of this function, to differentiate by sign. Then there would be no need for distinct correlation and dissimilarity functions.

One possible replacement for `direction` as the argument name is `MARGIN` which is used in `apply` family of functions. But the meaning in cross-correlation is a slightly different, perhaps that would be confusing.

antagomir (commented):
Seems good for now, see the general comments in the PR discussion thread.

Likely the computing part is possible to heavily optimize but I would leave that later, it can be done if needs arise. Doing the calculation per pair is probably expensive, compared to sweeping over tables and utilizing some parallel computing capabilities. That will require a bit more consideration but it is internal thing within the function and can be optimized and tested (if necessary) once the general function infrastructure has been set up and fixed.

TuomasBorman:
I have been trying to optimize the calculation, because in some point it starts to take too long

I wasn't able to improve looping over feature-pairs which is done when using functions `stats::cor.test`, and user's own functions

However, `stats::cor` can take matrices as an input, so that can be improved easily

antagomir:
Some ways to optimize:

1. make sure to not calculate (symmetric) associations twice (A-B vs. B-A) i.e. just upper (or lower) triangular matrix for corelations
2. fast & optimized implementations may exist (although would pile up dependencies)
3. can provide an option to reduce dimensionality: the correlation matrices are often very similar if you calculate them with a small random subset of the features, compared to all features.  Perhaps the user could choose to speed up by selecting a given fraction of features at random.

TuomasBorman:
1. Done
2. There are but I did not want to use them

Now we allow also user's own function. However, the input must be in specific format, e.g., `function(matrix, matrix)` is not allowed. 

There are some correlation functions that allow this kind of input, but then again `getExperimentCrossAssociation` might become too complicated because of different inputs and outputs

antagomir:
Should pass the tests now now if synced with master.

TuomasBorman:
1. Now this is allows user to use own function
2. Option for paired samples --> only associations between pairs are calculated
3. When association between two identical assays is measured, association is calculated for only unique variable pairs (feature1 vs feature2 --> values are assigned to other half of variable pairs (e.g. feature2 vs feature1)
    -  We have new dependency: `reshape2`. I tried `tidyverse` alternatives but it turned out that they were just too slow (all gained advantage was lost)
    - Now calculation is significantly quicker
4. When correlation is calculated without p-values, it is done with `stats::cor`-function which does it very efficiently

NAMESPACE
```diff
@@ -222,13 +222,16 @@ importFrom(decontam,isNotContaminant)
...
+importFrom(dplyr,filter)
 importFrom(dplyr,group_by)
 importFrom(dplyr,left_join)
 importFrom(dplyr,mutate)
 importFrom(dplyr,n)
 importFrom(dplyr,rename)
 importFrom(dplyr,select)
 importFrom(dplyr,tally)
+importFrom(reshape2,dcast)
+importFrom(reshape2,melt)
```
antagomir: If the motivation for a new dependency is speed, could we then consider `data.table::dcast` and `data.table::melt`?

They are even [faster, and also optimized for very large data sets](https://cran.r-project.org/web/packages/data.table/vignettes/datatable-reshape.html).

tests/testthat/test-5getExperimentCrossAssociation.R
```diff
@@ -259,27 +260,66 @@ test_that("getExperimentCrossCorrelation", {
...
+    cor <-  testExperimentCrossAssociation(mae, experiment1 = 1, experiment2 = 1, 
+                                           show_warnings = F)
+    # Get random variables and test that their duplicates are equal
+    for(i in 1:10 ){
+        random_var1 <- sample(cor$Var1, 1)
+        random_var2 <- sample(cor$Var1, 1)
+        expect_equal(as.numeric(cor[cor$Var1 == random_var1 & cor$Var2 == random_var2, c("cor", "pval", "p_adj")]),
+                     as.numeric(cor[cor$Var1 == random_var2 & cor$Var2 == random_var1, c("cor", "pval", "p_adj")]))
+    }
+    # Test that paired samples work correctly
```
antagomir: Some relevant dissimilarity functions (for instance, Kullback-Leibler divergence) can be asymmetric. 

It is good to check this here somehow but I would suggest to add an argument that defines if we can expect symmetric measure. In general, the users should be able to provide asymmetric measures as well, if they choose so. Unless this causes serious bottlenecks for the implementation.

tests/testthat/test-5getExperimentCrossAssociation.R
```diff
@@ -259,27 +260,66 @@ test_that("getExperimentCrossCorrelation", {
...
+    colnames(mat) <- colnames(tse2)
+    rownames(mat) <- rownames(tse2)
+    assay(tse2) <- mat
+    # Calculate with paired samples
+    cor_paired <- testExperimentCrossAssociation(tse1,
+                                                 experiment2 = tse2,  
+                                                 paired = T, 
+                                                 direction = "col", 
+                                                 show_warnings = F)
+    # Calculate all pairs
```
antagomir: It is always recommended (in R in general) to use full TRUE and FALSE instead of T and F.

antagomir:
Overall looks very useful. Could we add an example about with a function defined by the user, for instance how to do this with `vegan::vegdist` ?

tests/testthat/test-5getExperimentCrossAssociation.R
```diff
@@ -259,27 +259,87 @@ test_that("getExperimentCrossCorrelation", {
...
+    # When correlation between same assay is calculated, calculation is made faster
+    # by not calculating duplicates
+    expect_error(testExperimentCrossAssociation(mae, experiment1 = 1, experiment2 = 1, 
+                                                show_warnings = FALSE, symmetric_measure = "TRUE"))
+    expect_error(testExperimentCrossAssociation(mae, experiment1 = 1, experiment2 = 1, 
+                                                show_warnings = FALSE, symmetric_measure = 1))
+    expect_error(testExperimentCrossAssociation(mae, experiment1 = 1, experiment2 = 1, 
+                                                show_warnings = FALSE, symmetric_measure = NULL))
+    expect_error(testExperimentCrossAssociation(mae, experiment1 = 1, experiment2 = 1, 
+                                                show_warnings = FALSE, symmetric_measure = c(TRUE, TRUE)))
```
antagomir: just thinking if argument name `symmetric` would be enough (a bit shorter)?

FelixErnst (changes requested):
The data.table dependency doesn't fit. Either tidyverse of data.table, not both.

I will have a look whether I can remove it.

antagomir:
That's a good point. If we are able to choose, I would keep data.table due to its remarkable speed benefits. But tidyverse is so pervasive that it might be difficult to avoid.

FelixErnst:
@TuomasBorman please remove the data.table dependency. Thanks!

antagomir:
Btw. any estimate how much this will slow down the methods? Might be critical for larger data sets.

TuomasBorman:
If we want to speed up calculation when correlation between same assays is measured, we need to use other than tidyverse. I found out that it was very slow with this implementation; almost useless

FelixErnst:
This sometimes means one has to take step back and have a look where the slowdown is happening.

Usually this leads to a better overall implementation because one trips over weird code implemented by oneself. Talking from experience.

antagomir:
data.table has been specifically optimized speed in mind. If the implementation itself cannot be easily changed (good to check though), and the remarkable improvement mentioned by Tuomas is easily achieved by just changing the package that does melt/cast, then one option is to create another package that focuses on multi-table utilities. Indeed the speed can be a critical issue here as soon as the data sizes increase. Support should be provided also for larger studies if it is easy and feasible to arrange.

TuomasBorman:
Related to `.sort_variable_pairs_row_wise`... These are time that takes to  complete following code with different implementations

```
data("GlobalPatterns")

tse <- GlobalPatterns

tse <- tse[1:1000, ]

system.time(
testExperimentCrossAssociation(tse, symmetric = T)
)
```

data.table
   user  system elapsed 
202.131   0.703 248.867 

tidyverse 
   user  system elapsed 
238.842   1.038 292.780

apply
   user  system elapsed 
263.663   1.086 327.615 

rowmin/rowmax (CURRENT)
      user  system elapsed 
207.795   0.999 267.700 

Without optimization
   user  system elapsed 
400.592   1.511 494.785

antagomir:
Ok, this is not a bottleneck in the current version at least.

TuomasBorman:
Yep, the bottleneck is this, this loops through `variable_pairs`

```
# Calculate correlations
    correlations_and_p_values <- apply(variable_pairs, 1, 
                                       FUN = FUN_, 
                                       test_significance = test_significance, 
                                       assay1 = assay1, 
                                       assay2 = assay2,
                                       method = method,
                                       show_warnings = show_warnings, 
                                       association_FUN = association_FUN, 
                                       ...)
```

TuomasBorman:
I added there `random_sample` option. This takes a random sample size of a `random_sample` from the data

@antagomir I guess that was what you meant?

TuomasBorman:
Biggest changes in this PR
- User can give own function
- If correlation is calculated without p-values --> `cor()`
- `paired` option
- `show_warnings` was not working correctly in all cases
- ~`random_sample` option~
- `symmetric` option --> only half of the variable pairs are calculated
- ~`direction` option (is `MARGIN` better name?)~ `MARGIN` option
- `getExperimentCrossCorrelation` --> `getExperimentCrossAssociation`

antagomir:
1) `random_sample` argument

> I added there `random_sample` option. This takes a random sample size of a `random_sample` from the data
> 
> @antagomir I guess that was what you meant?

Just that it might be more transparent, and less maintenance, to just give example on how to subset features for this effect. Like 

`getExperimentCrossAssociation(tse[sample(seq_len(nrow(tse)), n), ], ...)`

-> I am not sure if this really requires its own function argument, at least an example. 


2) `direction` argument: I think that it would be better to stick to the same argument that are in common use in R, in the *apply family we have `MARGIN` for the same purpose, is there a need to change? I agree that `direction` seems nicer, it's about finding a compromise.

TuomasBorman:
I'm ok with both names, but maybe MARGIN would be better --> then we might want to consider the values of it. Currently, they can be "col" or "row". More suitable for MARGIN would be 1 and 2

I removed random_sample. It's quite easily done, so it unnecessary as you said

antagomir:
Yess - the random sample option could be in examples, at least if it can lead to considerable speedups.

TuomasBorman:
This is starting to be quite ready, howver one thing came into my mind

Currently sorting is disabled when `mode == "table"`. However, it is possible to store orders as a factor levels. We should probably consider to enable  sorting also in table mode.

antagomir:
I agree, it would be good to add that as an option (and example so it is clear).

TuomasBorman:
This one... There are lots of changes and all should work. I think this is ready

antagomir:
There seems to be one more "Changes requested" from FE. Is that solved, can it be ignored or not?

antagomir:
I have confirmed with @TuomasBorman that this long-pending PR is ready to be merged.

PR #197, opened 2022-01-01
Added philr example

antagomir:


antagomir:
The example could be simpler but I am not sure if it could be simplified much with the current methods. This highlights the possible need for a wrapper that would do all these operations for a given TreeSE object, and store the result in `altExp`.

FelixErnst:
I think this should be moved to an additional vignette or the OMA book.

The example does not use the `transformCounts` function and therefore doesn't fit into the man page of `transformCounts`.

antagomir:
I agree, but only if we conclude that philr transformation is not included in *mia*. 

The benefit is that we could make such tree-based transformations readily available for `TreeSE` users.

I would like to include it, the main drawback is the additional dependency (philr package).

FelixErnst:
> I agree, but only if we conclude that philr transformation is not included in _mia_.

The transformation is always done in the philr package, isn't it? You made a PR introducing the S3 method, so adding a wrapper to mia is not needed.

What is the difference for the user, if it is in the suggests and the information on how to use can be gotten from the vignette in mia or the philr package itself? Both can be accessed without having mia installed, whereas the man page is only accessible, if mia is already installed. If promoting the availability of interaction between TSE and philr is the main focus a vignette has much more reach, than the man page.

antagomir:
The justification for a wrapper would be to provide `philr` as one of the readily available transformation methods in `mia`, and add it to the input TreeSE object as a new `altExp`slot. 

This is analogous to the already implemented transformations, with the exception that the output is stored to `altExp`  whereas it is stored as a new standard assay for the currently available transformations.

The motivation for providing transformations is that they standardize the procedure and make it faster for the user to add and switch between different alternative transformation as needed. Most of the currently implemented transformations could be relatively easily added with the specific method `mia::transformCounts` but we have nevertheless seen the value in providing these as a handy wrapper for the user. The case is similar for the `philr` transformation, with equally much value as `mia::transformCounts` provides for the other currently available transformations. I think the motivation for having a convenient method for adding philr transformed altExp to TreeSE object is there; or otherwise we would need to question the wrappers for the other currently available transformations as well because even without `mia::transformCounts`, most of them could be added to TreeSE object with less code than we need now for `philr`

antagomir:
Just made a PR to [philr](https://github.com/jsilve24/philr/pull/21), adding pseudocount. 

Ideally, let us wait until that has been merged to *philr* and available via Bioconductor. I would then update this PR before merging as soon as the *philr* PR has gone through.

FelixErnst (changes requested):
Because this adds to the already quite heavy dependencies, I would prefer an alternative solution, even though this is in the Suggests. I know, you want to promote the use, but I think the example is not really good, if it is run. Can you wrap it in \dontrun{} tags so that you don't need the dependency?

antagomir:
The dontrun solution is ok if this is going to be only an example. In such case, the whole example can be moved to oma as the `philr` function is independt of **mia**.

But I think the main question is whether we want to support tree-based transformations in **mia**, or exclude them. 

The value for supporting tree-based transformations through a mia wrapper is more or less of equal value and importance than with other standard transformations, for which mia already now provides useful functionality through the `transform*` functions, and it has been generally agreed that these are useful to have. 

Supporting tree-based transformations in mia would be justified based on the argument that the framework is aimed to support the use of `TreeSE` (more than plain `SE`). Excluding support for tree-based transformations is making a disfavor on this and I think would be a notable design decision.

If this would add too many dependencies, we should think if there is a need for another package that only focuses on transformations (non-tree and tree based), this way managing the dependencies.

FelixErnst:
I think this has to be a seperate package, since it is a broad topic.

PR #199, opened 2022-01-03
remove na option added

microsud:


R/summaries.R
```diff
@@ -13,7 +13,10 @@
 #'   median or prevalence. Default is 'mean'.
 #'
 #' @param abund_values A \code{character} value to select an
-#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}} 
+#'   
+#' @param remove_na For \code{getTopTaxa} logical argument for calculation method
```
antagomir: I suggest to use the argument name `na.rm` which is used in many different functions and packages. It would be more standard.

antagomir (approved):

R/summaries.R
microsud: OK will make the change 😀

antagomir:
There is  build error, can you check and fix before we can merge? 

edit: I am working on it now.

antagomir:
The build error is now fixed in #202 , waiting to be merged to master. It's a few new lines in .github/workflows

antagomir:
Hmm.. the fix in #202 has not been merged to master yet.. seems to me that this was merged without the checks passing.

antagomir:
We should open this back after #202 has been accpepted..

PR #202, opened 2022-01-08
Fixes #137

### LINKED ISSUE #137: Document datasets
The current data sets could be documented a bit more extensively, at least including references and brief content descriptions.

Also consider updating the class from TreeSE to SE because these do not have a tree.

antagomir:
Data sets documented now.

I suggest to remove the "soilrep" data set, it is over a decade old and not used in any of the examples.

The "dmn_se" and "enterotype" data sets might later go in a clustering package, if such will be created.


antagomir:
I accept and merge this now so that we can get forward.

PR #203, opened 2022-01-08
Revert "remove na option added"

microsud:
Reverts microbiome/mia#199


antagomir:
The PR #199  was merged too early without syncing with master, and we needed to revert that here.

PR #205, opened 2022-01-11
Revert "Revert "remove na option added""

antagomir:
Reverts microbiome/mia#203

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.3.13
+Version: 1.3.14
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
              email = "felix.gm.ernst@outlook.com",
--- a/R/summaries.R
+++ b/R/summaries.R
@@ -13,7 +13,10 @@
 #'   median or prevalence. Default is 'mean'.
 #'
 #' @param abund_values A \code{character} value to select an
-#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}} 
+#'   
+#' @param na.rm For \code{getTopTaxa} logical argument for calculation method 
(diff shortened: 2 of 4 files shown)


R/summaries.R
```diff
@@ -13,7 +13,10 @@
 #'   median or prevalence. Default is 'mean'.
 #'
 #' @param abund_values A \code{character} value to select an
-#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}}
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}} 
+#'   
+#' @param na.rm For \code{getTopTaxa} logical argument for calculation method 
+#'              specified to argument \code{method}. Default is TRUE.
```
antagomir, line 19: These "methods" functions have also other possible arguments available, besides `na.rm`. 

We could allow the user to pass any of those available arguments through the "..." mechanism, I think we should allow this unless there are specific reason to not allow it. Then we could just remove the na.rm param description and state in "..." argument documentation (line 21 currently) that any other arguments available through \code{method} can be passed here, and cite the documentation of those functions for more details.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.3.13
+Version: 1.3.14
```
antagomir, line 3: should be checked before merge.

R/summaries.R
antagomir: @microsud

microsud: If we add `...` then every time we may need to specify `na.rm` in the function. By using `na.rm=TRUE` as default IMO is more user friendly.  Other arguments are never used, at least I have never used it for this particular function which returns specific output.

antagomir: Fair point. I am ok with this then.

antagomir (commented):
Can you update the NEWS file?

microsud:
@antagomir checks are clear, if you agree we merge this?

microsud:
Thanks!

FelixErnst:
What was this about?

PR #213, opened 2022-01-24
Get top taxa rank option

### LINKED ISSUE #212: getTopTaxa rank option?
It would handy if the `getTopTaxa` function could be used for any specified level, in the same way as `perSampleDominantTaxa` for instance.

It could work like this (but the rank option does not seem to be available yet, this returns ASVs):
`getTopTaxa(tse, top = 5, abund_values = "relabundance", rank="Phylum")`

This returns genus level info:
`perSampleDominantTaxa(tse, abund_values = "relabundance", rank="Genus")`

TuomasBorman:
Hi,

I added rank option to `getTopTaxa`. 

We use `na.rm` in two places. 

1. To remove NAs when calculating, e.g., rowsums
2. To remove NAs while agglomerating

In other summary functions, na.rm is used as a hidden (`...`) argument for `agglomerateByRank` (option 2), so I changed na.rm (option 1) to `remove_na`. 

Issue: https://github.com/microbiome/mia/issues/212

-Tuomas

### DIFF
--- a/R/summaries.R
+++ b/R/summaries.R
@@ -15,8 +15,11 @@
 #' @param abund_values A \code{character} value to select an
 #'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}} 
 #'   
-#' @param na.rm For \code{getTopTaxa} logical argument for calculation method 
+#' @param remove_na For \code{getTopTaxa} logical argument for calculation method 
 #'              specified to argument \code{method}. Default is TRUE. 
+#'   
+#' @param rank A single character defining a taxonomic rank. Must be a value of
+#' the output of \code{taxonomyRanks()}.
 #'
 #' @param ... Additional arguments:
 #'    \itemize{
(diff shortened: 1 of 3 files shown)


R/summaries.R
```diff
@@ -15,8 +15,11 @@
 #' @param abund_values A \code{character} value to select an
 #'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assayNames}} 
 #'   
-#' @param na.rm For \code{getTopTaxa} logical argument for calculation method 
+#' @param remove_na For \code{getTopTaxa} logical argument for calculation method 
 #'              specified to argument \code{method}. Default is TRUE.
```
antagomir, line 19: Hmm now I am puzzled which way this should be. In a vaaast number of R functions, they use `na.rm`. I would stick to that just for compatibility reasons.

microsud: I prefer `getTopTaxa` as it is now because this is very simple for users to do and avoids a lot of code for us to maintain `getTopTaxa` depending on `agglomerateByRank` and so on. 
```
top_families <- agglomerateByRank(tse, rank="Family", na.rm = TRUE) %>% getTopTaxa(...)
```

FelixErnst (changes requested):
I am not sure about his, because I don't see the point.

This works doesn't it?

```r
getTopTaxa(agglomerateByRank(x, rank = "Genus", na.rm=TRUE))
```

antagomir:
Yep, perhaps that would be sufficient. It could be added as an example in `getTopTaxa` manpage so that less educated users can get in speed more rapidly.

antagomir:
I would keep this open until the example has been added. Otherwise we will forget.

FelixErnst:
Guess what I did... 😄

PR #214, opened 2022-01-24
AgglomerateByRank: remove NA cols

### LINKED ISSUE #210: agglomerateByRank to drop the
Agglomeration from Species to Genus level (and other analogous things) still keeps the Species in rowData, although they become obsolete in the agglomeration. Could we drop the taxonomy ranks by default, or at least provide an option to do so?

```
data(GlobalPatterns)
rowData(agglomerateByRank(GlobalPatterns, "Genus"))
```

TuomasBorman:
Hi,

this adds new hidden option for removing those columns from rowData that have only NAs after agglomeration.

Issue: https://github.com/microbiome/mia/issues/210

-Tuomas


antagomir (approved):
Nice!

PR #216, opened 2022-01-27
splitBy implementation

### LINKED ISSUE #185: taxonomyRanks update?
**Problem**
The `taxonomyRanks` function does not consider "OTU" or "ASV" field as a taxonomic rank. This may be justified but it leads to a problem with `splitByRanks`. This function splits the abundance tables by the ranks listed by `taxonomyRanks`, and as a result the outcome from `splitByRanks` does not contain OTU/ASV table. In many cases the data analyst would like to have that as well. It is available through `assay(..., "counts")` but for automation purposes it would be quite handy to have access to OTU/ASV abundance table also through the `splitByRanks` output, so that operations can be run for all rank levels in a batch mode, and there is no need for a separate treatment of OTU/ASV table. See reproducible example below.

**Possible solutions**
1. Add OTU/ASV as an accepted taxonomic rank in `taxonomyRanks`. Potentially as an optional argument that would be actived in `splitByRanks`.
2. Modify `splitByRanks` so that it uses specified ranks, or `colnames(rowData(se))` by default instead of the outcome of `taxonomyRanks`

Solution 1 might help to avoid some confusion also in other context than just `splitByRanks` but there are pros and cons.

```
library(microbiomeDataSets)
mae <- HintikkaXOData()
tse <- mae[["microbiota"]]
```

`names(rowData(tse))`
[1] "Phylum"  "Class"   "Order"   "Family"  "Genus"   "Species" "OTU"    

`taxonomyRanks(tse)`
[1] "Phylum"  "Class"   "Order"   "Family"  "Genus"   "Species"

`names(splitByRanks(tse))`
[1] "Phylum"  "Class"   "Order"   "Family"  "Genus"   "Species"

### LINKED ISSUE #191: Split SE by colData group
There is a need to split SE objects by sample groups, typically defined by a colData field (subjectID, for instance).

Searches do not return anything readily available. We could implement the following in *mia* or make a PR in `SummarizedExperiment`. The latter may take a longer time?

```
library(mia)
data(GlobalPatterns)
se <- GlobalPatterns
field <- "SampleType"

# ideally, grouping could be either vector/factor of groups, or a name of colData field
splitSE <- function(se, grouping, use.names=TRUE) {
  # Sample indices for each group
  spl <- split(1:ncol(se), colData(se)[[grouping]])
  se_list <- lapply(seq_along(spl), function(i) se[,spl[[i]]])
  if (use.names) {names(se_list) <- names(spl)}
  return (se_list)
}
se_list <- splitSE(se, field)
```

TuomasBorman:
Hi,

I implemented `splitBy`. 

- It is possible to split data column-wise or row-wise 
- taxonomy ranks are not supported since this is splits the data differently than `splitByRanks`
- If grouping variable can be found from both `colData` and `rowData`, `direction` is required

Issue: https://github.com/microbiome/mia/issues/191

-Tuomas


R/splitBy.R
```diff
@@ -0,0 +1,181 @@
+#' Split \code{SummarizedExperiment} by grouping variables
```
antagomir: rows, cols, or both? mention explicitly

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#'
+#' @param x A
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#'   object
+#'
+#' @param grouping A single character value for selecting the grouping variable
+#'   from \code{colData} or \code{rowData}.
+#' 
+#' @param direction A single character value for selecting from where grouping
```
antagomir: This is often called `MARGIN` in the `apply` family of functions. How about using that?

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#'   variable should be searched. Must be "col" or "row".
```
antagomir: if (?) MARGIN is used as the argument name, we could also allow 1 (row) and 2 (col) as this is standard.

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#'   variable should be searched. Must be "col" or "row". 
+#'   
+#' @param ... 
+#' \itemize{
+#'   \item{\code{use_names} A boolean value to select whether to name elements of
+#'   list by their group names.}
+#' }
+#'
+#' @return
+#' List of \code{SummarizedExperiment} objects.
```
antagomir: the `splitByRanks` returns `SimpleList`. Should this do the same?

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#'
+#'
+#' @details
+#' \code{splitBy} split data based on grouping variable. Splitting can be done
+#' column-wise or row-wise. Returned value is a list of \code{SummarizedExperiment}
+#' objects; each element containing members of each group.
+#'
+#' @seealso
```
antagomir: Check the `@family` tag. Everything listed with that will be put under `@seealso`. Worth a try?

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#' data(GlobalPatterns)
+#' tse <- GlobalPatterns
+#' # Split data based on SampleType. 
+#' se_list <- splitBy(tse, grouping = "SampleType")
+#' 
+#' se_list[1:2]
+#' 
+#' # Create arbitrary groups
+#' colData(tse)$group <- sample(1:10, ncol(tse), replace = TRUE)
+#' rowData(tse)$group <- sample(1:10, nrow(tse), replace = TRUE)
```
antagomir: Splitting by Phylum could be one possible use case. Can we add one example on that?

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+#' # However, if you want to split data based on ranks, use splitByRanks
+#' se_list <- splitByRanks(tse)
+#' 
+#' # List of SE objects is returned. 
+#' # Each element is named based on their group name. If you don't want to name
+#' # elements, use use_name = FALSE
+#' se_list <- splitBy(tse, grouping = "SampleType", use_name = FALSE)
+#' 
+#' # If you want to combine groups back together, you can use cbind or rbind
+#' do.call(cbind, se_list)
```
antagomir: any equivalent to `unsplitByRanks`?

R/splitBy.R
```diff
@@ -0,0 +1,181 @@
...
+        group_indices <- split(1:ncol(x), colData(x)[[grouping]])
+        # Split data
+        se_list <- lapply(seq_along(group_indices), function(i){ 
+            temp <- x[ , group_indices[[i]] ]
+            return(temp)
+        })
+        # Split data in row-wise
+    } else if(direction == "row" ){
+        # Get sample indices for each group
+        group_indices <- split(1:nrow(x), rowData(x)[[grouping]])
```
antagomir: here and elsewhere I think we should use `seq_len` instead of `1:n`

antagomir (commented):
Nice!

R/splitBy.R
TuomasBorman: I think it is possible to link only manuals with that tag. 

Meaning that it is possible to link `splitByRanks` manual page. However, it seems that it is not possible to link functions directly --> we cannot add links to `splitByRanks` and `unSplitByRanks`. We can just add link to `splitByRanks`

R/splitBy.R
```diff
@@ -0,0 +1,264 @@
...
+#' @rdname splitBy
+#' @export
+setGeneric("splitBy",
+           signature = "x",
+           function(x, grouping = NULL, MARGIN = NULL, ...)
+               standardGeneric("splitBy"))
+
+#' @rdname splitBy
+#' @export
+setMethod("splitBy", signature = c(x = "ANY"),
```
FelixErnst: The scope of the function is to broad. To be compatible with the expectations from `splitByRank`, I think you need to mirror the S4 methods for `SE`, `SCE` and `TSE`. This is very important to allow future developments, e.g. tree clustering and such. Only with those interfaces will this be possible.

R/splitBy.R
```diff
@@ -0,0 +1,264 @@
...
+setMethod("unsplitBy", signature = c(x = "SimpleList"),
+    function(x, MARGIN = NULL, ...){
+        # Convert into a list
+        x <- as.list(x)
+        unsplitBy(x, MARGIN, ...)
+    }
+)
+################################ HELP FUNCTIONS ################################
+# Split data in column-wise or row-wise based on grouping variable. 
+.split_by <- function(x, grouping, MARGIN, use_names = TRUE, ...){
```
FelixErnst: `splitBy` does not perform agglomeration, but `splitByRanks` does. I would suggest you use the example with `mergeRows` and implement them. Otherwise `splitBy` is just a fancy wrapper for `[`.

Having set that an option `skip_agglomerate` is certainly a valid addition, but the focus should be to get as close as possible to what `splitByRank` does. That has also been the expectation when @antagomir raised the issue

FelixErnst (changes requested):
Thanks for the PR.

See the comments. Some rework is needed

R/splitBy.R
TuomasBorman: `splitByRanks `
1. agglomerates the data at every rank (or one specific rank if rank is specified)
2. Returns a list of `SE` objects; each includes data that is agglomerated at certain rank (Kingdom, Phylum, ...)

Should `splitBy` do the same for columns? 
1. agglomerate the data at every variable of `colData`
2. Return a list of SE objects, each object represent each variable (in case of GlobalPatterns data: sampleType, Primer, ..)


Currently, `splitBy` returns a list of `SE` objects. Each object represent each group of specified variable.
For example, it could include Soil, Skin,... when SampleType variable is chosen

antagomir: I think we need 

1. splitting for samples by groups in colData (e.g. each list element has sample per one sample category i.e. sampleType)
2. splitting for features by groups in rowData (e.g. each list element has features per one feature category i.e. Phylum)
3. the aggregated version for multiple fields as you describe above.

Potentially 1-2 are easy to handle with some one-liner; the 3 is more akin to `splitByRanks`.

I would support implementing 3 as proposed, and also checking if there can simple examples how 1-2 can be achieved with a short command (just 1-3 lines perhaps?).

FelixErnst:
I refactored the code. It is now less long and should work. The only change I made was to let row splitting take precedence.

The rowtree functionality needs to added, so maybe you like to continue with it. The next few I weeks I won't have time

R/splitBy.R
```diff
@@ -109,16 +109,26 @@ setGeneric("splitBy",
...
+        # Get the dimension that matches
         MARGIN <- which(dim(x) %in% length(f))
+        # If it matches with both directions
+        if(length(MARGIN) > 1 ){
+            # Get 2 if f can be found from colData, otherwise get 1
+            MARGIN <- ifelse(any( sapply(colData(x), 
+                                         function(var){all.equal(as.character(var), 
+                                                                 as.character(f))}) == TRUE ), 
+                             2, 1)
+        }
```
FelixErnst: A bit too complicated

TuomasBorman: The intention was to solve issue related to situation where `nrow `and `ncol` are equal. This could be solved easier way, however I think the example below is problematic

Currently, this goes through which is a problem, I think

```
data("GlobalPatterns")

tse <- GlobalPatterns

tse <- tse[1:10, 1:10]

splitBy(tse, f = colData(tse)$SampleType)
splitBy(tse, f = "SampleType")
```

My solution was wrong, because I did not take into account that `f `can be separate from `colData`/`rowData`. However, I think that the case above should give a warning ("Add factor into colData/rowData and use variable name or something")

Although, this is a rare case

TuomasBorman: Or if MARGIN could be specified by user (hidden option?)

FelixErnst: From my point of view a solution to that problem is available. You have outlined in the example above. So from my point of view, it falls in the category of the user should know. Otherwise it will only get more complicated on our part without adding any benefit for the user.

FelixErnst:
I think we are confusing things here. Current split by does two things, `split` or `mergeRows`. Since base `split` and S4Vectors `split` work on the rows of SE only, theis limitation was identified as not good and fixable. Some too long discussion and not enough thinking lead to weird implementations without nailing down the goal tightly.

So the take aways:
1. If `skip_aggplomerate = FALSE` the function is actually nothing more than `mergeRows` or `mergeCols`. Therefore remove the option altogether.
2. The name `splitBy` is too misleading. Rename the function to `splitOnDim` and retain the remaining functionality after completing 1.

Thanks

antagomir:
I wonder if `splitOnDim` is problematic when this can be for both rows and cols. In common terminology, dimension (Dim) refers to features. But this can also split by samples (per rowData or colData grouping). How about `splitOn`

FelixErnst:
I agree. Two functions called `splitOnRows` and `splitOnCols` are much better. Mirrors `mergeRows` and `mergeCols`...

antagomir:
Can we merge?

antagomir:
@TuomasBorman can you confirm if this is ready to merge? I am not certain as there seem to be "Changes requsted" - are those resolved?

TuomasBorman:
> @TuomasBorman can you confirm if this is ready to merge? I am not certain as there seem to be "Changes requsted" - are those resolved?

Seems that they are not yet solved. I will fix those when time allows

antagomir:
.. almost there!

TuomasBorman:
Now this does everything at least almost as intended. 

we have currently

- splitOn (generic, evaluates automatically it the split is done row or column-wise)
- splitOnRows
- splitOnCols

- unsplitOn (generic, evaluates automatically it the unsplit is done row or column-wise)
- unsplitOnRows
- unspitOnCols


Do we want these generic functions or are those *rows and *cols enough. It works with these generic functions, but it makes the implementation little bit more messy (splitOnRows is more straight-forward since the MARGIN is known)

antagomir:
I think that the *Rows and *Cols function would be enough. 

Except if there is no relevant difference and it is only about specifying the margin. In such case it would not make sense to separate the two.

Some check still seem to fail.

TuomasBorman:
*Rows and *Features are just for specifying the `MARGIN`

There are some cases where the `splitOn` approach does not work as intended

1. If `f` is a vector of strings/levels and nrow == ncol
2. If `f` is a single variable name, and it can be found from both colData and rowData. 

I think these two cases are the only ones, where it would be beneficial that user can specify `MARGIN`

As discussed above, there could be a hidden option for MARGIN, but the idea was not supported

antagomir (approved):

antagomir:
Ok right. In that case perhaps it is most clear to have separate *Row and *Col functions? 

Edit: and then the more generic ones no need to export,

TuomasBorman:
Well I think `splitOn` is enough. It works in most of the cases and is simple to use ( `splitOn(tse. "SampleType")` or `splitOn(tse, "Phylum")` or  `splitOn(tse. tse$SampleType)` )

There are some cases that gives unintended result:

- length of f matches with nrow and ncol
- f matches with column name from both rowdata and colData

Felix suggested that in these cases rowData takes precedence. 

However, I think that we should consider this since we know this possibility of error. That is why this version checks that f matches or asks user to specify `MARGIN`. That makes the checking little bit more complicated (more `IF`s ) but I don't that is a problem

I still have to add unit tests

TuomasBorman:
I think it's good

antagomir:
Just merge and close this PR when it is good.

Open an issue to OMA to add examples, then solve it.. let us make sure these things are covered in OMA. Adding examples will also reveal the potential weaknesses.

PR #220, opened 2022-02-07
Crosscorr bugfix

TuomasBorman:
Hi,

I found out that https://github.com/microbiome/mia/pull/188 does not fix all the problems related to equally named features. Currently, table is correctly converted into matrices: there are as many rows and columns as there are features etc...

However, the calculations are incorrect. 

assay contains multiple "Unknown" samples --> feature-pairs are created based on that --> there are multiple "Unknown vs variable 1" pairs for example --> function takes first "Unknown" feature even though that might not be the correct "Unknown" feature

For example, now all "Unknown" vs "Unknown" pairs have correlation 1. 

I solved this issue by calculating correlations with indices not with names. 

-Tuomas

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.3.18
+Version: 1.3.19
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut", "cre"),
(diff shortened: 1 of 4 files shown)


FelixErnst (approved):

FelixErnst:
Bump version, squash it, merge it 
image

PR #221, opened 2022-02-10
added splitByAlternativeRanks

### LINKED ISSUE #219: taxonomy ranks are too restictive
Hi,

in case one's taxonomy ranks fall outside of what is hard-coded in `mia::TAXONOMY_RANKS`, it breaks functions like `taxonomyRanks`.

For example:

```
> agglomerateByRank(my_tse, "Strain")
 Error: 'rank' must be a value from 'taxonomyRanks()' 
```

One solution would be to be able to customize TAXONOMY_RANKS. Or include **all** possible ranks in TAXONOMY_RANKS. Especially eukaryotes have a bunch of more ranks.

Thanks!
Bela

FelixErnst:
@antagomir @TuomasBorman please add some tests and check if it is sound. Thx

### DIFF
--- a/R/splitByRanks.R
+++ b/R/splitByRanks.R
@@ -146,6 +146,64 @@ setMethod("splitByRanks", signature = c(x = "TreeSummarizedExperiment"),
           }
 )
 
+
+################################################################################
+# splitByAlternativeRanks
+
+#' @rdname splitByRanks
+#' @export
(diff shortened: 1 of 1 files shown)


FelixErnst:
Not required, since oneliner produces the same result:

```r
lapply(as.list(rowData(x)[,c("Kingdom","Phylum")]), mergeRows, x = x)
```

PR #235, opened 2022-04-22
Repalce RDA with dbRDA

### LINKED ISSUE #227: Replace RDA with dbRDA
Hi,

currently `mia` has `runRDA` and `calculateRDA` functions that are wrappers for `vegan::rda()`. However, `vegan::dbrda` is more suitable because  there are also other options than euclidian dissimilarity (all that can be calculated with `vegdist`, I guess, + `dfun` option). 

We should consider replacing `vegan::rda` with `vegan::dbrda`. 

There are few differences that I found already

- No `scale` option --> can be dealt easily
- No _species scores_ returned (https://github.com/vegandevs/vegan/issues/254) --> we could add an option for that or just ignore

-Tuomas

TuomasBorman:
Hi,

I replaced RDA with dbRDA. 

There is a minor issue: `RDA` has `scale` argument, `dbRDA` does not have. I added there scaling, but it does not take `na.action `into account. This means that now scaling is done for all the values. Before it was done only for those values that were specified by `na.action`. --> slightly different results if there are missing values. 

I'm not sure if we should implement `na.action `also to scaling, since `scale` is not even implemented in` vegan:dbRDA.` 

Naming: when dbRDA is done with euclidean distances, it equals to RDA. Therefore, I think that the current name (runRDA) is still valid

Issue: https://github.com/microbiome/mia/issues/227

-Tuomas


antagomir:
Faling checks..?

TuomasBorman:
It is related to scaling 

"Some constraints were aliased because they were collinear (redundant)" --> because scaling is done inside the `rda` function the result is different

Now it works, but we should think about that scaling probably

antagomir:
Scaling is not part of the core dbRDA (or RDA) procedure and I think it should be optional.

Examples should therefore show scaling as an extra step before applying (db)RDA on the scaled data. Then the users can choose whether to used scaled or non-scaled data as input to dbRDA.

The Z transformation does scaling. If it does not have `na.action` argument already then that could be added to the `transform*` functions. Then the `transformSamples` would be used to scale the data before dbRDA.

The Z transformation is zero mean, unit variance in its standard form. If we allow shifting or scaling to other values, then a different name might be needed ("scaling"?). But I am not sure if this is really needed right now, perhaps that can be skipped.

The `runRDA` should be fine for name as you explained.

antagomir (approved):

antagomir (approved):

antagomir:
ready to merge?

TuomasBorman:
After these are passed then it is ready to be merged

antagomir (approved):

PR #238, opened 2022-04-28
check for NAs in assays

microsud:


R/summaries.R
```diff
@@ -428,3 +429,16 @@ setMethod("summary", signature = c(object = "SummarizedExperiment"),
     }
     return(names)
 }
+
+# Check NAs in assay, used when specifically when counts are expected
+.check_NAs_assay_counts <- function(x, abund_values){
+    assay.x <- .get_assay(x, abund_values)
+    if(any(is.na(assay.x))) {
+        stop("There are samples with NAs in 'assay'. ",
```
antagomir: Could we have here also the assay name (abund_values) printed, for clarity?

PR #249, opened 2022-05-08
Gha fix

antagomir:


antagomir:
This aims to fix GHA run 240. The problems are visible only after the merge.

PR #252, opened 2022-05-08
Gha fix

antagomir:


antagomir:
The check-bioc.yml has not been functional for several months. No way found to fix, removing for the time being.

https://github.com/microbiome/mia/actions/workflows/check-bioc.yml

PR #260, opened 2022-06-03
Added detection threshold option to getTopTaxa

### LINKED ISSUE #259: add detection value for getTopTaxa with method "prevalence"
Hi there, 
I am kind of new to this, so excuse me if this thread is futile. 
I am currently using your package and I think it's super neat. However, when I use getTopTaxa() with the method "prevalence", I get non-sensical results. 
I think the problem stems from here: 

In the getTopTaxa function: 
if (method == "prevalence") {
  taxs <- getPrevalence(assay(x, abund_values), sort = TRUE, 
                        include_lowest = TRUE)

by specifiying the input as "assay(x, abund_values)", the getPrevalenc "detection" default for "Any Object" sets in, and that default is set to 0. That results in all my features having a prevalence of 1 and thus it just alphabetically chooses the top 5. 
This could probably easily be fixed by either changing the default detection threshold to something that makes more sense, or even better, by incorporating the option of setting a detection threshold in the getTopTaxa() function. 

Thanks!

TuomasBorman:
Hi,

this fixes the issue https://github.com/microbiome/mia/issues/259

-Tuomas


antagomir:
Can we resolve the new conflict before merge?

PR #261, opened 2022-06-06
TreeSE import help function

### LINKED ISSUE #253: assay in other format than matrix causes an error
Hi,

many our functions expect that `assay` is in `matrix` format. However, that is not the case all the time. In `TreeSE`, `assay` can also be in `data.frame` format, for instance. When user imports `counts` data in` df` format, this causes an error which is hard to interpret. 

We should solve this somehow. We could 

1. Add checks for every function that is affected by this problem, (and convert the data in `matrix` format if possible)

1. Add `checkImport` function that checks that the data is imported correctly

-Tuomas

TuomasBorman:
Hi,

this is still a draft.

The idea:

- Helps user to create TreeSE from external files (like csv...)
- **User needs to read the files and add rownames to them** 
- Data is fed as matrix like objects (df, matrix...). Function converts data into right format, tells if rownames are not included, and subsets and orders data automatically and tells if some samples were dropped off 

Which data takes precedence?

- rowTree, rowData, and assays are linked. Currently, assays takes precedence, and rowData. and rowtree is manipulated accordingly
- Same think for columns: assays take precedence. If there are additional samples in colData, they are dropped off. On the other hand. if colData is missing a sample, it is added.

Should assay take precedence, or should only those samples be included that contain metadata in colData?

Related issue https://github.com/microbiome/mia/issues/253

-Tuomas


antagomir:
Does this have an example in OMA? Should it?

TuomasBorman:
Not yet, but I think it should have

However, I still have to check that it works properly since I'm not sure if I finished this (there was some discussion about if this gives any additional value...)

antagomir:
Ah right - - sorry I was confused, I thought this looked like that earlier discussion.. 

Shall we reverse the merge and reconsider it?

TuomasBorman:
Now this is reverted. We can discuss if this adds additional value

This function now

1. Gives more informative error messages ("sample 1" does not match)
    - TreeSE constructor itself gives error messages (they should be informative enough)
2. Convert data types (assay, rowData, and colData) into right format
    - The problem is usually caused because assay is not matrix. To functions where error occurs because of this, we could add a check
 3. Ensures that user gives rownames (otherwise slots are not linked)
     - There are examples in OMA that tells that rownames should be included
    
  --> we could just add those checks that check that assay is a matrix when rowSums/colSums is used

antagomir:
Yes perhaps that would be simplest. 

Should we also have a "validator" function that would essentially check whether the elements follow conventions that are typically expected in the mia framework? And it could tell if something is not as expected. (I hope I remember correctly that we do not have such thing yet)

TuomasBorman:
After the TreeSE is made, we can check if the data types are in correct format. But we cannot check for example, if row/colnames are correct. (If assay does not have rownames and colnames, they are replaced by rownames of rowData and colData)

I think the biggest problem is that if rownames are not provided. The second biggest thing is that the assay is not matrix ssince it causes errors.

antagomir:
Hmm.. right. 

If there is only assay format to check, then validator is a bit overkill. If there can be many things to check, the validator could be useful.

We could propose more warnings and checks to TreeSE constructor but that will require a separate issue and discussion in TreeSummarizedExperiment package. It would be the most straightforward place to include these checks.

TuomasBorman:
I actually discussed this in SummarizedExperiment. The purpose is to be very versatile. --> assay can be matrix, data.frame, 3D matrix... anything

In miaverse, we use only 2D matrices, but I'm not sure if it is smart to "disable" other data formats in the constructor...

TuomasBorman:
https://github.com/Bioconductor/SummarizedExperiment/issues/63

TuomasBorman:
In many times the problem is that user does not read the examples and manual...

But those checks (check that assay is matrix) should be added by us

antagomir:
Definitely not and I remember this discussion. 

But constructor could give warnings when there are missing row/col name issues etc. (if it does not do already).

And validator can diagnose most common problems for those who do not know otherwise.

TuomasBorman:
missing row/col name warning would be justified

antagomir:
To check if this is already available in the TreeSummarizedExperiment constructor. If not, we could open an issue to discuss it in that pkg.

TuomasBorman:
Or SummarizedExperiment?

antagomir:
Ah yes true.

PR #262, opened 2022-06-08
CLR transformation: remove relative transformation

### LINKED ISSUE #91: CLR transform doesn't include relative abundance step

TuomasBorman:
Fixes https://github.com/microbiome/mia/issues/91

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -1,4 +1,4 @@
-#' Transform counts
+#' Transform Counts
 #'
 #' These functions provide a variety of options for transforming abundance data.
 #' By using these functions, transformed table is calculated and stored in \code{assay}. 
@@ -29,12 +29,7 @@
 #' @param threshold A numeric value for setting threshold for pa transformation.
 #'   By default it is 0. (Only used for \code{method = "pa"})
 #'
-#' @param ... Additional arguments:
-#'     \itemize{
-#'        \item{\code{apply_rel_trans}}{ A single boolean value for specifying whether to
-#'        perform relative transfomration in prior to CLR transformation.
-#'        (By default: \code{relative = FALSE})
-#'    }
+#' @param ... additional arguments
 #'
 #' @details
 #' \code{transformCounts} or \code{transformSamples} and \code{transformFeatures}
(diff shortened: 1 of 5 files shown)


R/transformCounts.R
```diff
@@ -380,14 +378,15 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
-        if (method == "relabundance" && pseudocount > 0){
-            warning("Relative abundances vary in [0, 1]; adding a ",
-                    "pseudocount > 0 on relabundance will cause ",
+        if(  all(mat >= 0) && all(mat <= 1) && pseudocount > 0 ){
+            warning("The abundance table vary in [0, 1]; adding a ",
+                    "pseudocount > 1 on relabundance will cause ",
                     "non-sensicale results. Recommended to cross-check ",
                     "that the pseudocount choice is correct and intended.",
                     call. = FALSE)
         }
```
TuomasBorman: What is the purpose if this?

Pseudocount > 0, should not be added when doing relative transformation? (But negative is fine?).

Now I modified it so that when we have relative abundances, pseudocount > 0 should not be added (so that colSum 1 is not exceed)

However, I started to think what is the purpose of this, because at least in the case of CLR, we need to add pseudocount, and colSum will not be 1.

antagomir: I think this was intended to avoid situation where already relative data is transformed so that colSums!=1. 

But for CLR this is not a problem, colSums can be anything.

There could be a warning that pseudocount is changing the relative ratios but I tend to think that this is not needed as it is sort of self-evident that pseudocount is leading to changes in the data in all possible scenarios.

I would just remove this warning and enable pseudocount.

TuomasBorman: Additonally: I changed default value from pseudocount = FALSE to pseudocount = NULL

I think it is better because the value can be "numeric" or "no". (Not "numeric", "no", and "yes")

TuomasBorman:
@antagomir This good?

antagomir (approved):

antagomir (approved):
Yes!

PR #265, opened 2022-06-28
estimateDiversity fix

TuomasBorman:
1. If there is no rowTree and user wants to calculate all indices --> error because faith cannot be calculated
   - Not it is just warning, because user might want to calculate all but faith at the same time
2. If the matrix is non-numeric, this gives error. 
   - I added a check so that the error is more informative


antagomir:
Yes very good.

PR #269, opened 2022-07-01
Merge TreeSEs

### LINKED ISSUE #268: Function to merge two TSE objects?
Hi,

Is there something analogous to `merge_phyloseq`? Otherwise, this would be nice to have.

Thanks,
Bela

TuomasBorman:
Hi,

here is the draft

1. is `mergeTreeSummarizedExperiment` good name, or is `mergeTreeSE` better (I vote for latter)
2. This adds merges `rowData`, `colData`, and `assay` (`rowTree` is added if possible). Other data types are removed
3. I think `merge_phyloseq` adds` object2` to `object1`. (`nrow == nrow(object1)`. This merges so that if all samples and features differ between 2 objects the dimensions of final object are `dim(object1) + dim(object2) `

Issue: https://github.com/microbiome/mia/issues/268

-Tuomas


antagomir:
1. Also agree with the latter shorter name
2. Ok but if metadata can be made into a list, could it be a list of metadata slots from the merged objects? Or is more clear to remove.
3. a) What happens if there is overlap? Only the shared rows included in the merge? In that case, including all rows, when there is no overlap, is following a different logic and this can lead to problems. Solution could that a user can choose if they make a partial merge (shared rows object1 & object2 intersect) or full merge (all unique features  object1 & object2 union)
    b) Just wondering if we could use dplyr logic of inner/outer/left/right joins (see e.g [here](https://stackoverflow.com/questions/1299871/how-to-join-merge-data-frames-inner-outer-left-right)) so that the user could easily choose how they want to merge. Or is this an overkill? There can be different use cases and it would provide a much broader set of options if it is possible to take advantage of the existing, alternative merge options in R.

TuomasBorman:
2. metadata might be easy to include. 
3. This follows now full join (all unique features and all unique samples). 

Hmmm, it might be possible to implement other join approaches (left, right, inner...), but then we have a problem with lists (only full and inner is possible)

However, we could check if the list includes more than 2 elements --> full and inner
If only two elements --> left, right, inner, full

That might be actually quite easy to implement

antagomir:
2. If possible, I would support keeping the metadata this way. Users can always remove or modify it if they like and I do not think it hurts to keep it.
3. A clear option could be to follow the same logic than dplyr has for these operations. a) The pairwise merges can be done with the basic join operations, and ideally same operations could be activated for TreeSE objects as well, with the necessary limitations (discarding the parts like colData that cannot be sensibly merged); b) The list merges could be as suggested but if the pairwise option (a) is also available for pairwise cases, then the "list method" would not have to care about the number of objects in the list (two or more), it could be just one wrapper for lists of any size; and it could utilize the pairwise join operations in (a). This would make all nicely compatible with dplyr type of pipelines.

TuomasBorman:
All the samples are now preserved, but rows are depending on joining method

FULL --> all unique rows
INNER --> only shared rows
LEFT --> rows of the first object
RIGTH --> rows of the 2nd object

TuomasBorman:
Now good?

antagomir:
Unresolved conflicts. Otherwise, I will review asap..

TuomasBorman:
After your review, I can merge this so we keep up with the version numbering

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
+#' Merge TreeSE objects into single TreeSE object.
+#' 
+#' @param x a \code{\link{SummarizedExperiment}} object or a list of 
+#' \code{\link{SummarizedExperiment}} objects.
+#' 
+#' @param y a \code{\link{SummarizedExperiment}} object
```
antagomir: Optional?

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
...
+#' @param assay_name A single character value for selecting the
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}}
+#' to be merged. (By default: \code{assay_name = "counts"})
+#' 
+#' @param join A single character value for selecting the joining method.
+#' Must be 'full', 'inner', 'left', or 'right'. 'left' and 'right' are disabled
+#' when more than two objects are being merged.  (By default: \code{join = "full"})
+#' 
+#' @param missing_values NA, 0, or a single character values specifying the notation
+#' of missing values. (By default: \code{missing_values = 0})
```
antagomir: Would seem more logical to me to have NA as the default?

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
...
+#' 
+#' @param verbose A single boolean value to choose whether to show messages. 
+#' (By default: \code{verbose = TRUE})
+#'
+#' @param ... optional arguments (not used).
+#'
+#' @return A single \code{TreeSummarizedExperiment} object.
+#'
+#' @details
+#' This function merges multiple \code{SummarizedExperiment} objects. It combines
```
antagomir: Says to merge multiple SE (line 27) into a single TreeSE (line 24).

Is this sufficiently logical? Output and input types are different. In this case this might be justified since we explicitly want to operate with TreeSE objects. 

But I am thinking whether we should still keep type conversions and merge operations separate. Input and output would be same type (SE list -> SE; or TreeSE list -> TreeSE) and conversion would be done separately (for inputs or for output).

Otherwise, it would be good to clarify the motivation more in the @details part.

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
...
+#'   \item{\code{full} -- all unique features}
+#'   \item{\code{inner} -- all shared features}
+#'   \item{\code{left} -- all the features of the first object}
+#'   \item{\code{right} -- all the features of the second object}
+#' }
+#'
+#' @seealso
+#' \itemize{
+#'   \item{\code{TreeSummarizedExperiment::cbind}}
+#'   \item{\code{TreeSummarizedExperiment::rbind}}
```
antagomir: This seealso part could refer to the analogous dplyr package functionality so that it is clear it is being replicated here.

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
...
+#' }
+#'
+#' @name mergeTreeSE
```
antagomir: Is this name misleading if input can be a list of SEs?

R/mergeTreeSE.R
```diff
@@ -0,0 +1,438 @@
...
+#' tse <- mergeTreeSE(tse1, tse2)
+#' 
+#' # Merge a list of TreeSEs
+#' list <- SimpleList(tse1, tse2, tse3)
+#' tse <- mergeTreeSE(list, assay_name = "counts", missing_values = 0)
+#' tse
+#' 
+#' # With 'join', it is possible to specify the merging method. Subsets are used
+#' # here just to show the functionality
+#' tse <- mergeTreeSE(tse[1:10, 1:10],  tse[5:100, 11:20], join = "left")
```
antagomir: Worth considering if we could allow using the same names with dplyr, i.e. `inner_join()` etc. 

The advantage would be that the concepts would be directly analogous, it would just expand the scope of these functions to (Tree)SE objects. Are there obstacles for doing this, or would it be as simple as creating aliases?

antagomir (commented):
Really nice & probably very useful. Just a few comments to consider.

R/mergeTreeSE.R
TuomasBorman: 1 TreeSE 2 SCE 3 SE

The format is decided based on this hierarchy. --> e.g. if one object is TreeSE and others SE, the result is TreeSE

R/mergeTreeSE.R
TuomasBorman: Yes, it might. I was thinking that if we want to focus on TreeSE. However, it might be better to use more general name since the function is not limited to TreeSE

R/mergeTreeSE.R
TuomasBorman: Added aliases

R/mergeTreeSE.R
antagomir: Perhaps OK this way. But makes me wonder whether it would be better to require that the users make conversions between types explicitly before applying this? So that only similar types could be merged. This could potentially help to avoid confusion. On the other hand it is handy that anything can be merged. But I am not sure if it is so common to merge different types that making the transformations _before_ applying the merge function would be a big deal to do by those who really need it.

At the minimum, there should be a warning if the user merges different types (if not already).

R/mergeTreeSE.R
antagomir: I think it should be explicitly mentioned that this is optional (if I am correct that it is).

R/mergeTreeSE.R
antagomir: Yes.

R/mergeTreeSE.R
TuomasBorman: Yes. Actually, it is required when x is SE. When x is a list of SEs, then it is disabled

#' @param y a \code{\link{SummarizedExperiment}} object when \code{x} is a
#' \code{\link{SummarizedExperiment}} object. Disabled when \code{x} is a list

antagomir:
ok to merge

R/mergeTreeSE.R
TuomasBorman: How we should tackle this? Should we add a warning?

TuomasBorman: I added warning

TuomasBorman:
This is now ready

antagomir:
Really cool. 

Could we have a reference to this, or 1-2 examples in OMA? Also with tidyverse aliases. This way we can make sure this capacity becomes more widely known.

I am merging this now.

PR #271, opened 2022-07-07
abund_values to assay_name

### LINKED ISSUE #270: abund_values to assay_name
Change abund_values to assay_name

TuomasBorman:
https://github.com/microbiome/mia/issues/270


R/calculateDMM.R
```diff
@@ -137,8 +142,9 @@ setMethod("calculateDMN", signature = c(x = "ANY"), .calculate_DMN)
 #' @rdname calculateDMN
 #' @export
 setMethod("calculateDMN", signature = c(x = "SummarizedExperiment"),
-    function(x, abund_values = exprs_values, exprs_values = "counts", transposed = FALSE, ...){
-        mat <- assay(x, abund_values)
+    function(x, assay_name = abund_values, abund_values = exprs_values, exprs_values = "counts", 
+             transposed = FALSE, ...){
+        mat <- assay(x, assay_name)
```
antagomir, line 146: Is there a deprecation msg for exprs_values and abund_values? Would be good if not yet.

TuomasBorman: Not yet, I can add

TuomasBorman: Should we decide when we are removing exprs_values and abund_values?

TuomasBorman: Actually it is not possible to detect whether user has used assay_name, abund_values, or exprs_values when it is implemented like this

antagomir: Hmm right.. if there is a fluent way of doing this I think it would be worth it. 

However "abund_values" or "exprs_values" might be something that is used also in SCE tools, we might like to allow (not deprecate) synonyms that are used in other related frameworks.

TuomasBorman: exprs_values is used in SCE tools, abund_values is only used in mia

So should we deprecate abund_values and change the description of exprs_values (It has said "will be depracated at some point..." before this PR)

antagomir: Yes.

antagomir:
You could squash & merge when ready.

PR #276, opened 2022-07-11
SE to TreeSE

### LINKED ISSUE #273: Biom import
The mia package has now the function `makeSummarizedExperimentFromBiom`. This creates SE from biom data files.

However, we have decided to use TreeSE for all microbiome data sets, for consistency (SE is missing slots needed by some methods and is therefore suboptimal anyway).

Shall we convert this function into `makeTreeSEFromBiom`? 

Also check if the mia package has other functions that should need a similar conversion.

TuomasBorman:
https://github.com/microbiome/mia/issues/273

Should we also keep those longer formats (makeTreeSummarizedExperimentFromBiom)?


antagomir:
Hmm wouldn't hurt to have an alias on that as well?

antagomir:
Merge when ready.

PR #281, opened 2022-07-16
mergeSEs fix

TuomasBorman:
Fix problem that occurs when sample names are duplicated across objects.


TuomasBorman:
TODO: Check row/colLinks, whichTree, nodeLabs --> the linkage can be more complex than just rownames vs nodeLabels

TuomasBorman:
ready for review

TuomasBorman:
Still checking multiple tree issue

R/mergeSEs.R
```diff
@@ -719,7 +944,9 @@ setMethod("right_join", signature = c(x = "ANY"),
     df1$rownames_merge_ID <- rownames(df1)
     df2$rownames_merge_ID <- rownames(df2)
     # Merge data frames into one data frame
-    df <- merge(df1, df2, by = "rownames_merge_ID", all.x = all.x, all.y = all.y)
+    # df <- merge(df1, df2, by = "rownames_merge_ID", all.x = all.x, all.y = all.y)
+    # SHOULD WE USE MERGE OR LEFT_JOIN? IS EITHER FASTER, IF DID NOT FIND SIGNIFICANT DDIFFERENCES
+    df <- FUN(x = df1, y = df2, by = "rownames_merge_ID")
     # Add rownames and remove additional column
     rownames(df) <- df$rownames_merge_ID
```
TuomasBorman: is one of them quicker?

TuomasBorman:
Quite a lot of changes

Now this takes into account multiple row/colTrees and referenceSeqs

There might be things that could be done more efficiently, but at least this should do the job

antagomir (approved):

R/mergeSEs.R
antagomir: don't know

TuomasBorman:
Still changes... it was too slow

In previous version, the function checked directly if node label can be found from trees. 

--> this might lead to situation where label is not in the rowLinks even though it is in the rowTree

--> so the function updated every single rowTree-nodeLab pair which took too much time

(The advantage was that the returned object did not have duplicated trees)


This version just checks from the rowLinks the linkage between nodeLabs and trees. This is much faster but the disadvantage is that this leads to duplicated trees for example when subsets of same original dataset are not overlapping

antagomir:
I assume it is good enough to get forward. Things can be upgraded later by us or others.

TuomasBorman:
Yep, I will merge when check are ready

Issue #279, opened 2022-07-14
getExprimentCrossAssociation example fails occasionally


antagomir:
The example with `microbiomeDataSets::peerj32()` in getExprimentCrossAssociation fails occasionally in automated builds due to network failures. This might be better to replace with a data set that is in the mia package itself. Or we could add peerj32 or HintikkaXO data set in mia for this purpose.

TuomasBorman:
Yes, good idea. hintikkaXO is used in many examples in OMA, so that might be a better choice to add?

antagomir:
Yes I think that's good.

It is now in SE format in microbiomeDataSets, should be put in TreeSE if it is stored in mia.

PR #282, opened 2022-07-17
HintikkaXOData

TuomasBorman:
https://github.com/microbiome/mia/issues/279

All SEs are converted into TreeSEs


antagomir (approved):
Seems good! Could be merged.

R/mia.R
```diff
@@ -157,4 +157,42 @@ NULL
 #' \url{DOI:10.7717/peerj.13075/supp-2}
 "peerj13075"
 
-
+#' Multiomics dataset from a rat experiment studying effect of fat and prebiotics in diet
```
antagomir, line 165: Should we somewhere in this manpage mention that the data is provided in TreeSE format?

antagomir:
TreeSE format and having this readily available in a package seems to work better than online downloads via microbiomeDataSets.

If it could be easily done I would change examples using this data set into this mia version in mia, miaViz, OMA.

R/mia.R
TuomasBorman: Added format tag to all datasets

TuomasBorman:
Hmm, yes 

> TreeSE format and having this readily available in a package seems to work better than online downloads via microbiomeDataSets.
> 
> If it could be easily done I would change examples using this data set into this mia version in mia, miaViz, OMA.

Changed data from mia examples

--> yes we should also change the data from other places (I think we have just used hintikka data in OMA, and just in couple of places... at least I hope)

antagomir (approved):
ok to merge

PR #283, opened 2022-07-18
loadFromBiom bugfix

TuomasBorman:
The biom file has taxonomy information as a list of vectors. This led to an error

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.3.33
+Version: 1.3.34
 Authors@R:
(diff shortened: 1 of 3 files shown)


TuomasBorman:
In our example data ("extdata/rich_dense_otu_table.biom") sample metadata and feature data is in data.frame format

Here it is as a list 
https://www.ebi.ac.uk/metagenomics/api/v1/analyses/MGYA00377505/file/ERR2985255_FASTQ_SSU_OTU_TABLE_JSON.biom

I assume they can be in both formats

antagomir (approved):
Ok good. Could be merged.

PR #289, opened 2022-07-26
SilvermanAGutData from miaTime, resolve errors related to microbiomeDataSets

TuomasBorman:
We did not have a dataset with reference sequences. I added SilvermanAGutDataSet from microbiomeDataSets


tests/testthat/test-5getExperimentCrossAssociation.R
```diff
@@ -3,8 +3,25 @@ context("getExperimentCrossAssociation")
...
+            }
+        )
+        # Break if mae has the data
+        if( !is.null(mae) ){
+            break
+        }
+    }
+    # Run tests if the data fetch was successful
+    if( !is.null(mae) ){
     ############################### Test input ###############################
```
TuomasBorman, line 25: Tests used peerj32 data and some test reference values are hard-coded. I changed this little bit. --> it is tested 5 times if the data can be fetched from microbiomeDataSets. If it was successful, tests are run. otherwise they are skipped

R/mia.R
```diff
@@ -202,3 +196,37 @@ NULL
...
 "HintikkaXOData"
+
+#' SilvermanAGutData
+#' 
+#' The SilvermanAGutData dataset contains 16S rRNA gene based high-throughput 
+#' profiling of 4 in vitro artificial gut models. The sampling was done hourly 
+#' and daily to capture sub-daily dynamics of microbial community originating 
+#' from human feces. The data consists of 413 taxa from 639 samples. 
+#' The data set can be used to investigate longitudinal dynamics of microbial 
+#' community in a controlled environment.
```
antagomir: This is very nice time series data. Hence it would better with the miaTime package. Could we put it there instead?

TuomasBorman: Sure, I can make a PR and just copy-paste these

TuomasBorman:
Good to go(?)

antagomir:
Is this with time series? Just thinking if that would be better placed to miaTime?

TuomasBorman:
Actually, it is in miaTime. --> mia uses it in tests, so it is loaded from miaTime

antagomir:
perfect. good to go for me at least.

PR #292, opened 2022-08-02
getExperimentCrossCorrelation: colData variables

TuomasBorman:
We were discussing that it might be good to have an option to calculate correlation between assay and sample metadata. After considering different options, I ended up to these new arguments that are used to specify columns from colData. 

Any ideas/feedback?


antagomir (commented):
Seems feasible in principle. 

Although I agree that this could be potentially simplified by creating a distinct function for this new task (associations with colData). If it not obvious what works best, we can try this out and improve when there is more experience on the issues.

TuomasBorman:
For me this is quite intuitive (although I'm very biased...); coldata_variable1 (should it have a better name?) works just like assay_name1. Instead of fetching a matrix from assay, it fetches a variable from colData

Because of the amount of different arguments and options it might be little bit hard to grasp the idea, but I think we can improve examples (And we can improve this after feedback from users)

PR #293, opened 2022-08-05
Fix runRDA

TuomasBorman:
I found this part which is little bit confusing

Without variables from colData, RDA is PCA and dbRDA is PCoA, right? Variables are specified by using formula. The function did not give an error when formula was not specified.

vegan::rda worked without formula but vegan::dbrda is not. (We changed rda to dbrda months ago). This leads now to an internal error

Now if formula is not specified, an error is thrown


TuomasBorman:
Moreover, we could create a new variable that allows user to specify variables (variable = c("SampleType", "patient_status")

(if I read the code correctly) Currently `variable` is for feeding a data.frame, but I think that is not optimal. (Because df defines all the samples, the information should not be in external table,  it should be in colData)


EDIT: x can be SE or a matrix. If we allow matrix, then variables should also be a matrix...

EDIT2: when x is SummarizedExperiment, variables is not used --> we can add this functionality quite easily (variables are colnames of colData) and the default option for function would be just `data ~ variable1 + variable2 `

Sorry for being little bit unclear

TuomasBorman:
Something like this

antagomir:
> Without variables from colData, RDA is PCA and dbRDA is PCoA, right? Variables are specified by using formula. The function did not give an error when formula was not specified.

Correct. We could choose to require variables/formula always, this is the standard use for (db)RDA anyways. On the other hand, not having variables/formula corresponds to having a scalar intercept ~1 as the formula. This is a special case that can be useful in some comparisons. It might be better to include. Ideally this would be done via vegan, not sure why they chose to exclude (something to find out perhaps). But thanks to the above-mentioned equivalences one could add call for PCA/PCoA within the function for those cases. I would consider this if it can be easily done, and add unit tests including comparison with `vegan::rda`.

Re: variables. I am also not sure if we should just require that all is in colData. This will help to make sure that the relevant information is kept in the data object for later checks. If there is a need to allow the use of external data in the function call (which is not in colData), then it is important that the function interface and calls are as simple and clear as possible. I also find the `variable` part a bit confusing.

antagomir (commented):
See comments above.

TuomasBorman:
> , not sure why they chose to exclude (something to find out perhaps).

It seems that vegan::rda has a function [with](https://github.com/vegandevs/vegan/blob/master/R/rda.formula.R) formula and [without](https://github.com/vegandevs/vegan/blob/master/R/rda.default.R)

vegan::dbRDA has only one function which gives an error if formula does not exist https://github.com/vegandevs/vegan/blob/master/R/dbrda.R

antagomir:
If dfRDA accepts `. ~ 1` as formula then I think this should correspond to the case with "no formula".

TuomasBorman:
Yep, I actually just made a new version with `x ~1` when variables are not provided

antagomir:
perfecto

TuomasBorman:
Is this good?

R/runCCA.R
```diff
@@ -186,10 +229,10 @@ setMethod("calculateCCA", "SummarizedExperiment",
 #' @rdname runCCA
 #' @importFrom SingleCellExperiment reducedDim<-
 setMethod("runCCA", "SingleCellExperiment",
-    function(x, ..., altexp = NULL, name = "CCA")
+    function(x, ..., altExp = NULL, name = "CCA")
     {
-        if (!is.null(altexp)) {
-          y <- altExp(x, altexp)
+        if (!is.null(altExp)) {
+          y <- altExp(x, altExp)
```
TuomasBorman, line 235: Replaced altexp with altExp since we use similar argument with uppercasing in other function (Is it ok to replace like this without support for altexp?)

antagomir: I think it is ok now. Not wide user base yet.

antagomir (approved):
Seems good.

PR #295, opened 2022-08-09
estimateDiversity: multiple rowTrees

### LINKED ISSUE #286: rowTree can store multiple trees, this is not taken into account currently
`rowTree` and `colTree` slots can store multiple trees. Our methods do not take this into account. For example, `estimateFaith` takes just the first tree. 

I suggest to add `whichTree` argument. This takes character values that specify the `rowTree`. 

`tree <- rowTree(tse, "phylo")`

After that, the `tse` is subsetted based on `rowTree`. (Take only those rows that are present in tree.) The faith index is calculated based on subsetted data. 

Also other functions should be checked. Probably we can have similar functionality in those

TuomasBorman:
https://github.com/microbiome/mia/issues/286

This fixes estimateDiversity (I haven't checked other functions yet)

Should we change faith to Faith?


antagomir (approved):

antagomir:
"faith" as an index name?

Then we should also change Shannon, inverse Simpson, Chao1 etc. Many but not all of the indices have capitalized names. We could use the capitalized names but it might be more clear to always use lowercase. But I am not sure which choice will cause less confusion. We also had reports that "Shannon" was not working when they had to write "shannon"..

TuomasBorman:
I was referring your comment in the issue

> Faith is written with capital as it comes from person name..

I agree that we should keep the index name as it is; also it seems that the documentation is ok so no additional changes are required

antagomir:
In the documentation, written explanations, it is best to stick to the accepted capitalized names.

PR #297, opened 2022-08-11
getExpCrossCorr: check that assays match

TuomasBorman:


TuomasBorman:
If correlation between samples are calculated and features should match. (For example, if other experiment has taxa and another temperatures of patients) --> now it is checked that they match

antagomir (approved):

PR #298, opened 2022-08-11
Multiple row/colTrees: fix functions

### LINKED ISSUE #286: rowTree can store multiple trees, this is not taken into account currently

TuomasBorman:
https://github.com/microbiome/mia/issues/286

### DIFF
--- a/R/runDPCoA.R
+++ b/R/runDPCoA.R
@@ -97,7 +97,6 @@ setGeneric("calculateDPCoA", signature = c("x", "y"),
 {
     .require_package("ade4")
     # input check
-    # Check subset_row
     y <- as.matrix(y)
     if(length(unique(dim(y))) != 1L){
         stop("'y' must be symmetric.", call. = FALSE)
--- a/R/runDPCoA.R
+++ b/R/runDPCoA.R
@@ -42,6 +42,10 @@
 #'   (Please use \code{assay_name} instead. At some point \code{abund_values}
 #'   will be disabled.)
 #'   
+#' @param tree_name a single \code{character} value for specifying which
+#'   rowTree will be used in calculation. 
+#'   (By default: \code{tree_name = "phylo"})
+#'   
 #' @param dimred String or integer scalar specifying the existing dimensionality
 #'   reduction results to use.
 #'
@@ -161,7 +165,7 @@ setMethod("calculateDPCoA", signature = c("TreeSummarizedExperiment","missing"),
             warning("Not all rows were present in the rowTree specified by 'tree_name'.",
                     "'x' is subsetted.", call. = FALSE)
             # Subset the data
-            x <- x[ whichTree ]
+            x <- x[ whichTree, ]
         }
         dist <- cophenetic.phylo(tree)
         calculateDPCoA(mat, dist, ...)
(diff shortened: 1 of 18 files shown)


R/runDPCoA.R
```diff
@@ -141,11 +139,27 @@ setMethod("calculateDPCoA", c("ANY","ANY"), .calculate_dpcoa)
 #' @rdname runDPCoA
 setMethod("calculateDPCoA", signature = c("TreeSummarizedExperiment","missing"),
     function(x, ..., assay_name = abund_values, abund_values = exprs_values, 
-             exprs_values = "counts", dimred = NULL, n_dimred = NULL)
```
TuomasBorman, line 144: dimred and n_dimred were not used

R/runDPCoA.R
```diff
@@ -141,11 +139,27 @@ setMethod("calculateDPCoA", c("ANY","ANY"), .calculate_dpcoa)
...
+        #
+        # Get tree
+        tree <- rowTree(x, tree_name)
+        # Select only those features that are in the rowTree
+        whichTree <- rowLinks(x)[ , "whichTree"] == tree_name
+        if( any(!whichTree) ){
+            warning("Not all rows were present in the rowTree specified by 'tree_name'.",
+                    "'x' is subsetted.", call. = FALSE)
+            # Subset the data
+            x <- x[ whichTree, ]
```
TuomasBorman, line 158: It might be that the rowTree(x, "something") does not include all the rows (rowTree(x, "another") has those) --> the data is subsetted to match the tree

tests/testthat/test-5runDPCoA.R
```diff
@@ -11,4 +11,57 @@ test_that("runDPCoA", {
...
+    # )
+    # expect_error(
+    #     runDPCoA(esophagus, name = 1, assay_name = "counts", tree_name = "phylo", ncomponents = 1.3, ntop = "test",
+    #              subset_row = NULL, scale = FALSE, transposed = FALSE)
+    # )
+    
+    data("GlobalPatterns")
+    tse <- mergeSEs(esophagus, GlobalPatterns)
+    # expect_warning(runDPCoA(tse))
+    # expect_warning(runDPCoA(tse, tree_name = "phylo.1"))
```
TuomasBorman, line 66: https://github.com/microbiome/mia/pull/300

TuomasBorman:
I'm not sure where this comes from; I'm not able to reproduce it


https://github.com/microbiome/mia/runs/7951632999?check_suite_focus=true#step:20:262

antagomir:
I do not immediately recognize the source either... what should we do about it? If it works otherwise well.

TuomasBorman:
The error occurs here https://github.com/vegandevs/vegan/blob/fab100aa42dd10c1a780f654b6718926ff556d3b/R/ordConstrained.R#L357

Not 100% sure, but it looks that the `if `is not as strict as it should be

E: Though I'm not able to reproduce any examples. As the test should now work (Primer variable replaced with shannon) and we kind of know what causes this, I think we can ignore this (This should be very rare case, but it might be good to inform vegandevs about this)

TuomasBorman:
Good? This should be merged next

antagomir:
good as far as I can see - !

PR #299, opened 2022-08-12
Fix Unifrac

TuomasBorman:


R/calculateUnifrac.R
```diff
@@ -193,7 +202,7 @@ setMethod("calculateUniFrac",
 #'
 #' @export
 runUnifrac <- function(x, tree, weighted = FALSE, normalized = TRUE,
-                       BPPARAM = SerialParam()){
+                       nodeLab = NULL, BPPARAM = SerialParam(), ...){
```
TuomasBorman, line 215: Since the input is matrix and tree, I added nodeLab argument. (It might be that nodes have different names than rows have even though there are link.)

R/calculateUnifrac.R
```diff
@@ -212,14 +221,32 @@ runUnifrac <- function(x, tree, weighted = FALSE, normalized = TRUE,
...
+        stop("Incompatible tree and abundance table! Please try to provide ",
+             "nodeLab.", call. = FALSE)
+    }
+    # Merge rows, so that rows that are assigned to same tree node are agglomerated
+    # together
+    # If nodeLabs were provided, merge based on those. Otherwise merge based on
+    # rownames
+    if( is.null(nodeLab) ){
+        nodeLab <- rownames(x)
+    }
```
TuomasBorman: Check that nodeLabs specify links between tree nodes and rows/cols

R/calculateUnifrac.R
```diff
@@ -212,14 +221,32 @@ runUnifrac <- function(x, tree, weighted = FALSE, normalized = TRUE,
...
+             "nodeLab.", call. = FALSE)
+    }
+    # Merge rows, so that rows that are assigned to same tree node are agglomerated
+    # together
+    # If nodeLabs were provided, merge based on those. Otherwise merge based on
+    # rownames
+    if( is.null(nodeLab) ){
+        nodeLab <- rownames(x)
+    } 
+    x <- .merge_assay_by_rows(x, nodeLab, ...)
```
TuomasBorman, line 258: Tree might have just a subset of rows --> take a subset

antagomir (commented):
Seems good to me. I would consider adding some unit tests comparing to other unifrac implementations in R if this is possible.

PR #304, opened 2022-08-25
Kendall's tau, getExperimentCrossCorrelation

### LINKED ISSUE #301: Kendall's tau as default to cross-correlation functions
Kendall's tau could be recommended as non-parametric correlation measure over Spearman, although it is slower. 

I suggest that the cross-association functions provide Kendall's tau as the default measure. R has fast implementations for Kendall's tau, those could be used to gain some speedups.

TuomasBorman:
https://github.com/microbiome/mia/issues/301

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.5.10
+Version: 1.5.12
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut"),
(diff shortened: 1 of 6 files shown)


antagomir (approved):

antagomir:
Perhaps a small comment in the @details section:

We recommend the non-parametric Kendall's tau as the default method for association analysis. Kendall's tau has desirable statistical properties and robustness at lower sample sizes. Spearman rank correlation can provide faster solutions when running times are critical.

antagomir (approved):

Issue #306, opened 2022-09-13
Generic metaphlan importer


ChouaibB:
Hi,

The `mia` function `loadFromMetaphlan` has been working perfectly with metaphlan version 2 and 3.
After testing the new 4th version, it seems that output samples have an extra header that breaks the merging when using [the script provided by the authors](https://github.com/biobakery/MetaPhlAn/blob/master/metaphlan/utils/merge_metaphlan_tables.py). An example of the extra header which seems that it existed already at the 3.0.14 version:
https://forum.biobakery.org/t/merge-metaphlan-tables-py-broken/3507
Additionally, when running the metaphlans with different parameters additional results could be obtained, e.g. [metaphlan 2 with additional columns](https://s20709.pcdn.co/wp-content/uploads/2018/06/pasted-image-0-4.png) which is the case also for higher versions, which perhaps the user would like to merge and use at downstream analysis.

Modifying [the merging script provided by the authors](https://github.com/biobakery/MetaPhlAn/blob/master/metaphlan/utils/merge_metaphlan_tables.py) would eventually solve all these problems, and eventually different output tables will be produced breaking the standard output upon which  `loadFromMetaphlan` depends on. **The script could be modified especially since it seems no re-norming or transforming of data is performed.**

Could then adding/changing the following features at `loadFromMetaphlan` make it generic for all versions and data produced:

- [The checking of metaphlan file](https://github.com/microbiome/mia/blob/master/R/loadFromMetaphlan.R#L130), by parsing the header separately? And changing [this condition](https://github.com/microbiome/mia/blob/master/R/loadFromMetaphlan.R#L164) to be able to load merged files that do not have the "ncbi_id" column (for instance the new metaphlan 4 is based on the SGB database and when converting to GTDB the specific column is discarded, and file is not able to be loaded)
- Ability to internally merge from the individual files, e.g. by specifying folder path or a wildcard of files names?

Thank you.

antagomir:
How about having separate importers for versions 2,3,4? For those parts that are shared between all, one could use internal functions. This might help to keep things simple.

In addition there could be a wrapper that recognizes the version but I am not sure if this is needed or helpful, the user should know which version they are dealing with.

I would perhaps keep any merging steps explicit and show in examples how to do them? Then it stays a bit more transparent what is being done in the merging operations.

It would be good to have importer at least for the most common expected cases, like the individual files vs. total merge (and others if there are).

ChouaibB:
[...]

PR #308, opened 2022-09-19
draft #306

ChouaibB:
Hi,

This a draft solution related to the issue: #306 for making it compatible with metaphlan 3 and 4.

Thanks

### DIFF
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -149,13 +149,13 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
 }
 
 # Check that metaphlan file contains correct information
-.check_metaphlan <- function(data){
+.check_metaphlan <- function(x){
     # Getting column indices for rowdata and assay
-    col_idx <- unlist(lapply(data, is.character))
+    col_idx <- sapply(x, is.character)
     # Get rowdata columns
-    rowdata_columns <- data[ , col_idx, drop=FALSE]
+    rowdata_columns <- x[ , col_idx, drop=FALSE]
     # Get columns that go to assay
-    assay_columns <- data[ , !col_idx]
+    assay_columns <- x[ , !col_idx, drop=FALSE]
     # Initialize result 
     result <- TRUE
     
@@ -163,7 +163,6 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
     # rest of the columns represents abundances in samples.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
     if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        #any(grepl("id", colnames(rowdata_columns))) && 
         is.numeric(unlist(assay_columns)) ){
         result <- FALSE
     }
@@ -219,7 +218,7 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
              call. = FALSE)
     }
     # Getting column indices for rowdata and assay
-    col_idx <- unlist(lapply(table, is.character))
+    col_idx <- sapply(table, is.character)
     # Get those columns that belong to rowData
     rowdata <- table[, col_idx, drop = FALSE]
     # Get those columns that belong to assay
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -150,22 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
 
 # Check that metaphlan file contains correct information
 .check_metaphlan <- function(x){
-    # Getting column indices of character columns
-    col_idx <- sapply(x, is.character)
-    # Get rowdata column indices
-    col_rowdata_idx <- col_idx[col_idx==TRUE]
-    # Getting column indices of numeric columns
-    col_assay_idx <- sapply(x[,!col_idx], is.numeric)
+    # retrieving field types present in the data 
+    col_types <- sapply(x, class)
+    if(any(!(col_types %in% c("character", "numeric")))) {
+      warning("File contains other than character and numeric fileds.")
+      }
     # Initialize result 
     result <- TRUE
     
-    # Check if all assay data is numeric and error if other data type present
-    # Check rowdata exist
-    # Check rowdata column names that they contain right information.
+    # Check presence of character fields to be used for rowData
+    # and similarly numeric fields for abundance assay data.  
+    # Check rowData column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( all(col_assay_idx) && 
-        any(col_rowdata_idx) && 
-        any(names(col_rowdata_idx) %in% "clade_name") ){
+    if( all(c("character", "numeric") %in% col_types) &&
+        any(colnames(x) %in% "clade_name")){
         result <- FALSE
     }
     return(result)
@@ -219,12 +217,12 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
         stop("'assay_name' must be a non-empty character value.",
              call. = FALSE)
     }
-    # Getting column indices for rowdata and assay
-    col_idx <- sapply(table, is.character)
-    # Get those columns that belong to rowData
-    rowdata <- table[, which(col_idx), drop = FALSE]
+    # Getting column indices for numeric data
+    col_idx <- which(sapply(table, is.numeric))
     # Get those columns that belong to assay
-    assay <- table[, which(!col_idx), drop = FALSE]
+    assay <- table[, col_idx, drop = FALSE]
+    # Get those columns that belong to rowData
+    rowdata <- table[, -col_idx, drop = FALSE]
     # Parse taxonomic levels
     taxonomy <- .parse_taxonomy(rowdata[ , 1, drop = FALSE], sep = "\\|", column_name = "clade_name", ...)
     # Add parsed taxonomy level information to rowdata
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -150,10 +150,12 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
 
 # Check that metaphlan file contains correct information
 .check_metaphlan <- function(data){
+    # Getting column indices for rowdata and assay
+    col_idx <- unlist(lapply(data, is.character))
     # Get rowdata columns
-    rowdata_columns <- data[ , 1:2]
+    rowdata_columns <- data[ , col_idx, drop=FALSE]
     # Get columns that go to assay
-    assay_columns <- data[ , 3:ncol(data)]
+    assay_columns <- data[ , !col_idx]
     # Initialize result 
     result <- TRUE
     
@@ -161,7 +163,7 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
     # rest of the columns represents abundances in samples.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
     if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
+        #any(grepl("id", colnames(rowdata_columns))) && 
         is.numeric(unlist(assay_columns)) ){
         result <- FALSE
     }
@@ -216,10 +218,12 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
         stop("'assay_name' must be a non-empty character value.",
              call. = FALSE)
     }
+    # Getting column indices for rowdata and assay
+    col_idx <- unlist(lapply(table, is.character))
     # Get those columns that belong to rowData
-    rowdata <- table[, 1:2, drop = FALSE]
+    rowdata <- table[, col_idx, drop = FALSE]
     # Get those columns that belong to assay
-    assay <- table[, 3:ncol(table), drop = FALSE]
+    assay <- table[, !col_idx, drop = FALSE]
     # Parse taxonomic levels
     taxonomy <- .parse_taxonomy(rowdata[ , 1, drop = FALSE], sep = "\\|", column_name = "clade_name", ...)
     # Add parsed taxonomy level information to rowdata
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -153,8 +153,8 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
     # retrieving field types present in the data 
     col_types <- sapply(x, class)
     if(any(!(col_types %in% c("character", "numeric")))) {
-      warning("File contains other than character and numeric fileds.")
-      }
+        warning("The input data should not contain other than character and numeric fields.")
+    }
     # Initialize result 
     result <- TRUE
     
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -150,20 +150,26 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
 
 # Check that metaphlan file contains correct information
 .check_metaphlan <- function(x){
-    # Getting column indices for rowdata and assay
+    # Getting column indices of character columns
     col_idx <- sapply(x, is.character)
-    # Get rowdata columns
-    rowdata_columns <- x[ , col_idx, drop=FALSE]
-    # Get columns that go to assay
-    assay_columns <- x[ , !col_idx, drop=FALSE]
+    # Get rowdata column indices
+    col_rowdata_idx <- col_idx[col_idx==TRUE]
+    # Getting column indices of numeric columns
+    col_idx <- sapply(x[,!col_idx], is.numeric)
+    # Get assay column indices
+    col_assay_idx <- col_idx[col_idx==TRUE]
     # Initialize result 
     result <- TRUE
     
-    # Check rowdata column names that they contain right information, and check that 
-    # rest of the columns represents abundances in samples.
+    # Check presence of other data than numeric and character
+    # Check rowdata exist
+    # Check assay abundance data exist
+    # Check rowdata column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        is.numeric(unlist(assay_columns)) ){
+    if( sum(col_idx==FALSE)==0 && 
+        sum(col_rowdata_idx)!=0 && 
+        sum(col_assay_idx)!=0 &&
+        any(names(col_rowdata_idx) %in% "clade_name") ){
         result <- FALSE
     }
     return(result)
--- a/R/loadFromMetaphlan.R
+++ b/R/loadFromMetaphlan.R
@@ -155,20 +155,16 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
     # Get rowdata column indices
     col_rowdata_idx <- col_idx[col_idx==TRUE]
     # Getting column indices of numeric columns
-    col_idx <- sapply(x[,!col_idx], is.numeric)
-    # Get assay column indices
-    col_assay_idx <- col_idx[col_idx==TRUE]
+    col_assay_idx <- sapply(x[,!col_idx], is.numeric)
     # Initialize result 
     result <- TRUE
     
-    # Check presence of other data than numeric and character
+    # Check if all assay data is numeric and error if other data type present
     # Check rowdata exist
-    # Check assay abundance data exist
     # Check rowdata column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( sum(col_idx==FALSE)==0 && 
-        sum(col_rowdata_idx)!=0 && 
-        sum(col_assay_idx)!=0 &&
+    if( all(col_assay_idx) && 
+        any(col_rowdata_idx) && 
         any(names(col_rowdata_idx) %in% "clade_name") ){
         result <- FALSE
     }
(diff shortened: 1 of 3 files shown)


R/loadFromMetaphlan.R
```diff
@@ -150,18 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
 
 # Check that metaphlan file contains correct information
 .check_metaphlan <- function(data){
```
antagomir: better to use "x" as argument name

R/loadFromMetaphlan.R
```diff
@@ -150,18 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+    # Getting column indices for rowdata and assay
+    col_idx <- unlist(lapply(data, is.character))
```
antagomir: shouldnt this do the same faster:
`sapply(x, is.character)`

R/loadFromMetaphlan.R
```diff
@@ -150,18 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
     # Initialize result 
     result <- TRUE
     
     # Check rowdata column names that they contain right information, and check that 
     # rest of the columns represents abundances in samples.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
     if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
+        #any(grepl("id", colnames(rowdata_columns))) && 
         is.numeric(unlist(assay_columns)) ){
```
antagomir: doesn't is.numeric apply directly on assay_columns?

R/loadFromMetaphlan.R
```diff
@@ -216,10 +218,12 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
         stop("'assay_name' must be a non-empty character value.",
              call. = FALSE)
     }
+    # Getting column indices for rowdata and assay
+    col_idx <- unlist(lapply(table, is.character))
```
antagomir: same comment as above, why not sapply() instead of unlist(lapply())

R/loadFromMetaphlan.R
```diff
@@ -150,18 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
 # Check that metaphlan file contains correct information
 .check_metaphlan <- function(data){
+    # Getting column indices for rowdata and assay
+    col_idx <- unlist(lapply(data, is.character))
     # Get rowdata columns
-    rowdata_columns <- data[ , 1:2]
+    rowdata_columns <- data[ , col_idx, drop=FALSE]
     # Get columns that go to assay
-    assay_columns <- data[ , 3:ncol(data)]
+    assay_columns <- data[ , !col_idx]
```
TuomasBorman: add drop=FALSE

R/loadFromMetaphlan.R
```diff
@@ -150,18 +150,20 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
     # Initialize result 
     result <- TRUE
     
     # Check rowdata column names that they contain right information, and check that 
     # rest of the columns represents abundances in samples.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
     if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
+        #any(grepl("id", colnames(rowdata_columns))) &&
```
TuomasBorman: If this check is not needed¨, it's better to remove than comment

 "does metaphlan file has a column that includes "id"..."

TuomasBorman (changes requested):
Thanks, check Leo's comments

R/loadFromMetaphlan.R
ChouaibB: apparently not:
image

R/loadFromMetaphlan.R
ChouaibB: With the demo test done earlier and the discussion at #306  it seems for now the metaphlan4 is not producing that column name.

R/loadFromMetaphlan.R
antagomir: OK. 

By your definition in the code, the assay_columns seems to be "all non-character columns". 

Runnning unlist() on the matrix is potentially slow. 

Would it be more neat to do a similar check than with `col_idx` definition, something like `col_numeric` and use that to build the if() check instead? This way you would not need to unlist the whole matrix.

TuomasBorman (approved):
add info to NEWS, e.g. "'add function name here': added support for metahplan'add version number here'

antagomir:
Do we need unit tests here..?

antagomir:
Also - check that the documentation (roxygen / manpage) is updated. The @details section could mention something about the different versions and their limitations in the current implementation?

R/loadFromMetaphlan.R
ChouaibB: Is this correct? I have redone it this way:

```
# Check that metaphlan file contains correct information
.check_metaphlan <- function(x){
    # Getting column indices for rowdata and assay
    col_idx <- sapply(x, is.character)
    # Get rowdata column indices
    col_rowdata_idx <- col_idx[col_idx==TRUE] 
    # Get assay column indices
    col_assay_idx <- !col_idx[col_idx==FALSE]
    # Initialize result 
    result <- TRUE
    
    # Check rowdata exist
    # Check assay abundance data exist
    # Check rowdata column names that they contain right information.
    # If these requirements are met, give FALSE. Otherwise, give TRUE.
    if( sum(col_rowdata_idx)!=0 && 
        sum(col_assay_idx)!=0 &&
        any(names(col_rowdata_idx) %in% "clade_name") ){
        result <- FALSE
    }
    return(result)
}
```

ChouaibB:
> Do we need unit tests here..?
Shall I add that 3 sample demo metaphlan 4 ouput (as shown at #306 ) at exdata and run a test on it?


-> Yes whynot. But it should be artificial or open data, not confidential / proprietary data.
-> such data can go to inst/extdata in the package and read in R package codes with 
`system.file("my.csv", package='mia')`

R/loadFromMetaphlan.R
antagomir: Should we accept other than non-numeric assay data? (perhaps yes, I am just asking).

Because if not, then you could check: 1) char cols; 2) numeric cols; 3) other cols.  And generate error if there are any entries in (3). Otherwise, it would be safe to proceed with just picking the character and numeric cols separately and proceeding with them without need for further numerics checks.

antagomir (commented):
see above comments

R/loadFromMetaphlan.R
ChouaibB: Sure, no other than numeric assay data should be accepted.
I do agree there could be chances of wrong output from metaphlan or parsing from read.table. It would be much safer as you said to have the three steps checked. Would these steps be good?

```
# Check that metaphlan file contains correct information
.check_metaphlan <- function(x){
    # Getting column indices of character columns
    col_idx <- sapply(x, is.character)
    # Get rowdata column indices
    col_rowdata_idx <- col_idx[col_idx==TRUE]
    # Getting column indices of numeric columns
    col_idx <- sapply(x[,!col_idx], is.numeric)
    # Get assay column indices
    col_assay_idx <- col_idx[col_idx==TRUE]
    # Initialize result 
    result <- TRUE
    
    # Check presence of other data than numeric and character
    # Check rowdata exist
    # Check assay abundance data exist
    # Check rowdata column names that they contain right information.
    # If these requirements are met, give FALSE. Otherwise, give TRUE.
    if( sum(col_idx==FALSE)==0 && 
        sum(col_rowdata_idx)!=0 && 
        sum(col_assay_idx)!=0 &&
        any(names(col_rowdata_idx) %in% "clade_name") ){
        result <- FALSE
    }
    return(result)
}
```

ChouaibB:
> > Do we need unit tests here..?
> > Shall I add that 3 sample demo metaphlan 4 ouput (as shown at #306 ) at exdata and run a test on it?
> 
> -> Yes whynot. But it should be artificial or open data, not confidential / proprietary data. -> such data can go to inst/extdata in the package and read in R package codes with `system.file("my.csv", package='mia')`

I couldn't find unfortunately reference info about those 3 demo samples from snakemake tutorial.
However, based on the [source](https://github.com/biobakery/biobakery/wiki/metaphlan3#merge-outputs) of the current test file at exdata `merged_abundance_table.txt`, the raw sequences are provided as well at that tutorial. Shall I download those and run the metaphlan 4 pipeline on it and use the corresponding output merged table as test file?

antagomir:
Hmm. I think we should ask permission for that, unless it is data from publication that we could cite. 

If we do not have example data, then just skip those tests for now. We can test internally.

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
-.check_metaphlan <- function(data){
-    # Get rowdata columns
-    rowdata_columns <- data[ , 1:2]
-    # Get columns that go to assay
-    assay_columns <- data[ , 3:ncol(data)]
+.check_metaphlan <- function(x){
+    # Getting column indices of character columns
+    col_idx <- sapply(x, is.character)
+    # Get rowdata column indices
+    col_rowdata_idx <- col_idx[col_idx==TRUE]
```
antagomir: Does this work? To me it seems that col_idx is logical, then you identify cases that are TRUE; then you pick those cases from col_idx (same object) -> all resulting entries are TRUE by definition?

If you would pick index numbers then that should be given by `which(col_idx)` I think.

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+    # Getting column indices of numeric columns
+    col_idx <- sapply(x[,!col_idx], is.numeric)
```
antagomir: These are not "numeric columns" by definition because col_idx lists character columns, and its complement !col_idx lists non-character columns. The "non-character columns" do not equal to "numeric columns". 

Why not identify character columns with `col_idx <- sapply(x, is.character)` and numeric columns with `col_idx <- sapply(x, is.numeric)` and also check that there are no other types columns.

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+    # Get assay column indices
+    col_assay_idx <- col_idx[col_idx==TRUE]
```
antagomir: see above. Have you ran this code line by line to check the output?

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
-    # rest of the columns represents abundances in samples.
+    # Check presence of other data than numeric and character
+    # Check rowdata exist
+    # Check assay abundance data exist
+    # Check rowdata column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
-        is.numeric(unlist(assay_columns)) ){
+    if( sum(col_idx==FALSE)==0 &&
```
antagomir: `sum(col_idx==FALSE)==0` -> how about just `!any(!col_idx)` or `all(col_idx)`

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+    # Check presence of other data than numeric and character
+    # Check rowdata exist
+    # Check assay abundance data exist
+    # Check rowdata column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
-        is.numeric(unlist(assay_columns)) ){
+    if( sum(col_idx==FALSE)==0 && 
+        sum(col_rowdata_idx)!=0 &&
```
antagomir: any(col_rowdata_idx)

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+    # Check rowdata exist
+    # Check assay abundance data exist
+    # Check rowdata column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
-        is.numeric(unlist(assay_columns)) ){
+    if( sum(col_idx==FALSE)==0 && 
+        sum(col_rowdata_idx)!=0 && 
+        sum(col_assay_idx)!=0 &&
```
antagomir: any()

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,27 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+        any(names(col_rowdata_idx) %in% "clade_name") ){
         result <- FALSE
```
antagomir, line 167: should this be `result <- NULL` instead?

R/loadFromMetaphlan.R
```diff
@@ -216,10 +223,12 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
              call. = FALSE)
     }
+    # Getting column indices for rowdata and assay
+    col_idx <- sapply(table, is.character)
     # Get those columns that belong to rowData
-    rowdata <- table[, 1:2, drop = FALSE]
+    rowdata <- table[, col_idx, drop = FALSE]
     # Get those columns that belong to assay
-    assay <- table[, 3:ncol(table), drop = FALSE]
+    assay <- table[, !col_idx, drop = FALSE]
```
antagomir: i would use here numeric indices, instead of non-character indices.

antagomir (changes requested):

R/loadFromMetaphlan.R
ChouaibB: yes, I have tested with the two output files from metaphlan 3 and 4 mentioned at #306.
yes the idea was retrieving the indices of character columns, I will switch it to `which(col_idx)`.

R/loadFromMetaphlan.R
ChouaibB: yes they are not numeric columns, they are everything else but character columns.
The idea of `x[,!col_idx]` in sapply was to reduce the amount of checks to the rest of the columns that were not characters (using and saving computing results from first check). But true theoretically metaphlan outputs shouldn't have beyond 2-3 character columns, so not much is saved. I will put it to just `x`

R/loadFromMetaphlan.R
ChouaibB: yes,  I tried line by line before pushing changes. I will switch to `which`

R/loadFromMetaphlan.R
antagomir: I am sure that col_idx[col_idx==TRUE] returns logical vector, which(col_idx) return numeric vector. These are not the same. I guess which is what you need.

R/loadFromMetaphlan.R
ChouaibB: no, the function return value is supposed to be boolean, as it used inside a condition at:
```
if( .check_metaphlan(table) ){
        stop("Error while reading ", file,
             "\nPlease check that the file is in merged Metaphlan file format.",
             call. = FALSE)
    }
```

R/loadFromMetaphlan.R
antagomir: But if you can safely assume that there are only numeric and character columns then it would be safer to explicitly pick the numeric ones (and throw error if there are other categories)? Then you also do not need to check any more later if the columns are numeric.

R/loadFromMetaphlan.R
antagomir: col_idx[col_idx==TRUE] will return logical vector whose length is sum(col_idx==TRUE) and all elements are TRUE.

R/loadFromMetaphlan.R
antagomir: ok if it has to be so

R/loadFromMetaphlan.R
ChouaibB: Yes it does return logical vector, and I will switch it to `which`

R/loadFromMetaphlan.R
ChouaibB: The idea of the following checks (which I will update soon according to the other comments)
```
# Check presence of other data than numeric and character
    # Check rowdata exist
    # Check assay abundance data exist
    # Check rowdata column names that they contain right information.
    # If these requirements are met, give FALSE. Otherwise, give TRUE.
    if( sum(col_idx==FALSE)==0 && 
        sum(col_rowdata_idx)!=0 && 
        sum(col_assay_idx)!=0 &&
        any(names(col_rowdata_idx) %in% "clade_name") ){
        result <- FALSE
    }
```
1st throw error if some other type than numeric or character is present
2nd presence of at least one column at rowdata
3rd presence of at least one column at numeric data
and presence of "clade_name" among rowdata upon which the rest of the metaphlan parsing will depend on.
or sorry I didn't get what you mean?

antagomir: Ok good. Perhaps @TuomasBorman could also check if he has comments on the good practices in the implementation of such checks in general.

R/loadFromMetaphlan.R
ChouaibB: I will switch this to all(col_idx) however this will depend on the previous step of `col_idx <- sapply(**x[,!col_idx]**, is.numeric)`
without it, it won't detect if there any other types than numeric and character.

R/loadFromMetaphlan.R
ChouaibB: Sorry to re-iterate over this:
I have tested `which` with a case not having character columns but numeric columns and vice-versa. And it throws an error when called before reaching the if clause, making it non informative error to the user (not the error message supposed to be thrown by .check_metaphlan).
Initially in this function I did not have intention to subset any data for which the numeric indices would be needed. That's why I thought using boolean ones and operating on them at the if conditions later would be enough (which worked when tested line by line and by the files mentioned earlier at #306).
Sorry if I am inexperienced, but sub-setting with boolean values is not advised or unstable or?
Thanks

R/loadFromMetaphlan.R
antagomir: It is risky to convert all non-character variables into numerics here inside the function. 

What if you have factors, for instance? Like factor(c("A", "B", "C")). This would be converted into numeric c(1,2,3). Do we want to enforce such behavior?

Or should we catch and remove non-numeric fields already earlier, or provide a warning? 

I am now not sure why we want to consider other than character and numeric variables. What their purpose is in this context? Should we definitely allow them, or actively not allow them?

ChouaibB: Hi,
That's true, I didn't see that.
I agree, I hardly think metaphlan ouputs will contain other than numeric and character variables.
I reformulated the checks using `class` , I tested it for different set-ups and so far working, do you think this could be ok? (at latest commit)

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,21 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
-.check_metaphlan <- function(data){
-    # Get rowdata columns
-    rowdata_columns <- data[ , 1:2]
-    # Get columns that go to assay
-    assay_columns <- data[ , 3:ncol(data)]
+.check_metaphlan <- function(x){
+    # retrieving field types present in the data 
+    col_types <- sapply(x, class)
+    if(any(!(col_types %in% c("character", "numeric")))) {
+      warning("File contains other than character and numeric fileds.")
```
antagomir: The file does not contain those but the R object that was read from the file.

How about this:
"The input data should not contain other than character and numeric fields."

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,21 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
+      }
```
antagomir: This should be at the same indentation level with its "if"?

R/loadFromMetaphlan.R
```diff
@@ -149,20 +149,21 @@ loadFromMetaphlan <- function(file, sample_meta = NULL, phy_tree = NULL, ...){
...
-    # Check rowdata column names that they contain right information, and check that 
-    # rest of the columns represents abundances in samples.
+    # Check presence of character fields to be used for rowData
+    # and similarly numeric fields for abundance assay data.  
+    # Check rowData column names that they contain right information.
     # If these requirements are met, give FALSE. Otherwise, give TRUE.
-    if( any(colnames(rowdata_columns) %in% "clade_name") && 
-        any(grepl("id", colnames(rowdata_columns))) && 
-        is.numeric(unlist(assay_columns)) ){
+    if( all(c("character", "numeric") %in% col_types) &&
```
antagomir, line 165: do we need this check any more because it has already been checked on line 155?

ChouaibB: yes, the first check focuses on other classes than numeric and character, there could be a chance that these two needed are not present after the first check.

antagomir: We should avoid unnecessary duplicate checks to keep the code short, clean, and more easily maintainable. If this check is not needed then it should be removed.

R/loadFromMetaphlan.R
TuomasBorman: I think "the most important thing" is that there are sufficient documentation; what is the purpose of each check

R/loadFromMetaphlan.R
TuomasBorman: I guess this checks if both rowData and assay is provided; it is ok

First one just checks that there are no factor etc values

antagomir:
Let us close this for now but keep in mind if/when the metaphlan issues arise and there is a need for another check.

PR #310, opened 2022-09-26
Make rownames unique by default (aggByRank)

TuomasBorman:
As discussed in Slack, there might be rows that are named equally after agglomeration, Now thy are made unique by default


TuomasBorman:
Should we make rownames unique by default or not?

antagomir:
I would say unique by default.

PR #300, opened 2022-08-22
Modifications

TuomasBorman:
I splitted this PR since it had too many things

https://github.com/microbiome/mia/pull/298

PR #311, opened 2022-09-26
altExp to altexp

TuomasBorman:
https://github.com/microbiome/mia/pull/300

### DIFF
--- a/R/getExperimentCrossAssociation.R
+++ b/R/getExperimentCrossAssociation.R
@@ -34,15 +34,15 @@
 #'   (Please use \code{assay_name2} instead. At some point \code{abund_values2}
 #'   will be disabled.)
 #' 
-#' @param altExp1 A single numeric or character value specifying alternative experiment
+#' @param altexp1 A single numeric or character value specifying alternative experiment
 #'   from the altExp of experiment 1. If NULL, then the experiment is itself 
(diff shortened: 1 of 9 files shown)


R/getExperimentCrossAssociation.R
```diff
@@ -34,15 +34,15 @@
...
 #' 
-#' @param altExp1 A single numeric or character value specifying alternative experiment
+#' @param altexp1 A single numeric or character value specifying alternative experiment
 #'   from the altExp of experiment 1. If NULL, then the experiment is itself 
 #'   and altExp option is disabled. 
-#'   (By default: \code{altExp1 = NULL})
+#'   (By default: \code{altexp1 = NULL})
 #'   
-#' @param altExp2 A single numeric or character value specifying alternative experiment
+#' @param altexp2 A single numeric or character value specifying alternative experiment
```
TuomasBorman, line 42: Do we need 'altExp1' alias? This option was not in release version, and I think it is not that commonly used yet

TuomasBorman:
Is this good? If so, this is ready

R/getExperimentCrossAssociation.R
antagomir: ok to skip if it is not needed

antagomir (approved):

Issue #309, opened 2022-09-21
summary issue


antagomir:
**Observation**

I have tse object that only has `relabund` assay.

Running summary throws the following error:

> mia::summary(tse)
Error: There are samples with NAs in 'assay': counts . This function is limited to sequencing data only. Where raw counts do not usually have NAs. Try to supply raw counts

**Problem**

The error is in this case misleading in 3 ways:

1)  Assay "counts" does not exist although the error seems to indicate so.

> assayNames(tse)
[1] "relabund"

2) There are no NAs in the assay.

> sum(is.na(assay(tse, "relabund")))
0

3) If we specify the assay in function call, we get the following:

> mia::summary(tse, assay_name="relabund")
Error: There are samples that sum to 1 or less counts. Try to supply raw counts

-> The "relabund" assay we might like to allow cases where abundances do not sum to 1 (and are less than 1). This should at most throw a warning.

**Proposed solution**

1) Summary could summarize which assays there are in the TreeSE object, and then perhaps the specified assay will be the one to describe in more detail in summary output. Consider requiring that the `assay_name` argument is explicit, to avoid confusion. If the specified assay (here by default "counts") does not exist, tell that in the error.

2) Fixing 1 should fix this.

3) Change error to warning. It might be useful for user to know if assay abundances sum to something less than 1 but this should not throw full error. Or alternatively, show abundance sum for the first few samples as part of summary (and remove warning)

microsud:
Hi Leo- I will look at this issue in coming days.

microsud:
We currently have support for counts as these are usually more informative. For this information, relative abundance data are not helpful; hence, we had decided to explicitly support only counts data.
![image](https://user-images.githubusercontent.com/16476233/191984828-07b5feff-20d8-4683-b617-5ab269c1c901.png)

 So I will add two things. The `assay_name` will be set to NULL as default so the user has to specify a name. In any case, the function will first print the assay names that are detected before throwing an error. I still think it should throw an error if data are in relative or another transformed form. What do yo think?

antagomir:
Standard Metaphlan outputs are relative abundances. Many users obtain data readily in that form. 

It can be discussed of course whether this is good and whether it should be supported. But this affects a large number of potential users directly.

Should we consider allowing summaries also for relative abundance assay (optionally)? In addition to the count assay.

microsud:
[...]

PR #312, opened 2022-09-27
summary

microsud:
#309 

Made some changes for summary to address above issue


antagomir (approved):
Looks good to me, more changes can be done if needs arise.

Just ensure all reporetd issues have been taken into account. 

Should we consider unit testing?

R/summaries.R
```diff
@@ -368,7 +368,13 @@ setMethod("countDominantFeatures", signature = c(x = "SummarizedExperiment"),
 #'
 #' @export
 setMethod("summary", signature = c(object = "SummarizedExperiment"),
-    function(object, assay_name = abund_values, abund_values = "counts"){
+    function(object, assay_name = NULL, abund_values = NULL){
+        # Specify assay names for user
+        message("Following assays detected: ", call. = FALSE)
+        print(assayNames(object))
```
TuomasBorman: I'm wondering if this is useful information. Furthermore, because it is printing, user cannot just remove this info when disabling messages

R/summaries.R
```diff
@@ -368,7 +368,13 @@ setMethod("countDominantFeatures", signature = c(x = "SummarizedExperiment"),
...
+        # Check if assay name is specified
+        .check_abund_assay(object,
+                           assay_name = assay_name)
```
TuomasBorman: Does `.check_assay_present` from utils.R work here?

TuomasBorman (commented):
Thanks!

R/summaries.R
antagomir: `print` should not be used in functions; it should be with `message` always

antagomir: @microsud I noticed the comments are still pending. If you could add the requested change we can merge.

R/summaries.R
antagomir: @microsud

R/summaries.R
antagomir: Ok I am now fixing this in another PR to this branch..

R/summaries.R
antagomir: yes. resolving this.

antagomir:
Thanks @microsud !

PR #313, opened 2022-09-28
Remove deprecated functions

TuomasBorman:


R/calculateDistance.R
```diff
@@ -55,18 +55,6 @@ setGeneric("calculateDistance", signature = c("x"),
...
-    function(x, FUN = stats::dist, ...){
-        .Deprecated( msg = paste0("'calculateDistance' is deprecated. \n",
-                                  "Instead, use directly the function that is ",
-                                  "specified by 'FUN' argument. \n",
-                                  "See help('Deprecated')") )
-        .calculate_distance(mat = x, FUN = stats::dist, ...)
-    }
-)
-
 #' @rdname calculateDistance
```
TuomasBorman: Was the intention that we remove this completely (not just support for matrices)

In the file, it says "Will be removed by Bioc 3.15"

antagomir: Yep. It can be removed.

Issue #125, opened 2021-08-10
calculateAitchison


antagomir:
I propose to add a function `calculateAitchison`. 

This would be similar to `calculateUnifrac` but it will not require tree and it will return Aitchison distances. 

These are defined as Euclidean distances between clr-transformed values, by default calculated from the assay="counts" or assay="relabundances" (which should lead to same result, consider including a unit test for that..). 

The user should be able to choose between standard (default; with clr) and robust (optional; with rclr) transformation as part of the distance calculation.

The advantage is to standardize this calculation, it has become a relatively common choice and fluent access is needed for comparison purposes.

The default names given for the distances would be "Aitchison distance" and "Robust Aitchison distance"

As an alternative to implementing `mia::calculateAitchison` function, one can use method "aitchison" through coda.base::dist function because in most cases a user can just call a specific external distance function for runMDS and other ordinations: `d <- coda.base::dist(X, method = 'aitchison')`. The limitation with this is that the "Robust Aitchison" option is not available in standard distance functions as far as I know, this would require PR to vegan::vegdist or coda.base::dist or some other distance function, to add both Aitchison and Robust Aitchison. In principle, that would be perhaps even better solution than implementing our own versions of some rather general distances.

antagomir:
@FelixErnst - wondering if we should have a more clear default policy here - how low barrier making PRs to external packages instead or expanding `mia`. Here we clearly have two options.

FelixErnst:
`coda.base` seems to a single function package only, but it is using a Cpp implementation. The one fixed criteria for `runMDS` is that it must use the same inputs as `dist`. If that is the case, I would try to go the route of suggesting the function be added and optimized in these packages. 

Maybe implement it yourself and do a benchmark to make sure, where the right place for it is.

The general idea for keeping the dependency tree small is that, if you are only importing one function an entry in suggest and a soft dependency is OK. If it is more than one function or a class, you have to import anyway. In this case I would keep the package in suggest, because of the single function aspect and the rare chance someone is using that function on their first encounter with `mia`

antagomir:
Ok I opened an [issue to coda.base](https://github.com/mcomas/coda.base/issues/2) and to [vegan](https://github.com/vegandevs/vegan/issues/433) to discuss this.

antagomir:
Given, that adding these features in the above mentioned package may take considerable time (or perhaps not?), it might be useful to have, meanwhile, our own implementation `calculateAitchison` (with "standard" and "robust" options). 

[...]

PR #321, opened 2022-10-17
CLR & ALR from vegan

TuomasBorman:
https://github.com/microbiome/mia/issues/125


TuomasBorman:
Added ALR, vegan is utilized to calculate CLR

R/transformCounts.R
```diff
@@ -54,31 +58,15 @@
...
-#' sample-wise values.}
-#'
-#' \item{'rclr'}{ rclr or robust clr is similar to regular clr. Problem of regular
-#' clr is that logarithmic transformations lead to undefined values when zeros
-#' are present in the data. In rclr, values are divided by geometric mean
-#' of observed taxa and zero values are not taken into account. Zero values will
-#' stay as zeroes. Because of high-dimensionality of data, rclr's geometric mean of 
-#' observed taxa is a good approximation to the true geometric mean.
-#' (For details, see Martino et al. 2019.).
+#' \item{'alr'}{ Additive log ratio (clr) transformation, please refer to
```
antagomir: clr -> alr

R/transformCounts.R
```diff
@@ -27,9 +27,13 @@
...
 #'   By default it is 0. (Only used for \code{method = "pa"})
-#'
-#' @param ... additional arguments
-#'
+#'   
+#' @param ... additional arguments passed on to \code{vegan:decostand}:
+#' \itemize{
+#'   \item{\code{reference_values}:} {A single value which will be used to fill 
+#'   reference sample's column in returned assay when calculating alr. 
+#'   (default: \code{refernce_values = NA})}
```
antagomir: refernce_values -> reference_values

R/transformCounts.R
```diff
@@ -405,19 +399,22 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
     # Input check end
     
     # apply pseudocount, if it is numeric
-    if( is.numeric(pseudocount) ){
+    # in clr and alr, the pseudocount is handled in vegan::decostand
+    if( is.numeric(pseudocount) && !method %in% c("clr", "rclr", "alr") ){
```
antagomir: This could do but I see a potential risk in case vegan is now, or in later versions, using pseudocount for other methods than these three. But not sure what would be a better solution. 

If we use/need only vegan transformations here, then we could just port those and just pass the arguments to vegan. Perhaps it is worth checking, could simplify maintenance, if this would be an appropriate solution.

R/transformCounts.R
```diff
@@ -477,63 +479,64 @@ setMethod("relAbundanceCounts",signature = c(x = "SummarizedExperiment"),
...
-    }
+#' @importFrom DelayedMatrixStats colSums2
+.calc_clr <- function(mat, method = "clr", pseudocount = NULL, ...){
     # Calculate colSums
     colsums <- colSums2(mat, na.rm = TRUE)
     # Check that they are equal; affects the result of CLR. CLR expects a fixed
     # constant
     if( round(max(colsums)-min(colsums), 3) != 0  ){
-        warning("All the total abundances of samples do not sum-up to a fixed constant. ",
+        warning("All the total abundances of samples do not sum-up to a fixed constant.\n",
```
antagomir: not clear what this test does; consider adding a test function that tests a specific thing.

R/transformCounts.R
antagomir: Some of the transformations have a different name in vegan but at least most of the transformations we use are available there (as it is a generic ecology R pkg)

antagomir (commented):
Some observations.

antagomir:
INteresting fail:

── Error ('test-5transformCounts.R:336'): transformCounts ──────────────────────
Error in `rowMeans(clog)`: 'x' must be an array of at least two dimensions

R/transformCounts.R
TuomasBorman: This is just a note for user; CLR is should be applied to values that sum-up to a fixed constant (e.g., relative abundances)

This might be something that user should know / this warning can be removed

antagomir: Warning is good to keep. I was just wondering if the round(..., 3) is an ideal way to test this. Perhaps it works for now if there is no obvious solution.

R/transformCounts.R
TuomasBorman: All the transformations that we have are in vegan (log differs little bit)

Should we modify our function so that it is just a wrapper for vegan::decostand?

R/transformCounts.R
TuomasBorman: Ahhh ok, when I created the warning message, I found out that "all colsums are equal precisely" was too strict

antagomir: One potentially more clear way to do this might be:
`all.equal(x,y, tolerance=0.01)`

or 
`abs(x-y) < tolerance`

R/transformCounts.R
antagomir: I think it would make sense for now. This will simplify maintenance and reduce potential for error. It also makes quite a lot of sense to direct as much of the development of that part as possible to the same toolkit that is more generally used by ecologists.

We can have aliases for our favourite transformation names if we like.

The downside is if we want to add some non-vegan transformations later. Then it might be more work to change things agains. But this is not immediately in sight and perhaps it would not be that much work if there was something that has to be done that way.

antagomir: But: I have not evaluated this very thoroughlty, if you see reasons to not do this, let's discuss. Just that if there are no obvious major benefits, then I would just provide TreeSE interface to vegan here.

TuomasBorman: This is good plan, I think. I will keep the framework for our own functions available since there will still be 2 transformations that are not calculated with vegan:

- pa (vegan pa does not have threshold option)
- log10 (in vegan, log transformation is done little bit differently. I think we could have both "log10" and "log".)

antagomir (approved):

R/transformCounts.R
antagomir: Yes this is good. We should avoid having names that are identical with vegan but have a different meaning. In this case "log" (and log10) should be fine I think. It might be good to have log2 as well if it is easy to add.

TuomasBorman:
This is starting to be good.

**Notice: the functionality of `transformCounts` is changed.** It is now the main/general function that `transformSamples` and `transformFeatures` are using. Only difference is that latter have predefined `MARGIN` and methods that are possible to apply with them

Regarding the methods: are they ok?

- `transformCounts` have all available methods of `vegan::decostand` (excluding "pa") and our methods "pa", "log2" and "log10". 
- `transformSamples` have all methods that are usually applied sample-wise
- `transformFeatures` have all methods that are usually applied feature-wise

However, documentation of `vegan::decostand` seems to be more general than our -- it does not restrict using certain method in certain `MARGIN`. I do not have that kind of experience to say if certain method is ok to apply to both directions, so if you can comment...

Below are available methods

- "alr"
- "chi.square"
- "clr"
- "frequency"
- "hellinger"
- "log"
- "log10"
- "log2"
- "max",
- "normalize"
- "pa"
- "range"
- "rank"
- "rclr",
- "relabundance"
- "rrank"
- "standardize"
- "total",
- "z"

antagomir:
Seems very good.

Are the default behaviors for "pa" and "log" the same in _mia_ and _vegan_? If possible, this would be preferred in order to avoid unintended mistakes.

In principle, `transformCounts` could be enough, and it could default to transforming samples. We wanted to let the user to be more explicit with `transformSamples` and `transformFeatures` while also blocking the use of transformations that are usually only applied in one way. This could be problematic as well, however, and it does add some more maintenance burden. The differences are minor.

I would tend to only keep `transformCounts` in order to simplify this. That would imply deprecating the other two functions, and changing OMA examples (and manpages and unit tests and vignettes in miaverse) accordingly. This shouldn't necessarily be such a big deal, considering that the future maintenance might get simplified.

TuomasBorman:
In vegan, pa does not have thresholds (e.g., over 10 counts are present, under absent), our has

In vegan, log is log(x) + 1, our is log(x)

Just keeping transformCounts might be one option. I can also see benefits of keeping *Samples and *Features. I think it is little bit more clear as you said...

After this PR, I think there is no much maintenance issues regarding *Samples and *Features; they are quite hollow containing just MARGIN and method that are specific for each direction.

antagomir:
Ok.

Zero threshold is good for "pa"; less arbitrary than 10.

This log(x)+1 was new.. they use it in altGower distance. I agree it is confusing to use the name "log" for this kind of thing. We can consider support for altGower later unless there is some good justification add right now.

Thing with transform* is that it is mostly used for samples. If the margin argument could take "rows"/"cols"; and/or "features"/"samples" in addition to 1/2, then this could be made explicit as `transformCounts(..., margin="samples")` instead of `transformSamples(...)`. Blocking some options from the user in margin-specific manner might be slightly arbitrary. 

Both would work for me.

antagomir (approved):

TuomasBorman:
transformCounts thing: We cannot support e.g., MARGIN = "samples", without disabling the default behaviour of vegan (all methods have own default MARGIN, e.g. z-tranformation --> 1, clr --> 2... ) I think this is somewhat harmful --> MARGIN should be passed to vegan as it is

I think we could deprecate transformSamples and transformFeatures, but I make a new PR for that

antagomir:
Btw I think vegan is using samples x features matrices, whereas mia uses features x samples.. this is another complication for margin 1/2. 

Anyway even if defaults do not work so seamlessly, I guess we can still have synonyms like 1/"features" and 2/"samples" for instance (would that be more clear?). So the user can choose which way they like to denote the margin. 

If we think this is not too confusing when it is clearly stated how to do samples or features, then easier to keep transformCounts. But if syncing with vegan seems too confusing, perhaps then the transformSamples and transformFeatures should be the default functions, so that it is always explicit which one is being transformed (and no need for margin argument). Then the transformCounts could be an underlying workhorse, not necessarily even exported.

I am now hesitating which one is a better solution. If the latter does not add much maintenance burden it might be a bit more risk-free.

TuomasBorman:
I did not take that into account

MARGIN can be little bit confusing, because it means little bit dfferent thing in mia compared to vegan

However, if we state the direction clearly (like MARGIN="features"), I think there is not that big problem. IN that case, transformCounts should be good enough

I add option for MARGIN = 1/"row"/"features", let's consider the need for *features and *samples after that (they should be very easy to remove; updating examples need little bit work)

antagomir:
Sounds good. 

One more thing to consider is this: do we want to have a different meaning for margin=1 in mia and in vegan? Unless I am mistaken this is going to happen if we make margin=1 equivalent to margin="features". Because vegan has features on the cols, and mia has them on the rows. If this is the case it might be more safe to not allow numeric margins in mia at all. Simply to avoid confusion when arguments are passed to vegan (or not passed).

Btw: the transformation of clr (for samples) followed by z (for features) is rather common and useful in visualizations at least. Perhaps we could add a new joint transformation, the name could be "clrz" or "clr_z" or "scaled_clr". The latter is sometimes used I have seen, but the first two would be more explicit. Unit tests should check the equivalence to doing separate clr + z.

tests/testthat/test-5transformCounts.R
```diff
@@ -15,7 +15,7 @@ test_that("relabundance", {
     f <- rev(seq_len(ncol(relabundance(actual))))
     relabundance(actual) <- relabundance(actual)[,f]
     expect_equal(rel_mat,
-                 relabundance(actual)[,f])
+                 relabundance(actual), check.attributes = FALSE)
 })
```
TuomasBorman, line 15: We have this function called relabundance... I think it does not add any value; should we remove it?

It just fetches relative abundance table from assay. It does not calculate anything

Exactly same as assay(tse, "relabundance")

TuomasBorman:
> Btw: the transformation of clr (for samples) followed by z (for features) is rather common and useful in visualizations at least. Perhaps we could add a new joint transformation, the name could be "clrz" or "clr_z" or "scaled_clr". The latter is sometimes used I have seen, but the first two would be more explicit. Unit tests should check the equivalence to doing separate clr + z.

Hmm, maybe. But I think this not giving much additional value

tests/testthat/test-5transformCounts.R
antagomir: I agree. I always thought it is a bit confusing.

But it should be deprecated and not removed immediately so that users have time to adjust their workflows.

antagomir:
> > Btw: the transformation of clr (for samples) followed by z (for features) is rather common and useful in visualizations at least. Perhaps we could add a new joint transformation, the name could be "clrz" or "clr_z" or "scaled_clr". The latter is sometimes used I have seen, but the first two would be more explicit. Unit tests should check the equivalence to doing separate clr + z.
> 
> Hmm, maybe. But I think this not giving much additional value

The added value might be limited. One purpose is to simplify common tasks.

Alternatives:

- `transformFeatures(transformSamples(tse, "clr"), "z")`
- `transformCounts(transformCounts(tse, "clr", margin="samples"), "z", margin="features")`
- `transformCounts(tse, "clrz")`

The task is so common that the latest one might be quite handy for practical convenience.

TuomasBorman:
Good to go after checks are passed.

Let's solve issues related to

- relabundance function deprecation
- *Samples & *Features deprecation
- Need for clr + Z transformation

antagomir:
Really great. We should probably revise OMA based on this as well / opening issue there unless it can be immediately sorted out?

PR #327, opened 2022-12-05
mergeSEs fixes

### LINKED ISSUE #322: mergeSEs: merge all shared assays
Currently, it is possible to merge only one assay. It would be useful to allow merging of all shared assays (or multiple assays specified by user)

### LINKED ISSUE #326: mergeSEs: same variables but different type leads to an error
If objects have equally named variables but the data type differs (continuous age or categorical age), the variable cannot be merged together. This lead to an error

```
data("GlobalPatterns")
tse <- GlobalPatterns

tse1 <- tse

colData(tse)$SampleType <- 1

mergeSEs(tse, tse1)

```

How to fix this? Check if their type is the same. If not --> warning and rename (e.g. Age_character)

TuomasBorman:
https://github.com/microbiome/mia/issues/326
As discussed in the issue, if there are equally named variables but their class differ, this leads to an error. Now it is handled by checking martching variables and merging them together only if their class match. Otherwise, the original variable name is dropped and e.g., in the case of character and numerical class, two new variables called *original_name_class* are created. This should be good enough fix.

https://github.com/microbiome/mia/issues/322
Additionally, now it is possible to merge multiple assays. However, all the assays must be specified; it is not possible to just specify assay_name = "all". I think this is ok, since there are multiple cases where merging assay is not wise (normalization over whole dataset for instance).


antagomir:
Yep this is good. 

Re #322 perhaps function example could however show how to fetch the list of "all" assays and apply it in the merge case. The comments in the example could include a small word of advice for the novices to be cautious.

PR #337, opened 2023-02-01
Deprecate transformSamples, *Features, relabundance

TuomasBorman:
Related to PR https://github.com/microbiome/mia/pull/321

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.7.5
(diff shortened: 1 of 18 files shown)


TuomasBorman:
Related to CLR + Z:
+ Easier to perform CLR + Z transformation
- Creates maitenance issues and requires additional functionality
- Is not hard to perform, we have already examples

In my opinion, we should have only examples on this thing and new method should not be implemented

antagomir:
Ok.

Issue #194, opened 2021-12-30
agglomerateByPrevalence: Unknown category


antagomir:
The function `agglomerateByPrevalence` (and `agglomerateByRank`) work fine but there is one shortcoming. When the agglomeration is done to the higher level ranks (e.g. from ASV to Genus), then the features (rows) with missing rank information (e.g. NA genus in our example) are ignored. As a result, the agglomerated relative abundance matrix does not sum up to 100%, for instance. 

This is not incorrect but what remains missing is the information of the abundance of reads that were missing the rank information (see example below). This could be calculated as `1-colSums(assay(tse, "relabundance"))`. However, in some cases it would be handy to have a category that corresponds to unclassified reads. For instance, a row feature with the name "Unclassified". This would be handy for some methods.

Two possible solutions:
1) Add a function argument that allows the inclusion of "Unclassified" category during the aggregation for those reads that are missing the rank information at the given level; this argument could be off by default; or
2) Add an example that shows how to replace NAs in rowData with the category "Unclassified" before aggregation.

The latter solution might be preferred if the first one makes the function too complicated to understand and use.

At the same go, it would help to improve the manpage and examples of `agglomerateByPrevalence`, it is somewhat poorly documented currently. Consider a own dedicated manpage for this function.

```
library(mia)
data(GlobalPatterns)
tse <- GlobalPatterns
tse <- transformSamples(tse, "counts", "relabundance")


# Features with missing Genus mapping
inds <- which(is.na(rowData(tse)$Genus))

# Agglomerate the data to the Genus level,
# combine by prevalence but in practice do not filter anything
# with zero thresholds. This results in just aggregating the genera
# without filtering. Same as agglomerateByRank (which must be checked, too).
tse2 <- agglomerateByPrevalence(tse,
                             rank = "Genus",
                             detection = 0/100,
                             prevalence = 0/100,
                             as_relative = TRUE,
			     other_label = "Other")


# The read counts are the same between:
# 1) original (ASV level) assay data 
# 2) agglomerated genus level data
# when reads with no genus mapping are ignored;
# (same would hold for relabundances)
range(colSums(assay(tse, "counts")[-inds, ]) -
      colSums(assay(tse2, "counts")))

# However, the aggregated data does not have 100% relative abundance sum
# because the agglomeration function did not count the ASVs with NA genera.
print(colSums(assay(tse, "relabundance")))
print(colSums(assay(tse2, "relabundance")))

```

FelixErnst:
[...]

PR #341, opened 2023-02-17
assay_name switch

antagomir:
Switching the derivatives of assay_name to assay.type to support the same convention as scater and scuttle.

As discussed in [scater issue 189](https://github.com/Alanocallaghan/scater/issues/189).


antagomir:
To be merged only after the _scater_ changes in [scater PR #194](https://github.com/Alanocallaghan/scater/pull/194) have been merged.

TuomasBorman (approved):

PR #342, opened 2023-02-20
loadFromBiom: ranks from prefixes

TuomasBorman:
At least some BIOM files do not have ranks; ranks in column titles are in format "taxonomy1" and the rank is informed in the taxonomy name "k__bacteria". 

Now it is possible to get rank names from prefixes --> rowData columns get standard rank names instead of "taxonomy1"


antagomir:
Excellent.

PR #344, opened 2023-02-22
mergeSEs: simplify

TuomasBorman:
Merging trees took too long when optimal solution was the function tried to found an optimal solution and the function ended up trying all the combinations

I simplified the function:

1. rowTrees are ordered based on the number of row that they include
2. Duplicated rows are removed from the rowLinks
3. --> Each row of TreeSE is present just once
4. The function includes those trees that are still present in links
5. --> If dataset includes a tree that have all the rows, only one tree is included

The disadvantage of this solution compared to previous one is that it might be that the optimal solution is not found
--> the optimal solution might be tree1 and tree3, but the function takes tree1, tree2 and tree3

- tree1 is the largest tree, tree3 the smallest
- tree3 includes certain rows that are not present in tree2

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.7.6
+Version: 1.7.7
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut"),
              email = "felix.gm.ernst@outlook.com",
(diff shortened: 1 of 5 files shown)


antagomir (approved):

antagomir:
Should be good enough for now with this solution. Better ones can be always suggested later.

TuomasBorman:
Version bump and news update for this PR, https://github.com/microbiome/mia/pull/346, https://github.com/microbiome/mia/pull/345, https://github.com/microbiome/mia/pull/342, and https://github.com/microbiome/mia/pull/337

PR #346, opened 2023-02-23
Unifrac fix

TuomasBorman:
Unifrac calculation ended up to an error when the tree contained more tips than there are rows.

In this [paper](https://journals.asm.org/doi/10.1128/AEM.71.12.8228-8235.2005?url_ver=Z39.88-2003&rfr_id=ori:rid:crossref.org&rfr_dat=cr_pub%20%200pubmed), they removed those tips that were not present in samples --> Now the same thing is done in our function

### DIFF
--- a/R/calculateUnifrac.R
+++ b/R/calculateUnifrac.R
@@ -22,7 +22,7 @@
 #'   to the same type of features (aka. microorganisms).
 #'   
 #' @param nodeLab if \code{x} is a matrix, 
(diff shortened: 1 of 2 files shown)


antagomir (approved):

antagomir:
Perfect. 

In general, we could add explicit citation to publications from which we adopt specific solutions, and mention design/default choices in the @details field.

In this case the paper is already included in the citations and warning is given, perhaps this is enough in this case so I just merge. Thanks!

PR #349, opened 2023-03-14
Skip failed test due to absence of dependency package

0xMuluh:
🐛 Fix test run to avoid failed unit test as a result of dependency package not installed.
         - Before, unit test fails when dependency package is not installed
         - now, unit test skips if dependency package is not installed
       Fix typo in unit test file.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.7.9
+Version: 1.7.11
 Authors@R:
     c(person(given = "Felix G.M.", family = "Ernst", role = c("aut"),
              email = "felix.gm.ernst@outlook.com",
--- a/NEWS
+++ b/NEWS
@@ -73,3 +73,4 @@ Changes in version 1.7.x
 + loadFromBiom: name columns of rowData based on prefixes
 + Deprecate transformSamples, *Features, relabundance, ZTransform, relAbundanceCounts
 + mergeSEs: faster tree merging
++ Faith's index: fix bug that occurred when only one taxon is present
--- a/R/estimateDivergence.R
+++ b/R/estimateDivergence.R
@@ -13,7 +13,7 @@
 #'   will be disabled.)
 #'
 #' @param name a name for the column of the colData the results should be
-#'   stored in. By defaut, \code{name} is \code{"divergence"}.
+#'   stored in. By default, \code{name} is \code{"divergence"}.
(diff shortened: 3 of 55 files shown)


antagomir:
Thanks for the suggestion.

However, I am not sure if skipping tests in such situation is a desired solution, if the alternative is to install the dependencies. In the end we would like to catch all problems that can be solved.

It is in general good practice to open an [issue](https://github.com/microbiome/mia/issues) first to receive feedback on the proposed solution, before preparing PR.

TuomasBorman:
That was my suggestion. We already have this kind of thing implemented when miaTime is not installed. These packages are not required by mia (they are only required by certain functions), so they are not installed by default.

You have a point; tests should test all functions; dependency issues of tests should not occur for ordinary user

--> So I guess, we should also remove skipping of tests that require miaTime?

antagomir:
Right, interesting point. I see pros & cons. The miaTime package is not in Bioconductor so it is more experimental. It might be justified to skip such dependencies in unit tests as a special case. Perhaps that is ok if it does not concern the core functionality of the package?

0xMuluh:
Thank you for the feedback. I'm not quite sure but I was thinking the same thing, because the miaTime, phyloseq, and biomformat do not concern the core functionallity of the package. what do you suggest I do?

antagomir:
I think this good. Apologies for the delay, I had somehow not noticed this PR. Let us wait for the automated checks to pass.

antagomir:
I accidentally closed and reopened again..

0xMuluh:
oh yes. I see, a failed check.

antagomir:
@Daenarys8 do you need support in solving this?

0xMuluh:
I'm not sure but I  will take a look right away.  Apologies for the late reply

antagomir:
Still fail. Ask when support needed.

0xMuluh:
> Still fail. Ask when support needed.

Yes please.  I need help.  I checked the changes I made and have not seen what could cause the failed check.  It fails test actually with 24 failed test.  The irony is locally,  the unittest run fine.  I'm thinking it has something to do with some omitted packages( skip_if_not_installed(********) and skip_if_not(require(*********)) but I'm not quite sure.

TuomasBorman (approved):

antagomir (approved):

antagomir:
Awesome, PR merged!

PR #355, opened 2023-03-31
Update relabundance.R

neyjay12:
The code has been commented on. The function's intent was explained, and a deprecation warning and suggested replacement were also provided. An example of its use was given, and its input and output were described. To further specify which functions are being imported and exported, the @importFrom and @export tags were included.

### DIFF
--- a/R/relabundance.R
+++ b/R/relabundance.R
@@ -30,13 +30,24 @@
 #' # head(relabundance(GlobalPatterns))
 NULL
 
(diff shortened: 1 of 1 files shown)


antagomir (approved):
Very nice!

TuomasBorman (changes requested):
Have you ran devtools::document()? This seems to add additional information to the manual (I think this is very good for developers, but for user this is extra information that is not necessarily needed)

PR #359, opened 2023-04-13
SImplify mergeSE

TuomasBorman:
mergeSEs is simplified.

1. Find classes of columns --> if equally named variables do not have same class, add class to variables' name (e.g. age --> age_factor and age_numeric)
2. merge


antagomir:
I expect this can merged so we get fwd with the other PRs.

PR #362, opened 2023-05-16
Fix loadFromBiom: prefix match

TuomasBorman:


TuomasBorman:
Previously, "k__" matched also with "sk__" prefix that can be found from biom files. This is now fixed

I will update version and news, if this is ok

antagomir:
Does "sk__" have a meaning, or is it a typo?

TuomasBorman:
I guess it is in Biom files to denote upper class compared to Kingdom. One dataset had mitochondria which broke the import because we have not taken this "sk" class into account

antagomir:
Ok from my side. Just thinking if we should handle those special cases somehow.

TuomasBorman:
This affects only renaming of rowData columns.

1. User can specify whether he/she wants to rename columns
2. Columns are renamed based on match, e.g., "k__" --> "Kingdom"
3. If column is not matched, the column name is unchanged

We had discussion with Felix about TAXONOMY_RANKS in mia. --> we decided to include only "real" taxonomy ranks (ASV and OTU was not added). I think this is little bit similar situation.

antagomir:
Yes, good.

antagomir:
Status of this? Seems related to the discussion in the Tengeler data set PR data import part.

TuomasBorman:
This is good, I will update the branch

This is not directly related to Tengeler data PR --> there is something else happening

antagomir:
To be closed when possible, ideal to not have too many open PRs simultaneously.

antagomir:
Status of this?

TuomasBorman:
This is good to go

antagomir:
Ok. It is most clear if the initiator of the PR can close it. This way we can avoid premature closings. 

Confirm by closing the issue.

PR #364, opened 2023-06-05
Add option to run long tests

0xMuluh:
Added Option to run Long test as a result of timeouts on merida1(mac monterey 12.6.4 x86_64)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.9.2
+Version: 1.9.3
 Authors@R:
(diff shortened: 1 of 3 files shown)


TuomasBorman:
Hello! Thanks! https://contributions.bioconductor.org/long-tests.html Long test requires partition of tests to short and long tests. I guess this will not work as we planned. I guess we have to go back to the drawing board

TuomasBorman:
Well, you can try to make splitOn tests quicker. It seems to be the issue that causes TIMEOUT in mac

0xMuluh:
okay. will do.

Issue #363, opened 2023-06-02
Create a wrapper for the clusterRows method


Basil-Courbayre-Dussau:
The idea is to have a wrapper for bluster, especially the `clusterRows` function to facilitate its use in `mia`.
For now, clustering using bluster can be done this way
```R
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
tse <- transformCounts(tse, method = "relabundance")
x <- t(assay(tse, "relabundance"))
hclust.out <- clusterRows(x, HclustParam())
colData(tse)$clusters <- hclust.out
```

With the wrapper, the code would look like this:
```R
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
tse <- transformCounts(tse, method = "relabundance")

#Either this way with separate methods to cluster on the rows/cols
tse <- clusterRows(tse, "relabundance", HclustParam())

# Or this way with one unique method to cluster with an additional parameter
tse <- cluster(tse, "relabundance", transpose=T, HclustParam())
```
This wrapper would simply take the chosen assay, transpose it if necessary, call the `clusterRows` method and update the `tse` object by adding a `colData` or a `rowData`.

This wrapper could be useful to have a function that resembles the others in the package (in that it has the parameters for the `SummarizedExperiment` object and the assay: `assay_name`). It would allow to call the plotting functions immediately once the `clusterRows` method has been called.

As far as plotting goes, I believe it would stay the same:
```R
tse <- runMDS(tse, assay.type = "relabundance", FUN = vegan::vegdist, method = "bray")

plotReducedDim(tse, "MDS", colour_by = "clusters")
```

One remaining question is how the wrapper would work if we want the full information from the clustering: for example when calling clusterRows with the parameter `full= TRUE`.

This would give information about the clustering that can be later used (to plot the dendrogram for example) but it needs to be stored somewhere. I see two solutions here but there might be more.
* We can store the information in the metadata of the TSE which would be simple to use but this might not be the kind of content that is expected to be in the metadata. In addition, we could create functions to retrieve that information. For example:
```R
result <- cluster(tse, "relabundance", transpose=T, HclustParam(), full= TRUE)
tse <- result$tse
hclust.out <- metadata(tse)[["hclust"]]
```
* Or we could return an object containing the tse and the additional data. But that wouldn't be very handy. For example:
```R
result <- cluster(tse, "relabundance", transpose=T, HclustParam(), full= TRUE)
tse <- result$tse
hclust.out <- result$data
```

[...]

PR #365, opened 2023-06-07
Creation of a wrapper for bluster

Basil-Courbayre-Dussau:
Here is a first version of the wrapper for bluster's `clusterRows`. 
To see the start of the discussion, check out #363.
The code is fairly straightforward. The wrapper uses parameters to know on which assay the clustering should be done, and using the `MARGIN` parameter, it also does the clustering on the features or samples.
The result of the function is stored in `colData` or `rowData` and additional information can be found in `metadata`.

Please tell me if you have any remarks!


Basil-Courbayre-Dussau:
A possibility to address the fact that we can cluster on assays, altExps, ... is that I can create functions (one for each object). Or I can keep one function with a parameter indicating what kind of object should be handled. 
Tell me if you have a preference or other ideas!

antagomir:
I would do this similar to other functions that needed to specify this. @TuomasBorman can advice & review?

R/cluster.R
```diff
@@ -0,0 +1,92 @@
...
+            colData(x)$clusters <- clusters
+        }
+        
+        x
+    }
+)
+
+.check_margin <- function(MARGIN) {
+    if( !is.numeric(MARGIN) && !MARGIN %in% c(1, 2) ){
+        stop("'MARGIN' must be 1 or 2.", call. = FALSE)
```
antagomir: In getExperimentCrossAssociation.R we also use "Samples" and "Features"

In transformCounts we use "samples" and "features"

in addition to MARGIN = 1 or MARGIN = 2.

Should do the same here.

Also harmonize, we should use the same spelling form everywhere. I would go with the lowercase?

R/cluster.R
```diff
@@ -0,0 +1,92 @@
...
+#' the metadata of the object.
+#'
+#' @author Basil Courbayre
+#'
+#' @examples
+#' data(GlobalPatterns)
+#' x <- GlobalPatterns
+#'
+#' # Cluster on rows using Kmeans
+#' tse <- cluster(tse, KmeansParam(centers = 3))
```
antagomir: I would also add example on clustering samples (columns). Use vegan::vegdist options for sample dissimilarity.

Mention explicitly that clustering rows is usually clustering the features.

antagomir (changes requested):

R/cluster.R
```diff
@@ -0,0 +1,92 @@
...
+#'
+NULL
+
+#' @rdname cluster
+#' @export
+setGeneric("cluster", signature = c("x"),
+           function(x, BLUSPARAM, assay.type = assay_name, assay_name = "counts", 
+                    MARGIN=1, full = FALSE)
+               standardGeneric("cluster"))
```
TuomasBorman: atlExp parameter would be nice. See getExperimentCrossCorrelation

R/cluster.R
```diff
@@ -0,0 +1,92 @@
...
+
+#' @rdname cluster
+#' @export
+#' @importFrom bluster clusterRows
+setMethod("cluster", signature = c(x = "SummarizedExperiment"),
+    function(x, BLUSPARAM, assay.type = assay_name, assay_name = "counts", 
+             MARGIN=1, full = FALSE) {
+        # Checking parameters
```
TuomasBorman: name parameter would be also nice so that user can specify what is the name of the data that will be stored to metadata. Check for instance DMM from mia (I guess it has this implemented).

Also: what if the name is already used? I think the function should not overwrite

R/cluster.R
TuomasBorman: Utilize utils.R --> we probably have this check already there

R/cluster.R
```diff
@@ -0,0 +1,92 @@
...
+#'   row-wise / for features (1) or column-wise / for samples (2). Must be \code{1} or
+#'   \code{2}. (By default: \code{MARGIN = 1}) 
+#'   
+#' @param full A logical scalar indicating whether the clustering statistics 
+#'   from both steps should be put in metadata
+#'   
+#' @details
+#' When setting \code{full=TRUE}, the clustering information will be stored in
+#' the metadata of the object.
+#'
```
TuomasBorman: Add info that this is wrapper for bluster package. You can also add link to that package.

TuomasBorman: And what is the output?

TuomasBorman (changes requested):
Good!

R/cluster.R
Basil-Courbayre-Dussau: I haven't seen this in  utils.R but I can add it if you're ok.

R/cluster.R
Basil-Courbayre-Dussau: Should it display an error? Or append a number to the name and proceed? I think I could also do a prompt to insert a new name but that may not be the best idea since I haven't seen any prompt in R so far.

Personally, I prefer the error as the numbers can be confusing and it's easy to get lost in them.

R/cluster.R
Basil-Courbayre-Dussau: When an altExp is given, should I return:
* The altExp with the new data?
* Or the original SE with the altered altExp?

antagomir: original SE?

antagomir: Hmm.. can you check how this is done in other functions that have these arguments and follow similar conventions?

However I think it is relevant question whether we want to operate at the SE level here. If we operate at the (alt)Exp level, one can always insert the function function output in the altExp of the SE with `altExp(se) <- output`.

R/cluster.R
antagomir: Error might be safer than using a different name than user was providing.

Basil-Courbayre-Dussau: Should I also add a name parameter for the colData/rowData?

R/cluster.R
```diff
@@ -10,7 +10,12 @@
 #' @param BLUSPARAM A 
 #'   \code{\link[https://rdrr.io/github/LTLA/bluster/man/BlusterParam-class.html]{BlusterParam}}
 #'   object.
-#'
+#'   
+#' @param altExp A single character value indicating which alternate
```
antagomir: Is this "altExp" or "altexp" in the other functions?

We could use the @inheritParams to avoid duplication and divergence of documentation entries for common function arguments.

R/cluster.R
```diff
@@ -20,16 +25,29 @@
...
-#'   from both steps should be put in metadata
+#'   should be put in metadata.
+#'   
+#' @param name A single character value indicating the name of the clustering 
+#'   info in the metadata.
 #'   
 #' @details
+#' This is a wrapper for the 
+#' \link[https://bioconductor.org/packages/release/bioc/html/bluster.html]{bluster} 
+#' package.
```
antagomir, line 22: I think it is a wrapper for a particular function in that package, not wrapper for the whole package?

R/cluster.R
```diff
@@ -20,16 +25,29 @@
...
+#' 
+#' It returns a \code{SummarizedExperiment} object with clustering information 
+#' named \code{clusters} stored in \code{colData} or \code{rowData}.
```
antagomir: If the input is TreeSE (or SCE) does it return TreeSE (or SCE) respectively? Same input and output types would be essential.

R/cluster.R
```diff
@@ -39,54 +57,81 @@
 #'
 #' # Cluster on rows using Kmeans
 #' tse <- cluster(tse, KmeansParam(centers = 3))
-#'
+#' 
+#' # Clustering done on the samples using Hclust
+#' tse <- cluster(tse, 
+#'                MARGIN="samples",
```
antagomir: brings to my mind: do we also accept "rows" and "cols" as MARGIN? I think it would make sense. That should then apply to all functions that use MARGIN.

antagomir (changes requested):

R/cluster.R
Basil-Courbayre-Dussau: It is altexp, my bad, I'll change that.

R/cluster.R
Basil-Courbayre-Dussau: I agree that it's better for the user. But for the code it isn't great to have multiple values that have the same purpose. For now I've put 'rows', 'features', 'cols', and 'samples' as accepted values but I think we should maybe discuss this further.

R/cluster.R
TuomasBorman: Well, that is true --> check the test in transformCounts --> add that check to utils.R and  and change these unput checks accordingly (also getExperimentCrossAssociation)

R/cluster.R
TuomasBorman: That is true, everything is fine to me --> maybe easier to not implement altExp

R/cluster.R
TuomasBorman: Yes, add name parameter. --> if error, then the name should be checked in the beginning of the function (it would not be nice surprise to wait long time just to see this error)

R/cluster.R
TuomasBorman: In transformCounts, samples and features are the "official" values (used in documentation and in error message). --> 1, 2, rows, cols are also working, though

R/cluster.R
Basil-Courbayre-Dussau: I haven't found any checks for the MARGIN in test-5transformCounts and test-5getExperimentCrossAssociation. 
I also checked test-4splitOn but found nothing.
Should I just add my function to utils.R and check that the value of MARGIN is either "samples", "features", "rows", "cols", 1 or 2?

R/cluster.R
Basil-Courbayre-Dussau: Since I already did, should I let it as is? Or do I remove anything that's altExp related?

R/cluster.R
```diff
@@ -0,0 +1,141 @@
...
+#' @inheritParams bluster::clusterRows
+#' 
+#' @inheritParams runDMN
+#' @inheritParams transformCounts
+#'   
+#' @param altexp A single character value indicating which alternate
+#'   experiment to use for the clustering (if any). If the clustering is done
+#'   on the main experiment, set \code{altexp = NULL}. This is the default 
+#'   choice.
+#' @param data_name A single character value indicating the name of the
```
antagomir: Let us try to resort to _scuttle_ naming conventions (https://github.com/LTLA/scuttle/issues/21):

- dots as separators in argument names
- lower camel case in function names
- upper camel case in class names
- underscores in internal function names

R/cluster.R
```diff
@@ -0,0 +1,141 @@
...
+#' or \code{rowData}. 
+#'
+#' @name cluster
+#' @export
+#' 
+#' @author Basil Courbayre
+#'
+#' @examples
+#' data(GlobalPatterns)
+#' x <- GlobalPatterns
```
antagomir: The GlobalPatterns data is available in two different pkgs in two different formats. Hence, to avoid unintended problems, we prefer here `data(GlobalPatterns, package="mia")`

We usually like to use "tse" for abbreviation of tse objects, such for clarity.

antagomir: Also the next line wouldn't work if you use x - have you run `devtools::run_examples()`?

R/cluster.R
antagomir: The 1/2 is used in many other packages, and I think it would be necessary to have those. In principle we could only have 1/2 and throw out other. However from user perspective, "features" and "samples" or "rows" / "cols" is also pretty clear. I think the order of priority is 1/2 > features/samples > rows/cols. If we remove something, let's start from the right (rows/cols). 

If we have a single auxiliary hidden function to deal with different margin names then this is one-liner to include in functions, and only way to remove it would be to stick to 1/2 notation which would not need any translation then. We could discuss if this is clear enough. With examples, perhaps it would be, but I am not sure.

R/cluster.R
```diff
@@ -0,0 +1,141 @@
...
+                   name="clusters", data_name="clusters") {
+        # Checking parameters
+        .check_margin(MARGIN)
+        .check_data_name(x, data_name, MARGIN)
+        se <- .check_and_get_altExp(x, altexp)
+        if (full) {
+            .check_name(se, name)
+        }
+        .check_assay_present(assay.type, se)
+        assay <- assay(se, assay.type)
```
antagomir: This uses "assay" as both variable and function name. 

I would rename the variable name so that it is not function name. 

For instance, `mat <- assay(...)`

R/cluster.R
Basil-Courbayre-Dussau: Ok, I'm leaving the `assay_name` param as it's preexisting in the rest of the package. But if you think I should change it, tell me.

R/cluster.R
TuomasBorman: Ahh, sorry. I meant https://github.com/microbiome/mia/blob/0cd430a34d9d7bd4c3ef474e4535c179dd1c650a/R/transformCounts.R#L271

BAsed on this, create a .check_margin function --> it could also do the conversion between columns --> 2 https://github.com/microbiome/mia/blob/0cd430a34d9d7bd4c3ef474e4535c179dd1c650a/R/transformCounts.R#L429

and it return FALSE or correct MARGIN? Then we could modify this allowed MARGIN values from one place

R/cluster.R
TuomasBorman: I think you could leave it. Perhaps make it hidden with `...`? We can enable it in the future if we think it is useful

R/cluster.R
```diff
@@ -0,0 +1,156 @@
...
+#' colData(tse)$clusters
+#' 
+NULL
+
+#' @rdname cluster
+#' @export
+setGeneric("cluster", signature = c("x"),
+           function(x, BLUSPARAM, assay.type = assay_name, 
+                    assay_name = "counts", MARGIN="features", full = FALSE, 
+                    name="clusters", data.name="clusters", ...)
```
TuomasBorman: Some has space and some not (assay_name = "counts" & MARGIN="features")

R/cluster.R
```diff
@@ -0,0 +1,156 @@
...
+    }
+    if (length(MARGIN) != 1L 
+        || !(MARGIN %in% c(1, 2, "features", "samples", "columns", 
+                           "col", "row", "rows", "cols"))) {
+        stop("'MARGIN' must equal to either 1, 2, 'features', 'samples', 'columns', 'col', 'row', 'rows', or 'cols'.",
+             call. = FALSE)
+    }
+    MARGIN <- ifelse(MARGIN %in% c("samples", "columns", "col", 2, "cols"), 
+                     2, 1)
+}
```
TuomasBorman: Does this return anything?

R/cluster.R
```diff
@@ -0,0 +1,156 @@
...
+            altExp(x, altexp) <- se
+        } else {
+            x <- se
+        }
+        
+        x
+    }
+)
+
+.get_altExp <- function(...) {
```
TuomasBorman: I am not sure if this current implementation works, at least it is relatively hard to read --> this should do the same?

```
.get_altexp <- function(tse, altexp=NULL, ...) {
   .check_and_get_altExp(x, altexp)
}
```

R/cluster.R
TuomasBorman: It seems that it should

R/cluster.R
```diff
@@ -0,0 +1,156 @@
...
+#' @importFrom bluster clusterRows
+setMethod("cluster", signature = c(x = "SummarizedExperiment"),
+          function(x, BLUSPARAM, assay.type = assay_name, 
+                   assay_name = "counts", MARGIN="features", full = FALSE, 
+                   name="clusters", data.name="clusters", ...) {
+        # Checking parameters
+        MARGIN <- .check_margin(MARGIN)
+        altexp <- .get_altExp(...)
+        se <- .check_and_get_altExp(x, altexp)
+        .check_data_name(se, data.name, MARGIN)
```
TuomasBorman: data.name --> hmmm, how about clust.col?

R/cluster.R
TuomasBorman: Then you could return a list list(tse, altexp) to get the altexp value (which is required in line 105)

R/cluster.R
```diff
@@ -0,0 +1,156 @@
...
+        }
+        
+        # Setting clusters in the object
+        if (MARGIN == 1) {
+            rowData(se)[[data.name]] <- clusters
+        } else {
+            colData(se)[[data.name]] <- clusters
+        }
+        
+        if (!is.null(altexp) && any(dim(se) != dim(x))) {
```
TuomasBorman: This is not explicit why this works, you could add a comment. SummarizedExperiment does not have altExp --> in practice this following code should give an error if altexp is not NULL. However, the altExp and SummarizedExperiment is already checked in the beginning. --> you could add this info so that others do not get confused when maintaining this code

Additionally, remove this any(dim(se) != dim(x)) --> in practice altExp could have same dimensionality

TuomasBorman (changes requested):

R/cluster.R
Basil-Courbayre-Dussau: I am unsure to use col or row as it can apply to both

R/cluster.R
Basil-Courbayre-Dussau: True, it wasn't very clean.
I did a mix of the two versions, please tell me what you think of my next push

Basil-Courbayre-Dussau:
I encountered an issue while working on OMA. With Kmeans, we are supposed to use the distance matrix to do the clustering. But this wrapper will use an assay or an altExp. 
So we might need to reference another object for the clustering. I was thinking of the metadata but I wanted to hear your opinion. The possibilities I can think of are the following:
* Referencing an object in the metadata: `cluster(tse, metadata = "diss", Kmeans(centers = 3))`.
* Passing a `x` parameter to reference the object: `cluster(tse, x = diss, Kmeans(centers = 3))`. That object would then be the one where the clustering is done.
* Not addressing the issue to avoid the over-complexification of the function. The user would then have to user clusterRows with the distance matrix as a parameter and then add the result in colData or rowData: 
```r
x <- clusterRows(assay(tse, "counts"), Kmeans(centers = 3))
colData(tse)$clusters <- x
```
@TuomasBorman @antagomir

TuomasBorman:
I believe kmeans is not commonly used with dissimilarity matrix; at least I haven't heard. In kmeans clustering, means of some features are calculated. If the input is dissimilarity matrix, what are the features? --> I guess it results to unmeaningful results?

(sample 1 and sample 2 might be similar to each other [their similarity is 1] and sample 4 might be similar to sample 9. --> if means are calculated in this feature space, these all aforementioned samples are clustered together even though they are not similar necessarily.

--> so I guess this particular issue is ok. (hierarchical clustering already has dissimilarity parameter)

antagomir:
What original function are speaking about, exactly?

Is it possible to implement similarly to hierarchical clustering, in case distance matrix is required?

Distance matrices can be slow to compute, it is a good feature if clustering function can utilize previously calculated distances. This might be possible in some K-means implementations, I should check and refresh my memories.

Basil-Courbayre-Dussau:
I don't see to which hierarchical clustering function you are referring to, but I haven't encountered one with a dissimilarity parameter. 
As for the fact that Kmeans doesn't use distance matrix often, I encountered this problem in the [OMA book](https://microbiome.github.io/OMA/clustering.html#k-means-clustering), this is why I'm wondering if we need this. For now, in the book, there's an example that uses a distance matrix but when I wanted to change the Kmeans example to use `cluster`, the output wasn't the same as before since the clustering is done on the relabundance matrix and not the distance matrix.
I hope that clarifies things a bit.

antagomir:
In principle, I would be ready to leave out K-means for now if this would require more thinking and work. This method is used very little in microbiome analyses, and hierarchical clustering is sufficient for most of the basic use cases. More can be added later but I do not see immediate need, unless it is very easy to add.

The clustering methods that are specifically used in microbiome context - in particular DMM and PAM/CST approach - are essential to have.

Basil-Courbayre-Dussau:
Okay so is the wrapper ready to be merged?

TuomasBorman (approved):

TuomasBorman:
Couple things that I forgot: 

1. update NEWS file
2. update pkgdown/_pkgdown.yml file (add the function there)

TuomasBorman:
Looks good

antagomir (approved):
Seems good, ready to merge!

PR #366, opened 2023-06-08
loadFromMetaphlan fix

TuomasBorman:
Fixes:

1. Some metaphlan files do not have ID column. Those ones have only clade_name column going to rowData (This "first two columns" did not work)

2. read.table changed column names if number was the first character. It does not do that anymore.


antagomir:
Great!

antagomir (approved):

PR #370, opened 2023-06-16
Add LoadFromHumann

### LINKED ISSUE #369: Humann importer
There is currently no importer for humann data so I thought that it might be nice to have `loadFromHumann`

TuomasBorman:
Here is an importer for Humann output.


NAMESPACE
```diff
@@ -44,6 +44,7 @@ export(getUniqueTaxa)
 export(inner_join)
 export(left_join)
 export(loadFromBiom)
+export(loadFromHumann)
```
antagomir, line 47: I think this name is ok although also forms "HUMAnN" and "humann" are used in the original sources (see e,g, https://github.com/biobakery/humann)

R/loadFromHumann.R
```diff
@@ -0,0 +1,189 @@
+#' Import HUMAnN results to \code{TreeSummarizedExperiment}
```
antagomir, line 1: Just checking: are class names also tagged with `\code{}` ?

R/loadFromHumann.R
```diff
@@ -0,0 +1,189 @@
...
+#'   \item{\code{assay.type}:} {A single character value for naming 
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{assay}} 
+#'   (default: \code{assay.type = "counts"})}
+#'   \item{\code{removeTaxaPrefixes}:} {\code{TRUE} or \code{FALSE}: Should
+#'     taxonomic prefixes be removed? (default:
+#'     \code{removeTaxaPrefixes = FALSE})}
+#' }
+#'
+#' @details
+#' Import HUMAnN (version 3.0) results of pathways or gene families. Input must
```
antagomir: Import HUMAnN (currently version 3.0 supported) results of functional predictions based on metagenome composition (e.g. pathways or gene families).

R/loadFromHumann.R
```diff
@@ -0,0 +1,189 @@
...
+#' be in merged HUMAnN format. (See the HUMAnN documentation and
+#' \code{humann_join_tables} method.)
```
antagomir: specify which package has this function

R/loadFromHumann.R
```diff
@@ -0,0 +1,189 @@
...
+#' object
+#'
+#' @name loadFromHumann
+#' @seealso
+#' \code{\link[=loadFromMetaphlan]{loadFromMetaphlan}}
+#' \code{\link[=makeTreeSEFromPhyloseq]{makeTreeSEFromPhyloseq}}
+#' \code{\link[=makeTreeSEFromBiom]{makeTreeSEFromBiom}}
+#' \code{\link[=makeTreeSEFromDADA2]{makeTreeSEFromDADA2}}
+#' \code{\link[=loadFromQIIME2]{loadFromQIIME2}}
+#' \code{\link[=loadFromMothur]{loadFromMothur}}
```
antagomir, line 51: Should we use @details field to explain what is the difference between `load*` and `make*` functions?

R/loadFromHumann.R
```diff
@@ -0,0 +1,189 @@
...
+#'
+#' @export
+#' @author Leo Lahti and Tuomas Borman. Contact: \url{microbiome.github.io}
+#' 
+#' @references
+#' Beghini F, McIver LJ, Blanco-Míguez A, Dubois L, Asnicar F, Maharjan S,
+#' Mailyan A, Manghi P, Scholz M, Thomas AM, Valles-Colomer M, Weingart G,
+#' Zhang Y, Zolfo M, Huttenhower C, Franzosa EA, & Segata N (2021)
+#' Integrating taxonomic, functional, and strain-level profiling of diverse
+#' microbial communities with bioBakery 3 eLife 10:e65088.
```
antagomir: period after the article title. 

usually journal name with italics (`\emph{... }`?)

antagomir (approved):
Looks good! Approved with comments.

NAMESPACE
TuomasBorman: Yep, our function naming follows this scheme so I agree that this is ok. (I think loadFromHUMAnN might be too hard to remember which letters are in uppercase and which not) --> we can change this also in the future if needed

R/loadFromHumann.R
TuomasBorman: The correct place is to explain that e.g. in `makeTreeSEFromBiom` documentation.
Biom has `load*` and `make*` functions. Otherwise, I cannot actually see what is the difference. Why we have e.g. `makeTreeSEFromDADA2` and not `loadFromDADA2`?

In `*Biom` `load*` creates `TreeSE` from file path and `make*` creates `TreeSE `from biom object.

R/loadFromHumann.R
TuomasBorman: Added link to documentation

R/loadFromHumann.R
TuomasBorman: It seems that the we haven't really unified this. --> It would be good to go through the documentation and unify things (also error messages, e.g., if boolean value is required in input check, the error message is the same in all functions)

R/loadFromHumann.R
antagomir: Right. Is this something to unify (opening an issue)?

R/loadFromHumann.R
antagomir: Right. Is this something to unify (opening an issue)?

PR #378, opened 2023-06-27
Dmm improvement

Basil-Courbayre-Dussau:
Update of the DMM related functions to be more relevant with the new bluster param. Here is the list of functions that can be discussed:

- `calculateDMN`/`runDMN`: The first function outputs the DMN object, the other stores it in the metadata. The calculateDMN function can be deprecated in my opinion as I think its main purpose is to be used in `runDMN`. As for `runDMN`, I think it can be deprecated since it's the same as using the `cluster` function with `DMMParam`. It wouldn't add anything to let this as a wrapper for the `cluster` function.
- `getDMN`: getter for the dmm object in the metadata. I'm not convinced this needs to be here but for now I modified it to get the object in the metadata.
- `bestDMNFit`: returns the index for the best fit. This should be deprecated since the data is accessible in the metadata after doing the clustering: `metadata(tse)$clusters$k`
- `getBestDMNFit`: returns the dmm object for the best fit. This can be kept though it's just a getter.

I'll check that the tests run smoothly once the bluster PR has been merged.

Do you have any opinions on what could be done @TuomasBorman @antagomir ?


TuomasBorman:
I have second thoughts about what I said earlier...

Maybe we should remove all but those that are really necessary --> meaning that we should remove all and trust on the documentation.

1. We have other `calculate*` and `run*` methods, but it is easier to maintain if there are no additional wrappers especially when they do not add much.
2. Makes things maybe little bit easier but on the other hand the data is already easily available.
3. Same thing
4. This makes things little bit easier, however probably not too much. We can have an example on how to fetch this directly from metadata.

antagomir:
Ok to me for now. We can always add more later if there appears to be a need. The examples would be important however.

Basil-Courbayre-Dussau:
I'll update OMA with updated examples for DMM

antagomir:
Checks still failing?

antagomir:
We also have related functions in miaViz: 
https://microbiome.github.io/miaViz/reference/plotDMN.html

-> Can you check and open (and solve?) miaViz issue as relevant?

Basil-Courbayre-Dussau:
> Checks still failing?

I changed the getDMN function so that it would work with the new DMN but I reverted it since that would mean that all previous code would fail. Now the checks should pass

Basil-Courbayre-Dussau:
@antagomir @TuomasBorman 
I've finally managed to make the workflow work by adding `remotes::install_github("LTLA/bluster")` to the workflow. 
I haven't updated DESCRIPTION version as it's a small change.

Is this ok to merge?

R/calculateDMM.R
```diff
@@ -77,28 +77,19 @@
 #' tse <- TreeSummarizedExperiment(assays = list(counts = counts),
 #'                                 colData = colData)
 #'
+#' library(bluster)
+#' 
+#' # Compute DMM algorithm and store result in metadata
+#' tse <- cluster(tse, name = "DMM", DmmParam(),
```
antagomir: I would explicitly show the key DMM functions in the call (e.g. max clusters to test at least)

R/calculateDMM.R
```diff
@@ -77,28 +77,19 @@
...
+#'                MARGIN = "samples", full = TRUE)
+#' 
+#' # Get the list of DMN objects
+#' metadata(tse)$DMM$dmm
+#' # Get which objects fits best
+#' metadata(tse)$DMM$best
+#' # Get the model that fits best
```
antagomir: -> "Get the model that generated the best fit (i.e. this is the model used for the returned clustering)"

antagomir: A main purpose for having access to the model is that it contains cluster assignment probabilities for the samples. That could be used more, if we do not include wrapper for that immediately, I would at least show in the examples how to fetch the sample-cluster assignment probability matrix from the model.

antagomir: Another use if to characterize the components. The examples could also show how to fetch the DMM parameters from the model, in particular the cluster "centroids" (though they are not called centroids in this context I think but instead cluster profiles or something).

antagomir: To simplify readability, those exapmples could first fetch the single top model, then show how to do these two operations (cluster assignments and cluster profiles) for that top model.

antagomir (commented):
Nice! Just the few comments.

R/calculateDMM.R
Basil-Courbayre-Dussau: I don't know to which parameters you are referring to. The parameters accessible from `DMMParam` are: best (the optimal number of clusters), dmm, (seed) and prob (the probability matrix).
I could create a function to compute the centroids or cluster mean, but I don't really know what you mean by cluster profile. Could you explain what you meant?

antagomir: "prob" is the sample-cluster assignment matrix I think. The probabilisties there should some up to 100% per sample. Good to show how to fetch that matrix for the top model.

The cluster profiles mean the cluster-specific Dirichlet probabilities. I think these are included in the dmm (model).

Basil-Courbayre-Dussau: Ok, thanks. I updated the examples to show how to:

- Get the probabilities matrix for the best fit
- Get and display the dmm model that generated the best fit
- Get the weight of each component for the dmm model that generated the best fit

What do you think?

antagomir: Pefect.

antagomir:
Perhaps ready to merge and test further unless there are other open points (besides the failing check..)

Issue #358, opened 2023-04-05
adonis2/dbRDA/permanova automation wrappers?


antagomir:
The OMA section "Testing differences in community composition between sample groups" currently shows how to do dbRDA/PERMANOVA with vegan::adonis2. It might be beneficial to standardize this for TreeSE objects to support reporting (tables) and visualization. Not sure, however, if benefits outweight the costs. To discuss & consider.

TuomasBorman:
Done.

PR #381, opened 2023-06-29
Add signifance testing to RDA/CCA

TuomasBorman:
Related to the issue https://github.com/microbiome/mia/issues/358

SIgnificance is tested with anova.cca. Also the homogeneity of groups is tested (betadisper + anova)

Additionally a bug was fixed (line 243 and 343). If the output is subsetted in RDA/CCA, the result is now MAE with original and subsetted data. The altExp does not work when the samples differ.


antagomir (approved):

TuomasBorman:
Based on [vegan::betadisper documentation](https://rdrr.io/cran/vegan/man/betadisper.html), ANOVA or PERMANOVA are usual tests for betadisper results. Also, TukeyHSD can be performed. I think multiple choices for test can be quite overwhelming for ordinary user, and we should stick to certain test. 

Now PERMANOVA is the default choice, but there are options for ANOVA and TukeyHSD also. The test choice can be stated with `homogeneity_test` parameter which is hidden

I think the most usual case for results is that user wants only the summary table and not the` betadisper` and `PERMANOVA` objects --> `full` parameter is also hidden, but with that user can choose whether to get those objects also

antagomir:
Sounds good. 

Examples could be added to show both cases.

Those who need more capabilities can always proceed to using base functions in whatever way they prefer. Part of the automation is to standardize certain choices as defaults.

Issue #377, opened 2023-06-26
getPrevalence defaults


antagomir:
The getPrevalence argument `as_relative=TRUE` by default.

I realized that this can be confusing when calling the function with "counts" assay, and having it to change the behavior to as_relative by default. This would also be a problem if something else than counts (e.g. log10) would be used as the starting assay. 

I suggest to change the default to `as_relative=FALSE` in all prevalence functions (getPrevalence, getPrevalent* etc.)

antagomir:
After changing also check that OMA explanations will still be correct with respect to the new default. Perhaps no need for changes but needs to be checked.

PR #387, opened 2023-07-06
countDominantTaxa output table

himmil:


R/summaries.R
```diff
@@ -334,8 +334,7 @@ setMethod("countDominantFeatures", signature = c(x = "SummarizedExperiment"),
     tallied_data <- data %>%
         tally() %>%
         mutate(
-            rel.freq = round(100 * n / sum(n), 1),
-            rel.freq.pct = paste0(round(100 * n / sum(n), 0), "%")
+            rel.freq = round(n / sum(n), 3)
```
antagomir: If this is part of the function then provide the rounding digits as a function arguments (setting default to 3 is OK) so that users can ultimately choose as they like.

TuomasBorman (changes requested):
round could be an integer --> that would simplify the code

How about using digits as parameter name? round() has a parameter called that.

Also add input check for round parameter

antagomir:
We should use the same argument name than the `round` function itself, this will be more clear.

R/summaries.R
```diff
@@ -242,14 +242,14 @@ setMethod("getUniqueFeatures", signature = c(x = "SummarizedExperiment"),
 #' 
 #' @export
 setGeneric("countDominantTaxa",signature = c("x"),
-           function(x, group = NULL,  ...)
+           function(x, group = NULL,  digits = NULL, ...)
```
antagomir: If you do not give digits any default then I think it is not necessary to specify this argument, it can be just passed with "..."

R/summaries.R
```diff
@@ -242,14 +242,14 @@ setMethod("getUniqueFeatures", signature = c(x = "SummarizedExperiment"),
...
                standardGeneric("countDominantTaxa"))
 
 #' @rdname summaries
 #' @aliases countDominantFeatures
 #' @export
 setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
-    function(x, group = NULL, round = NULL, ...){
+    function(x, group = NULL, digits = NULL, ...){
```
antagomir: same

R/summaries.R
```diff
@@ -277,7 +282,7 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
         # Add dominant taxa to data
         data$dominant_taxa <- dominant_taxa
         # Gets an overview
-        .tally_col_data(data, group, round, name = "dominant_taxa")
+        .tally_col_data(data, group, digits, name = "dominant_taxa")
```
antagomir: hmm I guess the "name" argument should be also something that the users can customize, i.e. it should be available through the function arguments (either through ... passing if possible, or as explicit argument)

R/summaries.R
```diff
@@ -331,16 +336,13 @@ setMethod("countDominantFeatures", signature = c(x = "SummarizedExperiment"),
         data <- data %>%
             group_by(!!group, !!name)
     }
-    if (is.null(rounding)) {
-      rounding <- 3
-    }
-    else {
-      rounding <- as.integer(rounding)
+    if (is.null(digits)) {
+      digits <- 3
```
antagomir: this default should be given in the function call already, to make it explicit.

however, no rounding might be a good default; then the manpage example could additionally show how to do rounding to other digits,

antagomir (changes requested):

antagomir:
up

antagomir:
@himmil this PR has been open for a month; could we aim to close soon?

antagomir:
This PR has been merged but it says above that "This branch is out-of-date with the base branch", and this PR also hasnt been closed. Can we do that as well, or is something else still missing?

himmil:
Apparently the testing file needs to be edited as well once the output table has changed.

antagomir:
Yes, update all relevant files and complete the PR. Ask for support through Slack where necessary.

antagomir:
Resolve the conversation issues when they become solved. This way we can remove points that have been successfully addressed.

antagomir:
This branch is out-of-date with the base branch

antagomir:
Thanks - one more request.

Could you also deprecate the name `countDominantTaxa` and take into use new name `countDominantFeatures`?

You can check other deprecated functions for an example.

The unit tests and vignette should be updated accordingly, after merging also the OMA examples.

antagomir (approved):

antagomir:
check how to complete..

R/summaries.R
```diff
@@ -275,33 +294,35 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
             data <- data[rep(seq_len(nrow(data)), lengths(dominant_taxa_list)), ]
         }
         # Add dominant taxa to data
-        data$dominant_taxa <- dominant_taxa
+        data[name] <- dominant_taxa
```
TuomasBorman: data[[name]] or data[, name]

R/summaries.R
```diff
@@ -258,11 +265,23 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
                      call. = FALSE)
             }
         }
+        # name check
+        if(!is.null(name)) {
+          if(is.character(name)==F) {
+            stop("'name' must be a character value", call. = F)
+          }
+          # Replace spaces with underscores to enhance usability
+          name <- gsub(" ", "_", name)
```
TuomasBorman: Remove this, it's better to be exactly what user wants

R/summaries.R
```diff
@@ -258,11 +265,23 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
                      call. = FALSE)
             }
         }
+        # name check
+        if(!is.null(name)) {
+          if(is.character(name)==F) {
```
TuomasBorman: Change to --> 

if( is.character(name) && length(name) == 1 )

Always use FALSE and not F. F is just a variable that has default value of FALSE (user can change the value of F to something else and then problems occur).

R/summaries.R
TuomasBorman: More precise

TuomasBorman (changes requested):
Looks good, check comments; just minor things

antagomir:
still on the way?

R/summaries.R
himmil: Needed to make some other changes instead, because column names with spaces caused problems

himmil:
I also added the changes regarding to issues #393 and #377 into this PR, apologies for the inconvenience..

antagomir:
For clarity, perhaps you can announce when this PR is ready for review.

himmil:
It's ready for review.

antagomir:
ok - @TuomasBorman

R/summaries.R
```diff
@@ -258,11 +261,22 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
                      call. = FALSE)
             }
         }
+        # name check
+        if(!is.null(name)) {
+          if(is.character(name)==FALSE) {
+            stop("'name' must be a character value", call. = FALSE)
+          }
+        }
```
TuomasBorman: Remove this nested if --> 


    if(!.is_non_empty_string(name)){
        stop("'name' must be a non-empty single character value.",
             call. = FALSE)
    }

R/summaries.R
```diff
@@ -275,33 +289,37 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
             data <- data[rep(seq_len(nrow(data)), lengths(dominant_taxa_list)), ]
         }
         # Add dominant taxa to data
-        data$dominant_taxa <- dominant_taxa
+        colname <- "dominant_taxa"
+        data[[colname]] <- dominant_taxa
         # Gets an overview
-        .tally_col_data(data, group, name = "dominant_taxa")
+        .tally_col_data(data, group, digits, colname, name)
```
TuomasBorman: This removes the problem of spaces in column  names? --> Good!

One suggestion:

Do not feed name parameter to this function; it makes things little bit harder to read.

Remove these modifications regarding the naming of column, instead, store the table to variable and then change the column name.
```

        data$dominant_taxa <- dominant_taxa
        tab <- .tally_col_data()
        # Change the column name to correct one
```

R/summaries.R
```diff
@@ -258,11 +261,22 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
...
+        # name check
+        if(!is.null(name)) {
+          if(is.character(name)==FALSE) {
+            stop("'name' must be a character value", call. = FALSE)
+          }
+        }
+        
+        # digits for rounding
+        args <- list(...)
+        digits <- args$digits
```
TuomasBorman: THis line can be removed if you modify line
`
.tally_col_data(data, group, digits, colname, name)`

to 
`
.tally_col_data(data, group, digits, colname, name, ...)`

AND line

`.tally_col_data <- function(data, group, digits, colname, name){`

to
`
.tally_col_data <- function(data, group, digits = NULL, colname, name, ...){`

Your solution works, but for consistency (we have used this latter approach in other functions)

R/summaries.R
```diff
@@ -275,33 +289,37 @@ setMethod("countDominantTaxa", signature = c(x = "SummarizedExperiment"),
...
 )
 
+
 ################################ HELP FUNCTIONS ################################
 
 #' @importFrom S4Vectors as.data.frame
 #' @importFrom dplyr n desc tally group_by arrange mutate
-.tally_col_data <- function(data, group, name){
+.tally_col_data <- function(data, group, digits, colname, name){
     # Convert data to data.frame
```
TuomasBorman, line 320: You could add a check for `digits`. It must be an integer or NULL

R/summaries.R
```diff
@@ -321,23 +339,27 @@ setMethod("countDominantFeatures", signature = c(x = "SummarizedExperiment"),
...
     tallied_data <- data %>%
         tally() %>%
         mutate(
-            rel.freq = round(100 * n / sum(n), 1),
-            rel.freq.pct = paste0(round(100 * n / sum(n), 0), "%")
+            rel.freq = n / sum(n)
         ) %>%
         arrange(desc(n))
+    if(!is.null(digits)) {
+      tallied_data["rel.freq"] <- round(tallied_data["rel.freq"], digits)
```
TuomasBorman: @antagomir SHould the `rel.freq` be `rel_freq`? By default, taxa column is named `dominant_taxa` (with underscore)

Or then we could change `dominant_taxa` to `dominant.taxa`. I think we do not have naming convention to tables

TuomasBorman (changes requested):

R/summaries.R
himmil: Just letting you know that I've decided column name defaults to be with underscore and PR is ready for review if you agree

antagomir: nice!

antagomir (approved):

antagomir:
Hmm seems merging is blocked until @TuomasBorman approves too. 

(we could see if it is possible to remove the _requirement_ of multiple approvals)

TuomasBorman (approved):

himmil:
I don't have write access so could you either merge it for me or give me the permission?

PR #395, opened 2023-07-19
Readme

antagomir:
Added development instructiont to mia README. 

This could be useful for newcomers as a starting point. 

We can also consider other locations for this as relevant but here now to get started.


TuomasBorman (approved):
Very good, useful info

But is the right place the contribution information? https://github.com/microbiome/mia/blob/master/CONTRIBUTING.md

antagomir:
True. Good point. Will move it there. Linking from README.

Issue #394, opened 2023-07-19
Rename transformCounts to transformAssay


antagomir:
The `transformCounts` is somewhat misleading name bc this function allows transformation for any assay (not just counts assay). 

Let us change the name to `transformAssay` and deprecate `transformCounts`.

Remember to update
- unit tests
- manpage
- vignettes
- oma
- miaViz
- miaSim
- miaTime

antagomir:
Merged - remember to update the other resources as well.

It might be easiest to do this after we first updated this mia version to Bioconductor devel; we have many updates recently and it is good to bring them there for further testing (if not done already) ping @TuomasBorman when the time allows?

TuomasBorman:
Bioc devel should be updated in couple of days

antagomir:
Still on thing to fix.

Load example data first:

```
library(mia)
data(GlobalPatterns, package="mia")
```


Then running `transformCounts(GlobalPatterns)` gives:
> Error in transformCounts(GlobalPatterns) : 
>   could not find function "transformCounts"

-> This means that the `transformCounts` has not been deprecated properly. The old function name should continue working during some transition period. Looking at the code, [line 236 onwards](https://github.com/microbiome/mia/blob/master/R/transformCounts.R#L236) it seems that this actually should throw a warning but work. Investigate why it throws an error instead or warning. I also think that line 237 should have `.Deprecate` message instead of `warning`?

antagomir:
@ake123 can you check?

TuomasBorman:
This is also done @ake123 ? And this and other functions are deprecated properly? (Old ones are still available but function gives deprecation message)

ake123:
it was done!

antagomir:
can you close issues that are ready? after checking that deprecations are OK in mia, and function names corrected also in miaViz, miaTime, miaSim, OMA (if used in those)

ake123:
All fixed!

PR #396, opened 2023-07-20
change transformCounts to transformAssay

ake123:
change transformCounts to transformAssay (Rename transformCounts to transformAssay #394)


R/transformCounts.R
```diff
@@ -337,7 +337,7 @@ setMethod("transformFeatures", signature = c(x = "SummarizedExperiment"),
              pseudocount = 0,
              ...){
 
-        .Deprecated("transformCounts")
+        .Deprecated("transformAssay")
```
antagomir, line 345: Make sure that the `transformCounts` is still available if users call that but it should throw the informative warning. You can implement transformCounts as a wrapper that directly calls transformAssay after throwing the warning. I think there are some other examples of such function deprecation.

Make sure that this generates an informative error message, suggesting to change the function name in user workflows. You can check the other deprecated functions for examples on what kind of warnings they throw.

antagomir (changes requested):
Looks good overall.

R/transformCounts.R
antagomir: Close these as they become resolved..

R/transformCounts.R
```diff
@@ -211,6 +211,7 @@ setMethod("transformSamples", signature = c(x = "SummarizedExperiment"),
         # Call general transformation function with MARGIN specified
         x <- transformAssay(x = x, assay.type = assay.type,
```
antagomir, line 212: In examples we use `se` instead of `x` as we like to highlight the fact that these methods work apply to SE objects.

Can you still update the manpage examples with that?  Just changing x to tse. No need for other changes than the examples.

antagomir (approved):

antagomir (approved):
Great. Merge if ready.

antagomir:
(out of sync with base branch still)

Issue #343, opened 2023-02-20
altExp support for meltAssay 


antagomir:
It is not possible to choose altExp for meltAssay currently.

One can however call `meltAssay(altExp(tse, ...))` to do this for any altExp. Perhaps this is enough and there is no need to complicate the function?

An example could be added on the manpage, however as there have been questions about it.

antagomir:
altExp mechanism seems to work routinely also in `transformCounts` even though it is not defined in that function (i.e. it just works because the function implementation knows how to pass this argument further and make the selection; see transformCounts source code for an example).

```
library(mia)
data(GlobalPatterns)
tse <- GlobalPatterns
altExps(tse) <- splitByRanks(tse)
transformCounts(tse, assay.type="counts", method="relabundance", altexp="Genus")
```

For consistency let us make sure it works this way also for `meltAssay`. 

Importantly, it should be (?) possible to add this implicitly by doing similar programming than in `transformCounts`, instead of explicitly declaring altexp as a function argument. Inspect the `transformCounts` source code on how to do this.

antagomir:
@ake123 I suggest you try, ask @TuomasBorman if you bump into problems.

ake123:
This is solved!

PR #403, opened 2023-07-25
altExp support for meltAssay 

ake123:
altExp support for meltAssay #343


antagomir:
Remember to:
- upgrade version number in DESCRIPTION
- update NEWS file
-  squash and merge if ready

ake123:
Is the version other than 1.9.9 ?

antagomir:
The DESCRIPTION file in master/main branch gives the latest version (it is indeed 1.9.9). 

The numbering goes in numeric order so the next minor update is 1.9.10 etc.

antagomir:
I add this now anyway. Not so critical here.

PR #406, opened 2023-07-27
rowData clean-up from character artifacts

### LINKED ISSUE #303: Taxonomy parsing
Consider using these parsing functions in examples, or having analogous functions to 
- `phyloseq::parse_taxonomy_greengenes`
- `phyloseq::parse_taxonomy_qiime`
- `phyloseq::parse_taxonomy_default`

Currently OMA examples handle this with:

```
# Goes through the whole DataFrame. Removes '.*[kpcofg]__' from strings, where [kpcofg] 
# is any character from listed ones, and .* any character.
rowdata_modified <- BiocParallel::bplapply(rowData(se), 
                                           FUN = stringr::str_remove, 
                                           pattern = '.*[kpcofg]__')
```

Which is handy but wrappers might help less experienced users.

ChouaibB:
Hi,

This is related to the #303 discussion.
A solution draft for cleaning up feature_data/rowData loaded from biom files that might contain some character artifacts (e.g. `"`).
The attached pdf file represent tests of the function in question `makeTreeSEFromBiom` (which was also based on the example test at #303 ).

If it seems fine, I could update the examples at documentation using the function's argument, make unit tests, and bump the version.

Thanks.

[makeTreeSEFromBiom_local_test.pdf](https://github.com/microbiome/mia/files/12182870/makeTreeSEFromBiom_local_test.pdf)


R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -97,6 +97,10 @@ makeTreeSEFromBiom <- function(
         # Add correct colnames
         colnames(sample_data) <- colnames
     }
+    # Clean feature_data or rowData might contain possible character artifacts;
+    # so as colnames and feature_data could be parsed safely afterwards,
+    # we define here possible character artifacts that might occur.
+    patterns <- "\""
```
antagomir: How about making this an argument (the user could ultimately choose)?

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -122,6 +130,13 @@ makeTreeSEFromBiom <- function(
         feature_data <- do.call(rbind, feature_data)
         # Add correct colnames
         colnames(feature_data) <- colnames
+    # Otherwise if it is already a data.frame clean from artifacts
+    } else if (is(feature_data, "data.frame")) {
+        feature_data <- apply(
+            feature_data,
+            2,
+            gsub, pattern = patterns, replacement = "")
+        feature_data <- as.data.frame(feature_data)
```
antagomir: Remember that in SE the rowData is a DataFrame (not data.frame)

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: The default can be whatever seems to cover most cases.

antagomir:
checks to fix

ChouaibB:
> checks to fix

True, when I checked the job run log it seemed that it was related to other functions. And there seems to be not test script for `makeTreeSEFromBiom`. That's why I re-ran it, I stop the run now and do the updates first.

antagomir:
It is possible to place testdata in the package in inst/extdata/ but not sure if this is necessary for now.

ChouaibB:
> It is possible to place testdata in the package in inst/extdata/ but not sure if this is necessary for now.

Sure, I could give it a try and add the tests for `makeTreeSEFromBiom`  if otherwise changes to the function are ok?

antagomir:
Hmm - one more change: it would be good to explicitly detect if there are such non-standard characters to remove, then throw a warning if such characters are being removed? This would make the process more safe. By default there should not be such extra characters, I think, and if there are it may indicate some problems that the users may wish to be aware of.

That new file can be called with:
`system.file("extdata/myfilename.end", package="mia")`

ChouaibB:
> Hmm - one more change: it would be good to explicitly detect if there are such non-standard characters to remove, then throw a warning if such characters are being removed? This would make the process more safe. By default there should not be such extra characters, I think, and if there are it may indicate some problems that the users may wish to be aware of.
> 
> That new file can be called with: `system.file("extdata/myfilename.end", package="mia")`

That's actually true, thanks. Is it ok to keep the default "artifact" pattern to detect and remove as `\"` and then generalize it by the time? bit hard for me to know what kind of non-wanted (artifact) characters that might occur in the taxonomy data within a biom file :/ ; unless ofc if the user priorly knows and could use the `cleanTaxaPattern` argument, and eventually throw a warning about the removal?

I update the system.file too, thanks.

ChouaibB:
> > Hmm - one more change: it would be good to explicitly detect if there are such non-standard characters to remove, then throw a warning if such characters are being removed? This would make the process more safe. By default there should not be such extra characters, I think, and if there are it may indicate some problems that the users may wish to be aware of.
> > That new file can be called with: `system.file("extdata/myfilename.end", package="mia")`
> 
> That's actually true, thanks. Is it ok to keep the default "artifact" pattern to detect and remove as `\"` and then generalize it by the time? bit hard for me to know what kind of non-wanted (artifact) characters that might occur in the taxonomy data within a biom file :/ ; unless ofc if the user priorly knows and could use the `cleanTaxaPattern` argument, and eventually throw a warning about the removal?
> 
> I update the system.file too, thanks.

Sorry for the many replies :)
I was just thinking with my humble experience with Taxonomy data, it always contain alphabets (lower and upper cases), and numeric characters (alphanumeric characters) plus underscores and/or dashes, anything else should be an artifact to detect and remove (except the characters used to parse data into table e.g `sep=;` or `sep=|`), building such a regex do you think it would be safe and generalized for detection and removal (while ofc throwing the warning about it)?

antagomir:
Ok that could be the default now - but it should be argument that the user can tweak if necessary. We can later remove this if it turns problematic. The real solution would comply with biom standard, I am not sure what that says about the extra characters. There are pros & cons but in a way we also want to make this fluent to users so that they can focus on analysis of the data.

ChouaibB:
> Ok that could be the default now - but it should be argument that the user can tweak if necessary. We can later remove this if it turns problematic. The real solution would comply with biom standard, I am not sure what that says about the extra characters. There are pros & cons but in a way we also want to make this fluent to users so that they can focus on analysis of the data.

Before implementing anything, I was still curious about a way to detect non-standard character artifacts in Taxonomy data, so I made a small testing around based on an example file (attached in here):
`
# Example Data 

Data was copied from Aggregated_humanization2.biom (manually through bash 
command less); the section corresponding to taxonomy data.
Then saved to a text file; test_text.txt


## Reading data

Just as an example to test the search for artifacts, the text was split based on 
"id", then each resulting list was made as a character vector:

> test_text <- readr::read_lines("test_text.txt") %>% stringr::str_split("id") %>% 
    lapply(., function(l) paste(unlist(l), collapse = " "))

## Testing the regular expression on one example of the test_text

Retrieving taxonomy data including the separators that exists in taxonomy data 
which should be kept to parse them as a table afterwards.

Testing with one example, retrieving the taxonomy information usually wanted:

> grep("[[:alnum:]]|[-_,;|]|[[:space:]]",
     test_text[[1]] %>% stringr::str_split("") %>% unlist(),
     invert = FALSE, value = TRUE) %>% paste0(collapse = "")
     
result -> "  1726470, metadata taxonomy k__Bacteria, p__Bactero etes, c__Bactero ia, o__Bactero ales, f__Bactero aceae, g__Bactero es,  1726471, metadata "

Testing the negative (or invert) of the regular expression to retrieve the 
artifacts:

> grep("[[:alnum:]]|[-_,;|]|[[:space:]]",
     test_text[[1]] %>% stringr::str_split("") %>% unlist(),
     invert = TRUE, value = TRUE) %>% paste0(collapse = "")
result -> "[{\"\":\"\"\"\":{\"\":[\"\\\"\"\"\"\"\"\"\"\"\"\"\\\"\"]}}{\"\":\"\"\"\":{\""

To collect the unique list of artifacts to be cleaned later:

> grep("[[:alnum:]]|[-_,;|]|[[:space:]]",
     test_text[[1]] %>% stringr::str_split("") %>% unlist(),
     invert = TRUE, value = TRUE) %>% unique()
result -> "["  "{"  "\"" ":"  "\\" "]"  "}"

## Testing the regular expression on the whole test_text

Collecting the final unique list of character artifacts to be cleaned throughout
 the whole Taxonomy data:
 
> lapply(test_text, function(x) {
    grep("[[:alnum:]]|[-_,;|]|[[:space:]]",
         x %>% stringr::str_split("") %>% unlist(),
         invert = TRUE, value = TRUE) %>% unique()
}) %>% unlist() %>% unique()

result -> "["  "{"  "\"" ":"  "\\" "]"  "}"

Consequently, this could be perhaps an automated way of detecting, and forming
the patterns to be cleaned at the "makeTreeSEFromBiom" (which automatically 
will become 'cleanTaxaPattern="[{\":\\]}"') if for example the user
provide "cleanTaxaPattern=NULL" or smt similar (maybe even giving an value option 'auto' for instance),
 to hopefully safely parse the taxonomy data, which would safely parse rank names, taxonomy data afterwards in the function. 
`
[test_text.txt](https://github.com/microbiome/mia/files/12196337/test_text.txt)

Ofc with the current implementation the `biomformat::observation_metadata` does some parsing and cleaning of some of the above seen artifacts in the raw data file. The idea I was thinking if this seems reasonable, a further utility/helper function(s) I could implement that would double check by detecting, cleaning and throwing the warning.
P.S if it seems interesting please some hints how to improve the regex used in the example, thanks :)

antagomir:
Can you summarize in a much shorter way what is the problem and what is the proposed solution?

ChouaibB:
> Can you summarize in a much shorter way what is the problem and what is the proposed solution?

I will be committing the implementation I had in mind, then I guess it would show clearly.

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -14,8 +14,10 @@
 #' (default \code{rankFromPrefix = FALSE})
 #' 
 #' @param cleanTaxaPattern \code{character} or \code{regex}: If file have
```
antagomir: for function arguments we use style `clean.taxa.pattern`, however here I suggest shorter name like `clean.names`

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -37,16 +39,23 @@
...
 #'   x1 <- biomformat::read_biom(rich_dense_file)
 #'   se <- makeTreeSEFromBiom(x1)
 #'   # Convert SE to TreeSE
 #'   tse <- as(se, "TreeSummarizedExperiment")
 #'   tse
+#'   
+#'   # Cleaning artifacts from Taxonomy data
+#'   biom_object <- biomformat::read_biom(
+#'       system.file("extdata/testdata/Aggregated_humanization2.biom",
+#'                   package="mia"))
```
antagomir: might be more clear to first specify this as file name 

`f <- system.file(...)`

and then on separate line run `read_biom(f)`

is users want to copypaste the example they can just change the file name

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: I know we are currently not systematic on this but let us do this at least when we introduce new argument names

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -112,17 +125,20 @@ makeTreeSEFromBiom <- function(
...
-        feature_data <- lapply(
-            feature_data,
-            gsub, pattern = cleanTaxaPattern, replacement = "")
+        
+        # In case the list of taxa info have different lengths
+        if (!all(lengths(feature_data)==max_length)){
         # Get the column names from the taxa info that has all the levels that occurs
         # in the data
         colnames <- names( head( feature_data[ lengths(feature_data) == 
                                                    max_length ], 1)[[1]] )
```
antagomir: the "head" seems unnecessary here?

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
     return(colname)    
 }
+
+# Detect and clean non wanted characters from Taxonomy data if needed.
+.detect_taxa_artifacts_and_clean <- function(x, patterns) {
+    
+    # No cleaning
+    if (is.null(patterns)) {
+        warning("No Taxonomy cleaning was done to the data.",
+                call. = FALSE)
```
antagomir: Is it necessary to throw a warning on this - I think no extra cleaning should be the safe default, assuming that the biom file follows the standards? I would not throw warning when the file is simply being read as expected. Warnings are warranted if extra cleaning is done automatically.

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+    }
+    # Automatic cleaning
+    else if (patterns=="auto") {
+        # General regex pattern
+        PATTERN <- "[[:alnum:]]|-|_|\\[|\\]|,|;\\||[[:space:]]"
+        patterns <- .detect_taxa_artifacts(x, PATTERN, invert=TRUE)
+        # No artifacts found
+        if (patterns=="") {
+            warning("No artifacts found in the Taxonomy data.",
+                    call. = FALSE)
```
antagomir: This should be a "message" rather than a warning. 

However I would perhaps have a preference for just no extra messages when everything goes well here and there are no artifacts.

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+        # Clean from artifacts
+        } else {
+            x <- .clean_from_artifacts(x, patterns)
+            warning("The following artifacts: '", patterns, "' were cleaned from 
+                    the Taxonomy data.",
```
antagomir: "from the Taxonomy data" -> "from rank names in taxonomy data"

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+                    call. = FALSE)
+        }
+    # Clean with the character or regex provided
+    } else {
+        pattern <- .detect_taxa_artifacts(x, patterns=patterns)
+        # patterns provided not found
+        if (pattern=="") {
+            warning("The '", patterns, "' provided at 'cleanTaxaPattern' were
+                    not found in the Taxonomy data.",
+                    call. = FALSE)
```
antagomir: see above

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+        # patterns found and cleaned
+        } else {
+            x <- .clean_from_artifacts(x, pattern)
+            # warn what was cleaned
+            warning("The following artifacts: '", pattern, "' were cleaned from 
+                    the Taxonomy data.",
+                    call. = FALSE)
```
antagomir: see above - this seems repetitive compared to the above, is there a way to call the same detect & clean code just once?

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -224,3 +244,77 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+    }
+    return(patterns)
+}
+
+.clean_from_artifacts <- function(x, patterns) {
+    if (is(x, "list")) {
+        x <- lapply(x, gsub, pattern = patterns, replacement = "")
+    } else if (is(x, "data.frame")) {
+        x <- apply(x, 2, gsub, pattern = patterns, replacement = "")
+    }
```
antagomir, line 304: also add "else" statement to ensure that all cases are covered. Even if it does not perform specific  operations.

Or if the type must be list or data.frame then this could throw a warning about a wrong type (although the code should catch that already earlier in principle)

antagomir (changes requested):
nice!

antagomir:
@ChouaibB note that @RiboRings is now making some changes in OMA
https://github.com/microbiome/OMA/pull/316

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: Sure, I would prefer `clean.taxa.pattern` if it's ok? Because the operation targets only `rowData` not `colData`. Which I have been giving a thought too if necessary to parse it in such a way, sanity check e.g. cleaning numeric column from possible artifacts too and cast it to numeric, perhaps same for  character columns that are e.g supposed to be factors and such. In this case I would see `clean.names` perhaps generic  for  `rowData` and `colData` ofc for example adding an argument which TreeSE (or SE) slot is in question.

antagomir: How about `clean.taxa.names` - could be a bit more intuitive?

The warning message could also mention what value to set for this in case user wants to keep the original names straight from the file?

ChouaibB: Yes, it is, thanks :) .
Sorry, I might have not understood clearly. But do you mean instead of having removed (or substituting with `""`) the patterns mentioned at `clean.taxa.names`, rather add another argument e.g. `clean.taxa.names.replacement` and make the default `clean.taxa.names.replacement=""`?

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: Honestly, it was existing in the implementation. The [`feature_data <- biomformat::observation_metadata(obj)`](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L98) might result in a dataframe or a list. But the structure of the list returned was not clear to me. Following the rest of the original code after its loading; I understood that it's just a list of taxonomies present. I will try look for some example online when it's loaded as a list to understand the structure better; it was not very clear to me neither :/

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: I am not sure if I understand the question but the idea was that if the names would be manipulated in any possible way, then there should be a warning.

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: I just mean that the result is equivalent with or without "head()". If it does not add anything you can remove the head() part from the code?

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: Ahh ok, yes as I am updating the function now, I am keeping the warnings only when cleaning happens otherwise no warning is thrown.

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: I agree, I just need if it's ok, to test it with some real file example with `feature_data` being loaded as a list. So far when I have been developing and testing it as a list;  I was just using the `extdata/testdata/Aggregated_humanization2.biom` which returns  `feature_data` as a dataframe. And for the sake of testing I was manually casting it into a list to check that the part of the code that processes `feature_data` as a list is working. Or shall I just remove the `head` part and test it's behavior by similarly casting that `feature_data` as a list?

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: My intention was that when user is choosing "auto" cleaning, inform the user about all went well with no artifact found. Shall I change it to a message or remove it?

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: umm... this part goes throughout the whole `rowData` content not only taxonomy rank names (also looks at the actual species names, genera  names and so on). What could be more informative warning or message please?

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: Whatever works. My comment only meant that the head() in that call seemed irrelevant to have.

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: I would only throw a warning if something is modified, otherwise none.

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: ok

antagomir: then no need to change, or could you say "rowData" instead of Taxonomy data? Would that be less ambiguous?

ChouaibB: True it will be, thanks.

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: Yes, I will update.

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: What I understood from original implementation is that `feature_data` could be a list, dataframe or null. I didn't choose to call any cleaning when it's null [at that point](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L124C70-L124C70), `feature_data`  is constructed as `S4Vectors::make_zero_col_DFrame(nrow(counts))` which means empty content with nothing to clean?
And I guess the code does catch that earlier [at](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L123) ?

R/makeTreeSummarizedExperimentFromBiom.R
ChouaibB: At that part [of the code ](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L139-L140) it was originally assumed that the resulting `feature_data` is a list with unequal lengths.
For testing; so I used the same testing procedure mentioned above (casting the same `feature_data` from dataframe to list), however I changed `feature_data` as a list in a way the lengths are unequal. Then run [the code that handle `feature_data` as a list](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L127-L159) (results at screenshot). it seems using: 
`colnames <- names( head( feature_data[ lengths(feature_data) ==max_length ], 1)[[1]] )`
or
`colnames <- names( feature_data[ lengths(feature_data) ==max_length ][[1]] )`
Are producing null as colnames. But rather just calling `colnames <- names(feature_data)` would do the job; only if the originally loaded `feature_data` as a named list.
image

P.S. same file as used for testing before `extdata/testdata/Aggregated_humanization2.biom`

ChouaibB: I was just thinking, if in the case that the `feature_data` loaded from a biom file with unequal vectors whiting it (supposing that always the list structure is the same; a vector for each taxonomy rank in that list) but with no names.
Could the following for example be safer? (highlighted part):
 
image

antagomir: seems feasible

R/makeTreeSummarizedExperimentFromBiom.R
antagomir: ok?

ChouaibB:
One question/suggestion @antagomir @TuomasBorman :
If this implementation seems to be fine, should [`.detect_taxa_artifacts_and_clean` ](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L247-L276)and [`.detect_taxa_artifacts`](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L279-L297) and [`.clean_from_artifacts`](https://github.com/microbiome/mia/blob/issue_303/R/makeTreeSummarizedExperimentFromBiom.R#L299C1-L309) be moved to `R/taxonomy.R` or `R/utils.R` that potentially could help check the taxonomy parsing of other types of data (e.g. `makeTreeSummarizedExperimentFromDADA2.R` or `makeTreeSummarizedExperimentFromPhyloseq.R` ) or such?

antagomir:
You can move if it seems useful but I think it should be possible to use these regardless of the file where they are located.

ChouaibB:
Ahh ok, I will keep them where they are for now.
Regarding the failed checks ([link](https://github.com/microbiome/mia/actions/runs/5730212242/job/15528536310?pr=406)):
image
It seems they are not related to the tests of `makeTreeSummarizedExperimentFromBiom.R` (actually I guess it fails before even reaching the tests related of `test-10makeTreeSEFromBiom.R`; locally these tests all pass with no problem), is somebody working on debugging those test checks that fails or shall I also look at them so this can be potentially merged if all is ok otherwise?

antagomir:
Strange that we have such errors suddenly appearing. Would be great if you can fix on the same go?

ChouaibB:
> Strange that we have such errors suddenly appearing. Would be great if you can fix on the same go?

I will give it a try.

ChouaibB:
About the failed checks:
When I did the checks locally I took some notes of the reason of the error and the fix:
```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-0diversity.R:27:5'): diversity estimates ─────────────────────
`lambda` not equal to .simpson_lambda(assays(tse_idx)$relabundance).
names for current but not for target
── Failure ('test-0diversity.R:28:5'): diversity estimates ─────────────────────
`ginisimpson` not equal to colData(tse_idx)$gini_simpson.
names for current but not for target
── Failure ('test-0diversity.R:29:5'): diversity estimates ─────────────────────
`ginisimpson` not equal to .calc_gini_simpson(assays(tse_idx)$relabundance).
names for current but not for target
── Failure ('test-0diversity.R:31:5'): diversity estimates ─────────────────────
`invsimpson` not equal to colData(tse_idx)$inverse_simpson.
names for current but not for target
── Failure ('test-0diversity.R:32:5'): diversity estimates ─────────────────────
`invsimpson` not equal to .calc_inverse_simpson(assays(tse_idx)$relabundance).
names for current but not for target
## ==> solution was to unname the named returned vector of each right-hand 
## side of the test 

── Failure ('test-0diversity.R:66:5'): diversity estimates ─────────────────────
`test1` not equal to `test2`.
Names: 3 string mismatches
## ==> solution was to unname both since they have same values but different 
## naming of the vectors at each side of the test

── Failure ('test-0diversity.R:67:5'): diversity estimates ─────────────────────
round(test1, 6) not equal to c(7.256706, 6.098354, 7.278894).
names for target but not for current
## ==> solution was to unname left-hand side of test otherwise values were equal

── Failure ('test-8subsample.R:35:5'): subsampleCounts ─────────────────────────
colSums2(assay(tse.subsampled, "subsampled")) not equal to `expColSums`.
names for target but not for current
## ==> solution was to unname left-hand side of test otherwise values were equal

── Failure ('test-8subsample.R:51:5'): subsampleCounts ─────────────────────────
colSums2(assay(tse.subsampled.rp, "subsampled")) not equal to `expColSumsRp`.
names for target but not for current
## ==> solution was to unname left-hand side of test otherwise values were equal  
```
Hopefully now it passes the checks at github :)

antagomir (approved):
merge when ready unless there are further comments

ChouaibB:
> merge when ready unless there are further comments

I will merge it for now. Perhaps when the utility/helper functions within this one turn to be useful at other importers, they could be moved to `taxonomy.R`. I was just thinking that, by then for future development any taxonomy related helper functions could be found at `taxonomy.R` so the functions and utilities and the type of utilities needed could be "easily readable and found" under  `R/`  then all the developers instead of going through each function and searching for their helper function that could be re-used; then naming the type of utility functions under `R/` could be helpful and more easily readable. But this is just an opinion ofc :)

Issue #400, opened 2023-07-25
merge by variable name


antagomir:
Enable rowData variables in mergeRows. 

Currently, the groups must be fed as a vector (f parameter) but it would be nice to be able to specify the grouping variable directly from rowData as a column name. 

Same then for mergeCols.

Ping @Daenarys8 

For some background discussion see #389

0xMuluh:
okay noted. I was a bit confused which to do first, the variable change or the deprecation. I will do the deprecation then do a separate pr for the variable change.

antagomir:
Right, it is a bit in the air. Sounds good.

antagomir:
Closed via #408

Issue #397, opened 2023-07-21
 Enabling rowData variables in mergeRow


0xMuluh:
see: issue: #389 
We discussed about enabling rowData variables in mergeRows. Currently, the groups must be fed as a vector (f parameter) but it would be nice to be able to specify the grouping variable directly from rowData as a column name. (Same for mergeCols).

antagomir:
Agreed.

TuomasBorman:
Can you @Daenarys8 create a draft for this?

0xMuluh:
definitely.

antagomir:
Closed via #408

PR #408, opened 2023-07-28
merge by rowData

0xMuluh:
see #397 and #400


R/merge.R
```diff
@@ -103,6 +103,16 @@ setGeneric("mergeSamples",
            function(x, f, archetype = 1L, ...)
                standardGeneric("mergeSamples"))
 
+# Helper function to check if 'f' is a valid column name in rowData or colData
+.is_valid_column_name <- function(x, f, dim_type) {
+    if (dim_type == "rows") {
```
antagomir: by the way could we easilyl use `dim.type` instead of `dim_type`?

antagomir: Why we don't here use MARGIN (as usual for rows vs. columns; or features and samples)? See other functions for such cases. Consistency could be desirable.

antagomir (commented):
Seems good but see the comment on argument name. 

Checks still fail.

R/merge.R
0xMuluh: Oh this function is removed. I was testing several approaches.

0xMuluh:
> Seems good but see the comment on argument name.
> 
> Checks still fail.

I'm not sure about the fail, but test run fine on local. will check it out

antagomir:
up

antagomir:
ready?

0xMuluh:
> ready?

Yes.

antagomir:
Great. Are we now ready to address
https://github.com/FrederickHuangLin/ANCOMBC/issues/174

0xMuluh:
> Great. Are we now ready to address [FrederickHuangLin/ANCOMBC#174](https://github.com/FrederickHuangLin/ANCOMBC/issues/174)

Ohh yea.  Would make a PR

antagomir:
Status of this one?

PR #422, opened 2023-08-18
RDA, return weighted averages or linear combination scores

TuomasBorman:
CCA includes linear combinations ("u") values and weighted average ("wa") values. Currently, linear combinations are returned in runCCA, although, cca object is as an attribute, so all the data can be accessible from the cca object.

However, plotReducedDim for instance plots the matrix in reducedDim meaning that it plots the linear combination values even though weighted averages might be more interesting.

I added an option to choose the desired values (weighted averages as default).

Comments?


antagomir:
Sounds good to me at least and we can always refine later if new needs will arise. The default is something worth considering - which one that should be.

TuomasBorman:
So what I understood https://rdrr.io/rforge/vegan/man/cca.object.html

Ordination scores:

u --> correlation between sample metadata variables
v ---> correlation between features
wa --> u weighted with v

In this case, I think we should allow these all options, but wa should be the default

antagomir:
wa can be default; this is "scores" and "loadings" from the model specification, rather than correlations in traditional sense but yes.

TuomasBorman:
New parameter `scores` which can get values `"wa"` or `"u"`.

Issue #372, opened 2023-06-18
Unify documentation


TuomasBorman:
To polish the package, the documentation should be unified (this applies also to packages). This improves the readability and overall quality of the packages.

These listed things came to my mind, please extend it if you come up with other things

1. How we tag classes (\code{} or other). Discussed in here https://github.com/microbiome/mia/pull/370
2. Parameters. Some parameters are like `thisIsParameter`, others like `this.is.parameter`. We agreed to use the latter.
3. The error messages of input checks are not unified. If parameters `A` and `B` require a boolean value, the error message should be the same for them (e.g., "'A' must be a boolean value").
4. Similar to point 3. The parameter description should be unified. First give the expected class then explain what the parameter does.

antagomir:
@Daenarys8 is this ready?

antagomir:
Also related to #585

PR #423, opened 2023-08-24
Issue #372:  Unifying parameters

0xMuluh:
Ping #372
@antagomir @TuomasBorman 

Pushing some pending changes, more changes still to be done but would appreciate early suggestions.

### DIFF
--- a/R/makeTreeSummarizedExperimentFromBiom.R
+++ b/R/makeTreeSummarizedExperimentFromBiom.R
@@ -160,13 +160,7 @@ makeTreeSEFromBiom <- function(
     
     # Remove prefixes if specified and rowData includes info
     if(removeTaxaPrefixes && ncol(feature_data) > 0){
-        # Patterns for superkingdom, domain, kingdom, phylum, class, order, family,
-        # genus, species
-        patterns <- "sk__|([dkpcofgs]+)__"
-        feature_data <- lapply(
-            feature_data,
-            gsub, pattern = patterns, replacement = "")
-        feature_data <- as.data.frame(feature_data)
+        feature_data <- .remove_prefixes_from_taxa(feature_data)
     }
     
     # Adjust row and colnames
@@ -196,6 +190,15 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
 }
 
 ################################ HELP FUNCTIONS ################################
+# This function removes prefixes from taxonomy names
+.remove_prefixes_from_taxa <- function(tax_tab, patterns = "sk__|([dkpcofgs]+)__", ...){
+    tax_tab <- lapply(
+        tax_tab,
+        gsub, pattern = patterns, replacement = "")
+    tax_tab <- as.data.frame(tax_tab)
+    return(tax_tab)
+}
+
 # Find taxonomy rank based on prefixes. If found, return
 # corresponding rank. Otherwise, return the original
 # rank that is fed to function.


R/agglomerate.R
```diff
@@ -304,9 +324,9 @@ setMethod("agglomerateByRank", signature = c(x = "TreeSummarizedExperiment"),
 #' @aliases agglomerateByRank
 #' @export
 setMethod("mergeFeaturesByRank", signature = c(x = "TreeSummarizedExperiment"),
-          function(x, ..., agglomerateTree = FALSE){
+          function(x, ..., agglomerate.tree = FALSE){
               .Deprecated(old="agglomerateByRank", new="mergeFeaturesByRank", "Now agglomerateByRank is deprecated. Use mergeFeaturesByRank instead.")
-              x <- agglomerateByRank(x, ..., agglomerateTree = agglomerateTree)
+              x <- agglomerateByRank(x, ..., agglomerate.tree = agglomerate.tree)
               x
```
TuomasBorman: This file is ok

R/calculateJSD.R
```diff
@@ -124,16 +124,18 @@ setMethod("calculateJSD", signature = c(x = "SummarizedExperiment"),
 #' @importFrom DelayedArray getAutoBPPARAM setAutoBPPARAM
 #'
 #' @export
-runJSD <- function(x, BPPARAM = SerialParam(), chunkSize = nrow(x)){
+runJSD <- function(x, BPPARAM = SerialParam(), chunk.size = nrow(x)){
```
TuomasBorman: Add there also the old parameter like ypu have done with agglomerateByRank

R/calculateUnifrac.R
```diff
@@ -234,28 +245,30 @@ runUnifrac <- function(x, tree, weighted = FALSE, normalized = TRUE,
...
     # rows and tree labels
-    if( !(is.null(nodeLab) ||
-        (is.character(nodeLab) && length(nodeLab) == nrow(x) &&
-        all(nodeLab[ !is.na(nodeLab) ] %in% c(tree$tip.label)))) ){
-        stop("'nodeLab' must be NULL or character specifying links between ",
+    if( !(is.null(node.lab) ||
+        (is.character(node.lab) && length(node.lab) == nrow(x) &&
+        all(node.lab[ !is.na(node.lab) ] %in% c(tree$tip.label)))) ){
+        .Deprecated(old="nodeLab", new="node.lab", 
+                    "Now nodeLab is deprecated. Use node.lab parameter instead.")
```
TuomasBorman: Does this also give the deprecation message even though node.lab was actually used? --> with parameters we could just inform user in parameter description and remove these deprecation messages inside code

R/estimateDiversity.R
```diff
@@ -33,12 +33,23 @@
 #'   
 #' @param tree_name a single \code{character} value for specifying which
 #'   rowTree will be used to calculate faith index. 
-#'   (By default: \code{tree_name = "phylo"})
+#'   (By default: \code{tree_name = "phylo"})(Please use \code{tree.name} instead. At some point \code{tree_name}
+#'   will be disabled.)
```
TuomasBorman: You could keep this as simple as possible. "Alias for tree.name. Please use tree.name instead...."

R/mergeSEs.R
```diff
@@ -565,9 +565,9 @@ setMethod("right_join", signature = c(x = "ANY"),
     
     # Create a LinkDataFrame based on the link data
     links <- LinkDataFrame(
-        nodeLab = links[["nodeLab"]],
+        node.lab = links[["node.lab"]],
         nodeNum = links[["nodeNum"]],
-        nodeLab_alias = links[["nodeLab_alias"]],
+        node.lab_alias = links[["node.lab_alias"]],
```
TuomasBorman: Not sure if this is working. links variable is a LinksDataFrame from TreeSE, and it has these columns that are already named

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -229,15 +229,15 @@ setMethod("makePhyloseqFromTreeSummarizedExperiment", signature = c(x = "ANY"),
...
+    if( any(!( rowTree(x, tree.name)$tip) %in% rownames(x)) ){
         # If rowtree do not match, tree is pruned
-        x <- .get_x_with_pruned_tree(x, tree_name)
+        x <- .get_x_with_pruned_tree(x, tree.name)
     }
     # Get rowTree
-    phy_tree <- rowTree(x, tree_name)
+    phy_tree <- rowTree(x, tree.name)
     # Convert rowTree to phyloseq object
     phy_tree <- phyloseq::phy_tree(phy_tree)
```
TuomasBorman: In this kind of internal functions, it is not necessary to change parameter names. We should keep the code as stable as possible and modify it only when necessary.

This makes maintaining harder since there are large changes that do not really affect the functionality. Also it is harder to keep track the functional changes

--> so my suggestion is to keep internal functions and variables as they are, and only  modify those parameters that are visible for user.

TuomasBorman (changes requested):
Looks good!

Check comments, I would only modify parametrs that are visible for the user.

Issue #430, opened 2023-08-27
Failure in makePhyloseqFromTreeSE


antagomir:
I think this simple conversion should work out-of-the box:

```
library(mia)
data(enterotype, package="mia")
makePhyloseqFromTreeSE(enterotype)
```

However it produces the following error:

Error in validObject(.Object) : invalid class “phyloseq” object: 
 Component taxa/OTU names do not match.
 Taxa indices are critical to analysis.
 Try taxa_names()

-> Diagnose and fix?

PR #433, opened 2023-08-30
solving Failure in makePhyloseqFromTreeSE

ake123:
Solved the Failure in makePhyloseqFromTreeSE #430


R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -101,6 +101,7 @@ setMethod("makePhyloseqFromTreeSE",
             # Gets the taxonomic data from rowData, and converts it to tax_table
             tax_table <- as.matrix(rowData(x)[,taxonomyRanks(x)])
```
TuomasBorman: The problem seems to be that only genus level is available. By default drop parameter is TRUE, which means that vector is returned and rownames are lost

-->

as.matrix( rowData(x)[,taxonomyRanks(x), drop = FALSE] )

TuomasBorman (changes requested):

R/makephyloseqFromTreeSummarizedExperiment.R
ake123: Yes that's right and now it works.Thanks! 
Thought it has to do with the tax_table row names

TuomasBorman (approved):

PR #446, opened 2023-09-11
Update biom import examples

### LINKED ISSUE #425: biom import examples
Revise and clarify the biom import examples:
https://microbiome.github.io/mia/reference/makeTreeSEFromBiom.html

1. ignore the SE, make all examples import directly to TreeSE

2. do we need both `loadFromBiom` and `makeTreeSEFromBiom`? or would it be enough keep one of these?

RiboRings:
Hi! I have updated the examples on biom import. Checking for related things to do.


TuomasBorman:
Im wondering if this works also in system that does not have biomformat installed (especially if this works in Bioc system; sometimes they work in GHA but not in Bioc). biomformat is not dependency of mia, it is just dependency of loadFromBiom function

RiboRings:
I see! Then should it work if I put this back?

```
#' if(requireNamespace("biomformat")) {
#'   library(biomformat)
```

TuomasBorman (approved):

TuomasBorman:
> I see! Then should it work if I put this back?
> 
> ```
> #' if(requireNamespace("biomformat")) {
> #'   library(biomformat)
> ```

Well, in practice they could be removed, I think, if biomformat is listed in suggests https://pecanproject.github.io/pecan-documentation/master/declaring-dependencies-depends-suggests-imports.html

PR #447, opened 2023-09-14
Faith bugfix

TuomasBorman:
There were some special cases where Faith index was not working. For example, 1) if no taxa was present in a sample or 2) if taxon was not found from tree tip but from internal node.

I simplified the code. Now if taxa is present, pruning is done for the tree one layer at the time until all tips can be found from the TreeSE rows. This ensures that taxon, that was initially in internal node, is not lost. Faith is calculated directly from pruned tree.


R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -282,6 +282,8 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
             temp <- apply(temp, 1, paste, collapse = "")
             # Now NAs are converted into characters. Convert them back
             temp[ temp == "NA" ] <- NA
+            # Convert also empty strings to NA
+            temp[ temp == "" ] <- NA
```
TuomasBorman, line 286: There might be empty characters -->  convert them to NA

antagomir:
OK! 

However it will be good to make clear in manpage  @details how those cases are treated. If the internal node thing has alternative solutions, it would be good to mention them briefly - for instance just discarding such cases could be an option? (and even give an example on how to do that in case user prefers that).

antagomir (approved):

TuomasBorman:
Added

PR #451, opened 2023-09-25
Fix pseudocount bug for missing values in assay

RiboRings:
Hi!

I fixed this bug from OMA arising when pseudocount is given. It was due to missing values in assay:
```
library(mia)
data("HintikkaXOData", package = "mia")
mae <- HintikkaXOData
transformAssay(mae[[3]], assay.type = "signals",
               MARGIN = "features",
               method = "z", pseudocount = TRUE)

# Error in if (pseudocount && any(mat < 0)) { : 
#   missing value where TRUE/FALSE needed
```

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -551,7 +551,7 @@ setMethod("relAbundanceCounts", signature = c(x = "SummarizedExperiment"),
 .apply_pseudocount <- function(mat, pseudocount){
     if( .is_a_bool(pseudocount) ){
         # If pseudocount TRUE and some negative values, numerical pseudocount needed
-        if ( pseudocount && any(mat<0) ){
+        if ( pseudocount && any(mat < 0, na.rm = TRUE) ){
             stop("The assay contains some negative values. ",
                  "'pseudocount' must be specified manually.", call. = FALSE)
         }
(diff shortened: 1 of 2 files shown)


R/transformCounts.R
```diff
@@ -564,12 +564,12 @@ setMethod("relAbundanceCounts", signature = c(x = "SummarizedExperiment"),
     }
     # Give warning if pseudocount should not be added
     # Case 1: only positive values
-    if( pseudocount != 0 && all(mat>0) ){
+    if( pseudocount != 0 && all(mat > 0) ){
```
TuomasBorman: If missing values caused this, all() should also have na.rm?

R/transformCounts.R
```diff
@@ -551,7 +551,7 @@ setMethod("relAbundanceCounts", signature = c(x = "SummarizedExperiment"),
 .apply_pseudocount <- function(mat, pseudocount){
     if( .is_a_bool(pseudocount) ){
         # If pseudocount TRUE and some negative values, numerical pseudocount needed
```
TuomasBorman: You could test that if there are NA values, I don't know what to do with them so better to disable pseudocount=TRUE if there are NAs

TuomasBorman: or hmmm, maybe it is not necessary since after adding pseudocount the NAs are still NAs

TuomasBorman: But calculate pseudocount with ifelse(pseudocount, min(mat[mat>0], na.rm = TRUE), 0)

R/transformCounts.R
TuomasBorman: --> it is not necessary in this case since if there are NAs the result is just FALSE (not all are TRUEs)

R/transformCounts.R
TuomasBorman: Or hmm, I think the best choice is to disable this when NAs...

TuomasBorman (changes requested):
This PR is ok, just disable pseudocount=TRUE when there are NA values (pseudocount must be specfied manually...)

R/transformCounts.R
RiboRings: Just to make sure I added na.rm = TRUE, because:
```
> all(c(1, NA, 0) > 3)
[1] FALSE
> all(c(1, NA, 0) < 3)
[1] NA
> all(c(1, NA, 0) < 3, na.rm = TRUE)
[1] TRUE
```

TuomasBorman (approved):

PR #454, opened 2023-09-28
loadFromBiom fix

TuomasBorman:


TuomasBorman:
Error occurs when there is only one taxa present --> the df is transposed

antagomir:
> Error occurs when there is only one taxa present --> the df is transposed

One can use "drop":

```
L3 <- LETTERS[1:3]
char <- sample(L3, 10, replace = TRUE)
d <- data.frame(x = 1, y = 1:10, char = char)
```

`d[,1]`
 [1] 1 1 1 1 1 1 1 1 1 1

`d[,1, drop=FALSE]`
   x
1  1
2  1
3  1
4  1
5  1
6  1
7  1
8  1
9  1
10 1

TuomasBorman:
> > Error occurs when there is only one taxa present --> the df is transposed
> 
> One can use "drop":
> 
> ```
> L3 <- LETTERS[1:3]
> char <- sample(L3, 10, replace = TRUE)
> d <- data.frame(x = 1, y = 1:10, char = char)
> ```
> 
> `d[,1]` [1] 1 1 1 1 1 1 1 1 1 1
> 
> `d[,1, drop=FALSE]` x 1 1 2 1 3 1 4 1 5 1 6 1 7 1 8 1 9 1 10 1

Thanks, I made changes and now the solution is better. (However, the problem was in apply-apply part, not in subsetting)

PR #460, opened 2023-10-13
estimateAlpha

### LINKED ISSUE #417: Rarefaction as optional strategy for alpha diversity
Ref: Pat Schloss' [recent analyses](https://www.biorxiv.org/content/10.1101/2023.06.23.546313v1.full.pdf) on rarefaction in alpha & beta diversity. This addressess well some of the long debates and controversies surrounding rarefaction in microbial ecology.

Based on this I tend to think that rarefaction should be provided as an option for alpha diversity calculations. 

Here rarefaction refers to averaging alpha diversity estimates across many random rarifications (100x and 1000x are used in the preprint). This makes most difference in less diverse and low abundance environments but may influence also more rich ecosystems (see the preprint).

Options:
- write an apply/do loop around estimateDiversity and just provide that as an example in mia vignette / OMA
- provide rarefaction as an option in estimateDiversity directly (including user defined arguments for the number of rarification rounds, and for the read count threshold with some automated default such as the smallest read count among samples)

I am in favour of the second option.

Note:
- we have estimateDiversity, estimateRichness, estimateDominance etc.; these all should be updated
- or we can consider combining them into a single estimateAlpha function to simplify procedures and maintenance (perhaps a separate issue to deal with first to minimze extra work)

ChouaibB:
Initial draft implementation.


antagomir:
Related to #417

R/estimateAlphaWithRarefaction.R
```diff
@@ -0,0 +1,84 @@
...
+#' 
+#' @param seed a single \code{integer} value that creates the seeds for the 
+#' nround rarefaction.
+#' 
+#' @param args.sub argument list passed to \code{\link[mia:subsampleCounts]{subsampleCounts}}
+#' 
+#' @param FUN the alpha diversity function to be used; e.g.
+#'  \code{\link[mia:estimateDiversity]{estimateDiversity}},
+#'  \code{\link[mia:estimateEvenness]{estimateEvenness}}, 
+#'  \code{\link[mia:estimateRichness]{estimateRichness}}.
```
antagomir: Also include estimateDominance?

R/estimateAlphaWithRarefaction.R
```diff
@@ -0,0 +1,84 @@
...
+                                         args.sub=list(assay.type="counts",
+                                                       min_size=min(colSums(assay(x, "counts")), na.rm = TRUE),
+                                                       verbose=FALSE),
+                                         FUN=mia::estimateDiversity,
+                                         args.fun=list(index="shannon",
+                                                       assay.type="subsampled"),
+                                         name = args.fun$index){
+    # checks
+    if(!.is_an_integer(nrounds)) {
+        stop("'nrounds' must be an interger.",
```
antagomir: interger -> integer..

R/estimateAlphaWithRarefaction.R
```diff
@@ -0,0 +1,84 @@
...
+    }
+    if(!.is_non_empty_string(name)) {
+        stop("'name' should be a non empty string.",
+             call. = FALSE)
+    }
+    
+    # Generating seeds for every round
+    set.seed(seed)
+    SEEDS <- sample.int(10000, size = nrounds)
+    colData(x)[, name] <- lapply(seq(nrounds), function(i){
```
antagomir: IMO it is entirely sufficient to set one seed before the entire lapply; setting step-specific seed is an overkill

tests/testthat/test-10estimateAlphaWithRarefaction.R
```diff
@@ -0,0 +1,12 @@
...
+    tse <- GlobalPatterns
+    # Calculate the default Shannon index with 1 rarefaction round
+    tse <- estimateAlphaWithRarefaction(tse)
+    expect_true(any(grepl("shannon", colnames(colData(tse)))))
+
+    # Calculate the default observed richness with 10 rarefaction rounds
+    tse <- estimateAlphaWithRarefaction(tse, nrounds=10,
+     FUN=mia::estimateRichness, args.fun=list(index="observed"))
+    expect_true(any(grepl("observed", colnames(colData(tse)))))
+})
```
antagomir: If possible, I would run the same tests than are run for estimateDiversity (and other diversity functions) but with and without the rarefaction argument.

antagomir:
I am tending to think that we should combine all these estimateXXX functions into a single estimateAlpha function as your PR suggest. The other estimateXXX functions could be turned into internal functions to simplify the interface. That estimateAlpha function can then have rarefaction as optional argument (including the user-specified rarefaction depth; by default the lowest read count among the samples).

Things to consider:
1) previously exported estimateXXX functions would need deprecation message when they are moved into internal functions
2) the diversity index names should have suffixes, like "shannon_diversity", "observed_richness" etc.
3) there is a mechanism available in R that can pick those argument names even when just the (unique) start of the index name is given, like "shannon" would be recognized to mean "shannon_diversity" so the users can use shorter names if they need to but the output need to use the full names (for instance Simpson index is available both for diversity and evenness, therefore it must be indicated which one it is).

If unclear, we can discuss more.

I have some further comments in mind after the next PR draft on this.

antagomir:
The function has to be changed into estimateAlpha, which then includes rarefaction as an optional argument

antagomir (approved):

antagomir:
An idea: the method name (like "shannon") could also say "diversity", then the function would return all "diversity" indices and if one asks for "dominance" then those etc (richness / evenness / diversity / dominance / rarity are the options and the indices are listed in those corresponding functions now). If no method is specific, then all indices would be returned by default.

TuomasBorman:
I think this is not the best way to implement this. Better way is to keep those estimateDiversity.R and other files and just add estimateAlpha.R with estimateAlpha functionality (remove documentation from necessary places)

The problem with moving all the code to one file is that it is a nightmare for someone that tries to check what has been changed. It looks like everything has

antagomir:
I agree. The old and extensively tested functions can be kept as is. They should be just converted into internal functions that are not exported (as I already explictily suggested in the opening statement of this issue), and then called from a wrapper called `estimateAlpha()`. This should be also faster to implement than rewriting the entire function.

ChouaibB:
Alright, I misunderstood the comment I will make the changes.

antagomir:
Always aim at minimal and straightfwd changes unless otherwise agreed.

If you could update this PR accordingly that would be great!

ChouaibB:
> An idea: the method name (like "shannon") could also say "diversity", then the function would return all "diversity" indices and if one asks for "dominance" then those etc (richness / evenness / diversity / dominance / rarity are the options and the indices are listed in those corresponding functions now). If no method is specific, then all indices would be returned by default.

I didn't understand, could you please explain more.

ChouaibB:
> > An idea: the method name (like "shannon") could also say "diversity", then the function would return all "diversity" indices and if one asks for "dominance" then those etc (richness / evenness / diversity / dominance / rarity are the options and the indices are listed in those corresponding functions now). If no method is specific, then all indices would be returned by default.
> 
> I didn't understand, could you please explain more.

Do you mean smt like this?

```
.get_indices <- function(measure) {
    switch(measure,
           "diversity" = c("coverage_diversity", "coverage",
                           "faith_diversity", "faith",
                           "fisher_diversity", "fisher",
                           "gini_simpson_diversity", "gini_simpson",
                           "inverse_simpson_diversity", "inverse_simpson",
                           "log_modulo_skewness_diversity", "log_modulo_skewness",
                           "shannon_diversity", "shannon"),
           "dominance" = c("absolute_dominance", "absolute",
                           "dbp_dominance", "dbp",
                           "core_abundance_dominance", "core_abundance",
                           "gini_dominance", "gini", 
                           "dmn_dominance", "dmn",
                           "relative_dominance", "relative",
                           "simpson_lambda_dominance", "simpson_lambda"),
           "evenness" = c("camargo_evenness", "camargo",
                          "pielou_evenness", "pielou",
                          "simpson_evenness",
                          "evar_evenness", "evar",
                          "bulla_evenness", "bulla"),
           "richness" = c("ace_richness", "ace",
                          "chao1_richness", "chao1",
                          "hill_richness", "hill",
                          "observed_richness", "observed"))
}
```

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+#' 
+#' @param x a \code{\link{SummarizedExperiment}} object.
+#' 
+#' @param assay.type the name of the assay used for
+#'   calculation of the sample-wise estimates.
+#'   
+#' @param assay_name a single \code{character} value for specifying which
+#'   assay to use for calculation.
+#'   (Please use \code{assay.type} instead. At some point \code{assay_name}
+#'   will be disabled.)
```
antagomir: This is an alias for assay.type - how about just mentioning that and having all explanations in assay.type?

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+#' # Shows the estimated observed richness
+#' colData(tse)$observed_richness
+#' 
+#' @importFrom dplyr %>% 
+#' 
+#' @rdname estimateAlpha
+#' @export
+estimateAlpha <- function(x, assay.type = "counts", assay_name = NULL,
+                          index = c("coverage_diversity", "fisher_diversity",
+                                    "faith_diversity", "faith",
```
antagomir: Why there is now both "faith_diversity" and "faith"?

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+                           "simpson_lambda_dominance", "simpson_lambda"),
+           "evenness" = c("camargo_evenness", "camargo",
+                          "pielou_evenness", "pielou",
+                          "simpson_evenness",
+                          "evar_evenness", "evar",
+                          "bulla_evenness", "bulla"),
+           "richness" = c("ace_richness", "ace",
+                          "chao1_richness", "chao1",
+                          "hill_richness", "hill",
+                          "observed_richness", "observed"))
```
antagomir: imo it would be better to have just one version for each name, the one with the _ending?

the rest is done with parsing  if necessary

i would leave the shorter names out from this entirely. I think there is a partial matching mechanism in R to allow users to use shorter argument names in the function calls, as long as the string is a unique match. You can try that out. If it doesnt work, we can add it later.  For now, I would use unique argument names within the function.

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+}
+
+.alpha_rarefaction <- function(x,
+                               nrounds=1L,
+                               seed=123,
+                               args.sub=list(assay.type="counts",
+                                             min_size=min(colSums(assay(x, "counts")),
+                                                          na.rm = TRUE),
+                                             verbose=FALSE),
+                               FUN=estimateDiversity,
```
antagomir: should this rather be estimateAlpha?

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+                               args.fun=c(index="shannon",
+                                             assay.type="subsampled"),
+                               ...,
+                               name = args.fun$index) {
+    set.seed(seed)
+    colData(x)[, name] <- lapply(seq(nrounds), function(i){
+        x_sub <- do.call(subsampleCounts, args = c(list(x), args.sub))
+        suppressWarnings(x_sub <- do.call(FUN, args = c(list(x_sub),
```
antagomir: add code comment what this is expected to do

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+                                                        list(...))))
+        colData(x_sub)[, args.fun$index, drop=FALSE]
+    }) %>% as.data.frame() %>% rowMeans() %>% as.data.frame()
+    return(x)
+}
+
+.parse_name <- function(index, name, measure) {
+    # don't change name if defined by user
+    if (name==index) {
+        if (measure %in% unlist(strsplit(index, "\\_"))) {
```
antagomir: something like this might be faster 

vapply(strsplit(word,"-"), `[`, 1, FUN.VALUE=character(1))

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+            name = index
+        } else {
+            name = paste0(index, "_", measure)
+        }
+    } else {
+        return(name)
+    }
+}
+
+.estimate_diversity <- function(x, assay.type = "counts", assay_name = NULL,
```
antagomir: in internal functions i think there is no need to use "assay_name"

it could be just renamed into assay.type as one of the first things in the exported estimateAlpha function. Then all internals can just use assay.type. This will simplify everything.

antagomir: same comment for all internals

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+    estimateDominance(x, assay.type=assay.type, assay_name=assay_name,
+                      index=index, ntaxa=ntaxa, aggregate=aggregate,
+                      name=name, ...)
+}
+
+.estimate_evenness <- function(x, assay.type = assay_name, assay_name = "counts",
+                               index = c("camargo", "pielou", "simpson_evenness",
+                                         "evar", "bulla"),
+                               name = index, ...) {
+    estimateEvenness(x, assay.type = assay.type, assay_name = assay_name,
```
antagomir: I think we could also deprecated estimateRichness, estimateDiversity, estimateEvenness, estimateDominance now on the same go. They should still remain working but give a deprecation warning that suggests that users use estimateAlpha

tests/testthat/test-10estimateAlpha.R
```diff
@@ -0,0 +1,51 @@
+test_that("Estimate Alpha Diversity Indices with Rarefaction", {
```
antagomir: make sure the tests work with both "shannon" and "shannon_diversity". Not necessary to add to all tests but one 

i think some index name is the same for richness/diversity/evenness although with a different meaning. If there were such cases, add test that it gives different result whether it is indexname_evenness or indexname_diversity

tests/testthat/test-10estimateAlpha.R
```diff
@@ -0,0 +1,51 @@
...
+    data(GlobalPatterns, package="mia")
+    tse <- GlobalPatterns
+    ## Testing diversity
+    # Calculate the default Shannon index with no rarefaction
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "shannon")
+    expect_true(any(grepl("shannon_diversity", colnames(colData(tse)))))
+    # Calculate same index with 10 rarefaction rounds
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "shannon",
+                         rarify=TRUE, nrounds=10, name="shannon_10")
```
antagomir: I would use "n.iter" instead of "nrounds" in the argument name

antagomir: it would be good to add some subsampling depth also here in the test

tests/testthat/test-10estimateAlpha.R
```diff
@@ -0,0 +1,51 @@
...
+    ## Testing Dominance
+    # Calculate the default gini_dominance index with no rarefaction
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "gini_dominance")
+    expect_true(any(grepl("gini_dominance", colnames(colData(tse)))))
+    # Calculate same index with 10 rarefaction rounds
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "gini_dominance",
+                         rarify=TRUE, nrounds=10, name="gini_dominance_10")
+    expect_true(any(grepl("gini_dominance_10", colnames(colData(tse)))))
+    # comparing the estimates
+    expect_false(any(colData(tse)$gini_dominance==colData(tse)$gini_dominance_10))
```
antagomir: for simplicity, it would be good to write all these cases everywhere like tse$gini_dominance instead of the longer colData(tse)$gini_dominance ?

tests/testthat/test-10estimateAlpha.R
```diff
@@ -0,0 +1,51 @@
...
+    
+    ## Testing Evenness
+    # Calculate the default pielou index with no rarefaction
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "pielou")
+    expect_true(any(grepl("pielou_evenness", colnames(colData(tse)))))
+    # Calculate same index with 10 rarefaction rounds
+    tse <- estimateAlpha(tse, assay.type = "counts", index = "pielou",
+                         rarify=TRUE, nrounds=10, name="pielou_10")
```
antagomir: Also add some test that two different rarifications with 10 iterations would give almost the same result?

antagomir (commented):
Nice! Some comments added. Perhaps TB has more

R/estimateAlphaWithRarefaction.R
TuomasBorman: Regarding setting seed. We have this kind of solution in couple of functions, but this is not recommended way according to Bioconductor. Seed should be set outside the function (Both solutions are fine by me)

R/estimateAlpha.R
TuomasBorman: Do we even need that alias since we are removing it and this is a new function?

R/estimateAlphaWithRarefaction.R
antagomir: I also think it is best to set the seed outside of the function. Users still have all the control over this, and it simplifies the function.

R/estimateAlpha.R
antagomir: True. Alias is not needed for estimateAlpha. 

For other (earlier) diversity estimation functions maybe.

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+                                    "simpson_evenness", "evar_evenness",
+                                    "bulla_evenness",
+                                    "ace_richness", "chao1_richness", "hill_richness",
+                                    "observed_richness"),
+                          name = index,
+                          ...,
+                          rarify=FALSE,
+                          seed = 123,
+                          nrounds=10,
+                          rarefaction_depth=min(colSums(assay(x, "counts")), na.rm = TRUE)){
```
TuomasBorman: rarefaction_depth could be NULL by default which then is calculated inside the function. --> that way user can run this with default settings by only specifying the assay.type

OR (maybe better if works) can you replace "counts" with assay.type? Does it work?

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+    if(!.is_an_integer(nrounds)) {
+        stop("'nrounds' must be an integer.",
+             call. = FALSE)
+    }
+    if(!(is.double(rarefaction_depth) & rarefaction_depth > 0)) {
+        stop("'rarefaction_depth' must be a non-zero positive double.",
+             call. = FALSE)
+    }
+    FUN <- NULL
+    if(index %in% .get_indices("diversity")) {
```
TuomasBorman: MIght be good to have a comment what this block does

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+    } else {
+        stop("'index' is coresponding to none of the alpha diversity measures.",
+             call. = FALSE)
+    }
+    
+    if (rarify) {
+        .alpha_rarefaction(x, nrounds = nrounds, seed = seed,
+                           args.sub = list(assay.type=assay.type,
+                                           min_size=rarefaction_depth,
+                                           verbose=FALSE),
```
TuomasBorman: RUn BiocCheck::BiocCheck(), its seems that this indentation is not correct

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+                                                        args.fun,
+                                                        list(...))))
+        colData(x_sub)[, args.fun$index, drop=FALSE]
+    }) %>% as.data.frame() %>% rowMeans() %>% as.data.frame()
+    return(x)
+}
+
+.parse_name <- function(index, name, measure) {
+    # don't change name if defined by user
+    if (name==index) {
```
TuomasBorman: index can be a vector of indices? Does this then work?

R/estimateAlpha.R
```diff
@@ -0,0 +1,248 @@
...
+        if (measure %in% unlist(strsplit(index, "\\_"))) {
+            name = index
+        } else {
+            name = paste0(index, "_", measure)
+        }
+    } else {
+        return(name)
+    }
+}
+
```
TuomasBorman: Might be good to have comment what this does

R/estimateAlpha.R
TuomasBorman: Why warnings are suppressed? Add comment there. These should be avoided

TuomasBorman (changes requested):
COding style should more or less match between functions even though we have multiple developers. For example there is no many comments

R/estimateAlpha.R
ChouaibB: I thought that's what you requested, user can use both to refer to the alpha measure to be used?

R/estimateAlpha.R
ChouaibB: not actually I should change it to .estimate_diversity

R/estimateAlpha.R
ChouaibB: not working

R/estimateAlpha.R
ChouaibB: Yes, deprication message was added to those functions at their own R files.

tests/testthat/test-10estimateAlpha.R
ChouaibB: ok.
I checked all indices of each alpha measure before implementing, and it turns out actually all are unique to all functions.

tests/testthat/test-10estimateAlpha.R
ChouaibB: ok, i will add one with 10 and another with 20.

R/estimateAlpha.R
ChouaibB: I will check that out

R/estimateAlpha.R
ChouaibB: Actually good point. I have made so far in such a way that it handles one index. I will upgrade it to handle multiple ones.

R/estimateAlpha.R
ChouaibB: Due to the deprication message added to the alpha functions, it needed to be supressed when called from estimateAlpha to pass the test.

R/estimateAlpha.R
ChouaibB: it's removed

antagomir: I think this is automatically done in R functions when the shorter name matches the full name.

tests/testthat/test-10estimateAlpha.R
antagomir: and initalized with different random seeds

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
+#' Estimate alpha indices using rarefaction
```
antagomir: I think "using rarefaction" can be removed. It is one option among others. The main thing is estimation of alpha indices.

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' 
+#' The function estimates alpha diversity measures optionally using n rounds of rarefaction,
```
antagomir: measures -> indices

antagomir: remove "n rounds of" it is a technical detail

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' given the rarefaction depth, then stores results at \code{\link{colData}}.
```
antagomir: remove "given the rarefaction depth"

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' 
+#' @param x a \code{\link{SummarizedExperiment}} object.
+#' 
+#' @param assay.type the name of the assay used for
+#'   calculation of the sample-wise estimates.
+#'   
+#' @param index a \code{character} vector, specifying the alpha diversity measures
```
antagomir: measures -> indices

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#'   to be calculated
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
+#'   indices specifying the alpha diversity measures used.
```
antagomir: remove "specifying the alpha diversity measures used"

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#'   
+#' @param ... optional arguments.
+#'
```
antagomir: remove extra empty line

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#'   to be calculated
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
+#'   indices specifying the alpha diversity measures used.
+#'   
+#' @param ... optional arguments.
+#' 
+#' 
+#' @param rarify logical scalar: Should the alpha diversity measures be estimated
```
antagomir: measures -> indices (everywhere)

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#'   using rarefaction? (default: \code{FALSE})
+#' 
+#' @param seed a single \code{integer} value as the seed used for the nround
+#'  rarefaction.
```
antagomir: remove the seeding from the function as discussed elsewhere today

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' 
+#' @param n.iter a single \code{integer} value for the number of rarefaction
+#' rounds.
+#' 
+#' @param rarefaction_depth a \code{double} value as for the minimim size or
```
antagomir: new argument names should use point notation, this would be "rarefaction.depth"

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' @examples
+#' 
+#' data("GlobalPatterns")
+#' tse <- GlobalPatterns
+#' 
+#' # Calculate the default Shannon index with no rarefaction
+#' tse <- estimateAlpha(tse, assay.type = "counts", index = "shannon")
+#' 
+#' # Shows the estimated Shannon index
+#' colData(tse)$shannon_diversity
```
antagomir: or just tse$shannon_diversity

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#'
+#'# Calculate observed richness with 10 rarefaction rounds
+#' tse <- estimateAlpha(tse, assay.type = "counts", index = "observed_richness",
+#' rarify=TRUE, n.iter=10)
+#' 
+#' # Shows the estimated observed richness
+#' colData(tse)$observed_richness
```
antagomir: same

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+#' 
+#' # Calculate the default Shannon index with no rarefaction
+#' tse <- estimateAlpha(tse, assay.type = "counts", index = "shannon")
+#' 
+#' # Shows the estimated Shannon index
+#' colData(tse)$shannon_diversity
+#'
+#'# Calculate observed richness with 10 rarefaction rounds
+#' tse <- estimateAlpha(tse, assay.type = "counts", index = "observed_richness",
+#' rarify=TRUE, n.iter=10)
```
antagomir: Do we need this argument "rarify"?

Another option would be to just use rarefation depth.

That can be set to Infinity by default (or to max read count across samples). 

If rarefaction depth is higher than the read count in the sample with the highest read count, then rarefaction should be skipped. 

This way we can reduce the number of arguments needed. Just activate rarefaction is the depth is lower than max read count? If user does not specify anything and goes with defaults, then there is no rarefaction (and also no n.iter rounds).

R/estimateAlpha.R
antagomir: same

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+                                    "simpson_lambda_dominance",
+                                    "camargo_evenness", "pielou_evenness",
+                                    "simpson_evenness", "evar_evenness",
+                                    "bulla_evenness",
+                                    "ace_richness", "chao1_richness", "hill_richness",
+                                    "observed_richness"),
+                          name = index,
+                          ...,
+                          rarify=FALSE,
+                          seed = 123,
```
antagomir: remove seed from the function

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+                          n.iter=10,
+                          rarefaction_depth=min(colSums(assay(x, assay.type)), na.rm = TRUE)){
+    # checks
```
antagomir: more explanations needed; add a comment line for each substantial step below to help reader understand what it does. Check other functions for examples.

R/estimateAlpha.R
```diff
@@ -0,0 +1,247 @@
...
+                                                             name=name[i]),
+                                                        list(...))))
+        }
+    }
+    return(x)
+}
+
+## Helper functions
+
+.get_indices <- function(measure) {
```
antagomir: nice!

R/estimateAlpha.R
antagomir: ok

R/estimateDiversity.R
```diff
@@ -183,7 +183,9 @@
 #' "Faith",  "LogModSkewness")
```
antagomir, line 188: does this function estimateDiversity now have a deprecation message?

antagomir (commented):
JUst some rather minor comments.

antagomir:
After this PR is accepted, we need to remember update OMA.

R/estimateAlpha.R
ChouaibB: I thought the result of the discussion was that the seed is ok to keep outside? subsampling will use internally other seeds, and having a seed outside the call of subsampling will keep results reproducible?

R/estimateDiversity.R
ChouaibB: yes

R/estimateAlpha.R
antagomir: Yes. Outside of the function entirely.

antagomir: It will keep results reproducible if you set a seed, then call a function.

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+            if (index[i]!="simpson_evenness") {
+                index[i] <- gsub("_evenness", "", index[i])
+            }
+            FUN <- .estimate_evenness
+        } else if (any(grepl(index[i], .get_indices("richness")))) {
+            name[i] <- .parse_name(index[i], name[i], "richness")
+            index[i] <- gsub("_richness", "", index[i])
+            FUN <- .estimate_richness
+        } else {
+            stop("'index' is coresponding to none of the alpha diversity indices.",
```
antagomir: Instead of saying what it is not, say what it should be.

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+            x <- .alpha_rarefaction(x, n.iter = n.iter, seed = seed,
+                                    args.sub = list(assay.type=assay.type,
+                                                    min_size=rarefaction.depth,
+                                                    verbose=FALSE),
+                                    FUN=FUN,
+                                    args.fun=list(index=index[i], assay.type="subsampled"),
+                                    ...,
+                                    name=name[i])
+        } else {
+            suppressWarnings(x <- do.call(FUN, args = c(list(x, assay.type=assay.type,
```
antagomir: Add comment on what this is supposed to do

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+                               args.sub=list(assay.type="counts",
+                                             min_size=min(colSums(assay(x, "counts")),
+                                                          na.rm = TRUE),
+                                             verbose=FALSE),
+                               FUN=.estimate_diversity,
+                               args.fun=c(index="shannon",
+                                          assay.type="subsampled"),
+                               ...,
+                               name = args.fun$index) {
+    set.seed(seed)
```
antagomir: seed still included despite the discussion otherwise?

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+}
+
+.parse_name <- function(index, name, measure) {
+    # parsing name string to use as a column name at colData when storing estimates
+    if (name==index) {
+        # check if suffix of the alpha indices if present at index
+        # otherwise keeping suffix as a name if name not defined by user.
+        if (measure %in% unlist(strsplit(index, "\\_"))) {
+            name = index
+            } else {
```
antagomir: are the indentations ok here? 

also name=index should be name <- index here i think

R/estimateAlpha.R
ChouaibB: Once again, the earlier discussion resulted in keeping the seed for reproducibility? the subsampling function does it internally for each iteration, so keeping a seed outside the function call will enable the user to reproduce results....?

antagomir: Seed can be given outside the function and the results remain reproducible. 

Bioconductor recommends not having an additional seed in the functions.

I do not see why we need the seed inside function for reproducibility. You can test this yourself: apply the function two times without setting seed outside -> different results; set seed outside and immediately run the function -> same results every time

ChouaibB: ok I checked the subsampling function, I see now that it would produce the same seed over and over. But then results are not reproducible for the whole process due to [link](https://github.com/microbiome/mia/blob/master/R/subsampleCounts.R#L87C28-L87C61) ... but guess that's the problem of the subsampling function. I remove it from here.

antagomir: Yes, let us set seed outside of function if necessary, and remove from the function itself. Then the function result will remain reproducible.

This should always generate the same:

```
set.seed(24452)
estimateAlpha(...)
```

If this is not the case, prove me wrong.

ChouaibB: I already said setting outside the function is reproducible. "Outside the function" was not clear at comments for me....
but setting one seed before the lapply is not going to make the rarefaction happen since same subsampling results will be produced in every iteration.....

ChouaibB: That's why the communication was confusing after asking at the beginning to set one seed before the lapply...

antagomir: I do not think that statment is true: if you set seed outside the function, and then run lapply within the function, the lapply will not produce the same subsampling results on every iteration. If it does, that is unexpected and I would like to get a reproducible example that demonstrates that.

ChouaibB: Or perhaps I have cognitive problems too, which does not make me a good candidate for being a developer... IDK it's possible...

antagomir: You can try it out: run lapply() in those alternative ways:

# One general seed before lapply
1) 
set.seed(123); res1a <- lapply(....)
set.seed(123); res1b <- lapply(....)
-> res1a and res1b will yield identical outcomes -> result is reproducible

# Same seed initialized again at each iteration
2) res2 <- lapply(...., function (x) {set.seed(123); ... })
-> this will yield different outcome than res1a or res1b because here the seed is reset to 123 at each new iteration instead of just continuing the seedings from where the previous iteration left

# Random seed per each iteration
3) res3 <- lapply(...., function (x) {set.seed(sample(1e6, 1)); ... })
-> this yields different outcome from res1a, res1b, res2 since it uses different (randomly generated) seed at every iteration

ChouaibB: > I do not think that statment is true: if you set seed outside the function, and then run lapply within the function, the lapply will not produce the same subsampling results on every iteration. If it does, that is unexpected and I would like to get a reproducible example that demonstrates that.

Yes it won't, it will produced only the first iteration the same and the rest is not reproducible, and the question I had in mind is how would you make the whole iterations reproducible to have final reproducible results. That's why I still do think that generating a list of seeds and passing on each iteration the corresonding seed to subsampleCount is simple, safe and definitely reproducible...

ChouaibB: ok, I see your point now, I didn't know that setting one seed before lapply will make reproducible result:
image

antagomir: Yes, this is what I said: the result for the entire loop is reproducible when the seed it set before applying the loop.

antagomir: Is this now clear?

ChouaibB: Yes it is, I tested also outside the function and gives same results.
But what's honestly not clear is why you asked me to put it before the lapply at the begining then re-asked to remove it, and Tuomas also was ok with both solutions and you didn't comment? :) I have noticed this multiple times and it's the same story in every PR I make... I am not really motivated for this especially when I have other tasks at hand waiting for me to be done...  It's ok I understand I am not good enough of a developer, so perhaps I should do smt else instead? or go to coding 101. Maybe I could just open issues when encounter some and come here when I am good enough someday?

antagomir: The discussion evolved. 

Initially I thought that before lapply is enough. Then Tuomas pointed out that Bioc does not recommend including seed in the function. It sounded reasonable, so I changed my thinking and concluded that we can remove the seed from the function entirely. I might have not communicated that clearly in a hurry as I mistakenly thought I stated it in the later comments that the seed should be placed outside of the function.

The other points we should discuss elsewhere than in the public Github issues, thanks.

antagomir: My apologies, if I have not expressed the task with sufficient clarity, will try to improve on that now.

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
+#' Estimate alpha indices.
```
TuomasBorman: alpha diversity indices

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+                                    "absolute_dominance", "dbp_dominance",
+                                    "core_abundance_dominance", "gini_dominance",
+                                    "dmn_dominance", "relative_dominance",
+                                    "simpson_lambda_dominance",
+                                    "camargo_evenness", "pielou_evenness",
+                                    "simpson_evenness", "evar_evenness",
+                                    "bulla_evenness",
+                                    "ace_richness", "chao1_richness", "hill_richness",
+                                    "observed_richness"),
+                          name = index,
```
TuomasBorman: I think this function should be implemented in generic/method style like the other functions

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+    # checks
+    if(is.null(index) & any(!sapply(index, .is_non_empty_string))) {
+        stop("'index' should be a character vector.",
+             call. = FALSE)
+    }
+    if(!.is_an_integer(n.iter)) {
+        stop("'n.iter' must be an integer.",
+             call. = FALSE)
+    }
+    if(!(is.double(rarefaction.depth) & rarefaction.depth > 0)) {
```
TuomasBorman: is.double --> is.numeric

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+                                    "simpson_evenness", "evar_evenness",
+                                    "bulla_evenness",
+                                    "ace_richness", "chao1_richness", "hill_richness",
+                                    "observed_richness"),
+                          name = index,
+                          ...,
+                          n.iter=10,
+                          rarefaction.depth=max(colSums(assay(x, assay.type)), na.rm = TRUE)){
+    # checks
+    if(is.null(index) & any(!sapply(index, .is_non_empty_string))) {
```
TuomasBorman: You can use argument matching here. I think we have used in other functions. match.arg or something

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+            x <- .alpha_rarefaction(x, n.iter = n.iter,
+                                    args.sub = list(assay.type=assay.type,
+                                                    min_size=rarefaction.depth,
+                                                    verbose=FALSE),
+                                    FUN=FUN,
+                                    args.fun=list(index=index[i], assay.type="subsampled"),
+                                    ...,
+                                    name=name[i])
+        } else {
+            # Estimate index without rarefaction
```
TuomasBorman: Why there is a warning? Add comment

TuomasBorman: suppressing warnings should be avoided. Probably there is a good reason for that, but it might be good to add info why it is done

R/estimateAlpha.R
```diff
@@ -0,0 +1,245 @@
...
+                           "gini_dominance", "dmn_dominance", "relative_dominance",
+                           "simpson_lambda_dominance"),
+           "evenness" = c("camargo_evenness", "pielou_evenness", "simpson_evenness",
+                          "evar_evenness", "bulla_evenness"),
+           "richness" = c("ace_richness", "chao1_richness", "hill_richness", 
+                          "observed_richness"))
+}
+
+.alpha_rarefaction <- function(x,
+                               n.iter=1L,
```
TuomasBorman: As a note, you could add line break after `function(` --> the code would look nicer

vignettes/mia.Rmd
```diff
@@ -226,7 +226,7 @@ and results are stored automatically in `colData`. Selected indices can be
 calculated individually by setting `index = "shannon"` for example. 
 
 ```{r}
-tse <- estimateDiversity(tse)
+suppressWarnings(tse <- estimateDiversity(tse))
```
TuomasBorman: You could use this new function

TuomasBorman (changes requested):
Looks good! I will check this in more detail (try the function) when I get my computer working

R/estimateAlpha.R
ChouaibB: That is handy but then the user cannot call also with the short names of indices as requested.

TuomasBorman: ahh true, sorry for hassle

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+#'   
+#' @param index a \code{character} vector, specifying the alpha diversity indices
+#'   to be calculated
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
+#'   indices.
+#'   
+#' @param ... optional arguments.
+#'
```
TuomasBorman: Add default values

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+                   n.iter=10,
+                   rarefaction.depth=max(colSums(assay(x, assay.type)), na.rm = TRUE)){
+              # checks
+              if(is.null(index) & any(!sapply(index, .is_non_empty_string))) {
+                  stop("'index' should be a character vector.", call. = FALSE)
+                  }
+              if(!.is_an_integer(n.iter)) {
+                  stop("'n.iter' must be an integer.", call. = FALSE)
+                  }
+              if(!(is.numeric(rarefaction.depth) & rarefaction.depth > 0)) {
```
TuomasBorman: is.numeric(rarefaction.depth) && rarefaction.depth > 0) --> use two &

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+                  faith_diversity, gini_simpson_diversity,
+                  inverse_simpson_diversity, log_modulo_skewness_diversity,
+                  shannon_diversity, absolute_dominance, dbp_dominance,
+                  core_abundance_dominance, gini_dominance,
+                  dmn_dominance, relative_dominance, simpson_lambda_dominance,
+                  camargo_evenness, pielou_evenness, simpson_evenness,
+                  evar_evenness, bulla_evenness, ace_richness, chao1_richness,
+                  hill_richness or observed_richness.",
+                       call. = FALSE)
+                  }
```
TuomasBorman: Check already in input check that every index can be found. There you can also parse the name. --> fail fast

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+        x_sub <- do.call(subsampleCounts, args = c(list(x), args.sub))
+        # calculating the diversity indices on the subsampled object
+        # warnings are supressed due to the depricated warning of the alpha
+        # measure functions
+        suppressWarnings(x_sub <- do.call(FUN, args = c(list(x_sub),
+                                                        args.fun,
+                                                        list(...))))
+        # Storing estimate results
+        colData(x_sub)[, args.fun$index, drop=FALSE]
+    }) %>% data.frame(.) %>% rowMeans(.) %>% data.frame(.)
```
TuomasBorman: You could split these to multiple steps. I mean first you could calculate the value to variable a, then do these modifications to it and then return

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+                  "log_modulo_skewness","shannon"),
+        name = index, ...) {
+    estimateDiversity(x, assay.type=assay.type, index=index, name=name, ...)
+}
+
+.estimate_dominance <- function(
+        x,
+        assay.type = "counts", 
+        index = c("absolute", "dbp", "core_abundance", "gini", "dmn",  "relative",
+                  "simpson_lambda"),
```
TuomasBorman: Move these functions under estimate*.R file. Move the functionality of estimateDiversity under .estimate_diversity. --> Then we can use these functions in the future internally without warnings

R/estimateAlpha.R
```diff
@@ -0,0 +1,274 @@
...
+#'   indices.
+#'   
+#' @param ... optional arguments.
+#' 
+#' @param n.iter a single \code{integer} value for the number of rarefaction
+#' rounds.
+#' 
+#' @param rarefaction.depth a \code{double} value as for the minimim size or 
+#' rarefaction.depth. (default: \code{min(colSums(assay(x, "counts")), na.rm = TRUE)})
+#'
```
TuomasBorman: I think the default value should be NULL --> no rarefaction. It is more clear, I think

TuomasBorman (changes requested):
Final comments.

Lastly, run BiocCheck::BiocCheck() and fix warnings/notes related to this function (Check the mia.BiocCheck folder and the file inside it)

R/estimateAlpha.R
ChouaibB: ?
Is this not working?

TuomasBorman: Yes, it is working. However, it is little bit hard to follow. I am thinking if we need to debug things, so it is easier that thinkgs are done in steps rsther than with one go

ChouaibB: I do think it's easy to follow and it's straight forward. What is complicated there at the end of the day?

ChouaibB: it is perfectly fine in my opinion, you could perhaps change i tin future fixes and commits or what so ever.. :)

TuomasBorman: It's not too complicated but it is in general easier to debug when there are clear steps (I dont say that there isnt already)

This is ok like this but good to keep in mind; it is sometimes quite hard to debug someone else's code so it is better to take this into account already when creating the function

Moreover, Bioc guidelines recommend similar coding style. These things are just for the sake that the psckages are as good as possible

ChouaibB: I do think that using dplyr grammar for coding is already clear...

ChouaibB:
> Final comments.
> 
> Lastly, run BiocCheck::BiocCheck() and fix warnings/notes related to this function (Check the mia.BiocCheck folder and the file inside it)

The earlier devtools::check() there were no error (ofc otherwise won't commit) and no warnings, few notes that regards other functions. But I could test with BiocCheck::BiocCheck() now.

R/estimateAlpha.R
TuomasBorman: As I said, this is good like this, but we have preferred base grammar, no big deal

TuomasBorman:
> > Final comments.
> > 
> > Lastly, run BiocCheck::BiocCheck() and fix warnings/notes related to this function (Check the mia.BiocCheck folder and the file inside it)
> 
> The earlier devtools::check() there were no error (ofc otherwise won't commit) and no warnings, few notes that regards other functions. But I could test with BiocCheck::BiocCheck() now.

Yes, bioccheck cheks the coding style also

R/estimateAlpha.R
TuomasBorman: Easiest and tidiest way is to just rename estimateDiversity with .estimate_diversity and create a new exported function estimateDiversity which utilizes this "old" function

Cannot see if you have already made changes because Im on my phone

R/estimateAlpha.R
TuomasBorman: Related to rarefactiln.depth parameter

Or mayne we could use TRUE/FALSE values and then numeric

TRUE --> default rarefsction with message that says the rarefaction depth

FALSE --> no rarefaction

Numeric --> use user-defined depth

ChouaibB:
I couldn't get the BiocCheck installed but the results of the devtools::check are as follows (complete log attached):
image
[00check.log](https://github.com/microbiome/mia/files/13226620/00check.log)
It seems to be there are no warning regarding the function but perhaps others.

Anyway, you could please do whatever you want with these commits and the issue.
Thanks and good luck! :)

TuomasBorman:
Ok, there was atleast some indentation problems (Bioc has a rule that indentation must be multiplication of 4 spaces)

But these are easy and fast to fix, thanks!

ChouaibB:
Thank you! good luck with all! :)

antagomir:
up - if anything more needed from my side to close this PR let me know

TuomasBorman:
At least the name should be addAlpha to follow new naming scheme (related to that we can have a discussion about naming, I will send you msg)

antagomir:
@ChouaibB can you follow the suggestion from @TuomasBorman and update? Then actively follow-up on this until the PR is approved by us and merged successfully.

antagomir:
Related to #417

TuomasBorman:
This looks good, the name should be addAlpha- And also it might be best to move deprecated functions to deprecate.R file

antagomir:
Ok @ChouaibB let's take this up again and aim to finalize.

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
+#' Estimate alpha diversity indices.
```
antagomir: header line should not include period (Bioc guideline)

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' 
+#' The function estimates alpha diversity indices optionally using of rarefaction,
```
antagomir: Remove "The function" and capitalize estimates; 

"using of" -> "using";

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' then stores results at \code{\link{colData}}.
```
antagomir: at -> in

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' @param x a \code{\link{SummarizedExperiment}} object.
+#' 
+#' @param assay.type the name of the assay used for
+#'   calculation of the sample-wise estimates (default: \code{assay.type = "counts"}).
+#'   
+#' @param index a \code{character} vector, specifying the alpha diversity indices
+#'   to be calculated.
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
```
antagomir: can we use @inheritParams for arguments that are shared with other functions?

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' tse$shannon
+#'
+#' # Calculate observed richness with 10 rarefaction rounds
+#' tse <- addAlpha(tse, assay.type = "counts", index = "observed_richness",
+#' rarefaction.depth=min(colSums(assay(tse, "counts")), na.rm = TRUE), n.iter=10)
+#' 
+#' # Shows the estimated observed richness
+#' tse$observed_richness
+#' 
+
```
antagomir: remove this empty line

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' 
+#' # Calculate the default Shannon index with no rarefaction
+#' tse <- addAlpha(tse, assay.type = "counts", index = "shannon")
+#' 
+#' # Shows the estimated Shannon index
+#' tse$shannon
+#'
+#' # Calculate observed richness with 10 rarefaction rounds
+#' tse <- addAlpha(tse, assay.type = "counts", index = "observed_richness",
+#' rarefaction.depth=min(colSums(assay(tse, "counts")), na.rm = TRUE), n.iter=10)
```
antagomir: split on multiple lines, one line per argument for easier read

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+#' # Shows the estimated observed richness
+#' tse$observed_richness
+#' 
+
+#' @name addAlpha
+#' @rdname addAlpha
+#' @export
+setGeneric(
+    "addAlpha", signature = c("x"), function(
+        x, assay.type = "counts",
```
antagomir: Check: is this PR suggesting an entirely new function "addAlpha", or is this modification of the already existing function "addAlpha" that did not have rarefaction before?

Ideally, this PR will extend the existing alpha diversity functionality and only add the parts that are needed to add rarefaction.

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+    "addAlpha", signature = c("x"), function(
+        x, assay.type = "counts", 
+        index = c(
+            "coverage_diversity", "fisher_diversity", "faith_diversity",
+            "gini_simpson_diversity", "inverse_simpson_diversity",
+            "log_modulo_skewness_diversity", "shannon_diversity",
+            "absolute_dominance", "dbp_dominance", "core_abundance_dominance",
+            "gini_dominance", "dmn_dominance", "relative_dominance",
+            "simpson_lambda_dominance", "camargo_evenness", "pielou_evenness",
+            "simpson_evenness", "evar_evenness", "bulla_evenness",
```
antagomir: fix intendations

R/addAlpha.R
```diff
@@ -0,0 +1,229 @@
...
+    "addAlpha", signature = c(x = "SummarizedExperiment"), function(
+        x, assay.type = "counts", 
+        index = c(
+            "coverage_diversity", "fisher_diversity", "faith_diversity",
+            "gini_simpson_diversity", "inverse_simpson_diversity",
+            "log_modulo_skewness_diversity", "shannon_diversity",
+            "absolute_dominance", "dbp_dominance", "core_abundance_dominance",
+            "gini_dominance", "dmn_dominance", "relative_dominance",
+            "simpson_lambda_dominance", "camargo_evenness", "pielou_evenness",
+            "simpson_evenness", "evar_evenness", "bulla_evenness",
```
antagomir: fix intendations

R/estimateEvenness.R
```diff
@@ -114,28 +114,25 @@
...
-                              "bulla"),
-                    name = index, ...)
-               standardGeneric("estimateEvenness"))
+setGeneric(
+    ".estimate_evenness",signature = c("x"),
+    function(
+        x, assay.type = assay_name, assay_name = "counts",
+        index = c("pielou", "camargo", "simpson_evenness", "evar", "bulla"),
+        name = index, ...)
+        standardGeneric(".estimate_evenness"))
```
antagomir: is there an actual change here? if not, then it would be better to keep the original format so we can focus the review on the actual changes. Same comment goes for all content.

antagomir (changes requested):

antagomir:
I will add some comments - wait until that before further changes.

TuomasBorman:
Also important to check that the code follows Biconductor guidelines.

1. BiocCheck::BiocCheck()
2. Fix issues related to these lines that are added by this PR (there are not many. I quickly looked that indentation was off in couple of lines)

R/estimateEvenness.R
ChouaibB: yes: estimate_evenness --> .estimate_evenness

R/addAlpha.R
ChouaibB: This is PR is about to add the `addAlpha` function and also depricate and change previous estimate*Alpha* functions.

R/addAlpha.R
ChouaibB: umm.. I tried to check but couldn't spot a benefit here? since now all the estimate*alpha* functions are transferred to the current implementation? (kinda addAlpha now is the parent function?)

antagomir: Yes, addAlpha is the main function.

R/addAlpha.R
antagomir: Perfect.

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
+#' Estimate alpha diversity indices
+#' 
+#' The function estimates alpha diversity indices optionally using rarefaction,
```
antagomir, line 3: Change: "The function estimates" -> "Estimate"

antagomir: Remove the words: "optionally using rarefaction," (this will be explained later in arguments and @details field)

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#' @param x a \code{\link{SummarizedExperiment}} object.
+#' 
+#' @param assay.type the name of the assay used for calculation of the
+#'   sample-wise estimates (default: \code{assay.type = "counts"}).
+#'   
+#' @param index a \code{character} vector, specifying the alpha diversity 
+#'   indices to be calculated.
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
```
antagomir: We could use @inheritParams for : "assay.type" and "name" arguments from some other functions?

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#'   indices(By default: \code{name = index}).
+#'   
+#' @param ... optional arguments.
```
antagomir: This "..." should be the last argument to explain (by convention).

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#' 
+#' @param n.iter a single \code{integer} value for the number of rarefaction
+#'    rounds (By default: \code{n.iter = 10}).
+#' 
+#' @param rarefaction.depth a \code{double} value as for the minimim size or
```
antagomir: minimim -> minimum

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#'   
+#' @param ... optional arguments.
+#' 
+#' @param n.iter a single \code{integer} value for the number of rarefaction
+#'    rounds (By default: \code{n.iter = 10}).
+#' 
+#' @param rarefaction.depth a \code{double} value as for the minimim size or 
+#'    rarefaction.depth. (By default: \code{rarefaction.depth = NULL})
+#' 
+#' @return \code{x} with additional \code{\link{colData}} named after the index
```
antagomir: I think `\code{x}` should be replaced by `\code{\link{SummarizedExperiment}}`?

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#' 
+#' # Shows the estimated Shannon index
+#' tse$shannon
+#'
+#' # Calculate observed richness with 10 rarefaction rounds
+#' tse <- addAlpha(tse,
+#'  assay.type = "counts",
+#'  index = "observed_richness",
+#'  rarefaction.depth=min(colSums(assay(tse, "counts")), na.rm = TRUE),
+#'  n.iter=10)
```
antagomir: indentations should be multiples of 4 spaces (Bioc guidelines)

R/addAlpha.R
antagomir: If this is not given, does it use min(colSums(assay(tse, "counts")), na.rm = TRUE) by default?

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+    detected[["search"]] <- index
+    return(detected)
+}
+
+# This function rarifies the data n.iter of times and calculates index for the
+# rarified data. The result is a mean of the iterations.
+#' @importFrom DelayedMatrixStats colMeans2
+.alpha_rarefaction <- function(
+        x, assay.type, n.iter, rarefaction.depth, FUN, index, name, ...){
+    # Calculating the mean of the subsampled alpha estimates ans storing them
```
antagomir, line 597: ans -> and

R/addAlpha.R
```diff
@@ -0,0 +1,234 @@
...
+#'   indices to be calculated.
+#'   
+#' @param name a name for the column(s) of the colData the results should be
+#'   stored in. By default this will use the original names of the calculated
+#'   indices(By default: \code{name = index}).
+#'   
+#' @param ... optional arguments.
+#' 
+#' @param n.iter a single \code{integer} value for the number of rarefaction
+#'    rounds (By default: \code{n.iter = 10}).
```
antagomir: By default rarefaction should not be applied.

If n.iter=1 then rarefaction should be applied once at the given rarefaction depth. That would be the standard single-round rarification.

By default n.iter should be switched off. This could be n.iter=0, or n.iter=FALSE, or n.iter=NULL. Perhaps the last one n.iter=NULL would be the least ambiguous. Also n.iter=0 should lead to the same result, calculating alpha diversity without any rarefaction.

Add here the following explanation:  "Rarefaction is not applied by default (n.iter=NULL); if n.iter is a positive integer then rarefaction is applied (see @details section)." -> Then in @details explain how n.iter and rarefaction.depth operate.

antagomir: Unit tests should then check those conditions as well.

antagomir (commented):
Looks very good for now.

antagomir:
Some remaining comments + BiocChecks will be a must.

We should probably aim to merge this very soon, then do some further testing on the fresh new version.

R/addAlpha.R
ChouaibB: It was using it before but now it's set to NULL by default.

antagomir: Ok - what happens when rarefaction is activated, and this has its default value (NULL) - no rarefaction?

ChouaibB: When NULL no rarefaction, when not NULL it is checked to be strictly positive number to activate rarefaction:
```
if( !is.null(rarefaction.depth) && 
            !(is.numeric(rarefaction.depth) && rarefaction.depth > 0)) {
            stop("'rarefaction.depth' must be a non-zero positive double.",
                call. = FALSE)
        }
```
I thought this was as requested before? I could try change it to the implementation suggested at the below comment if needed?

antagomir: We have two rarefaction parameters: rarefaction.depth and n.iter; I think one of these should be enough to activate rarefaction, otherwise users might get confused if two different things need to be activated to make it work. Perhaps an intuitive solution would be to use n.iter to activate rarefaction, and if it is activated then the rarefaction depth must be defined in any case and min(readcounts) is a standard default. 

This might have been requested to be NULL before we had the n.iter and multiple rounds added, at that time it made sense as it was the only rarefaction parameter.

Now reconsidering this, how about the following:

- n.iter=0 by default -> no rarefaction (same result would occur if n.iter=NULL or n.iter=FALSE)
- n.iter>0 (should be integer; should be checked) -> rarefaction + rarefaction.depth would be applied  

rarefaction.depth default value min(colSums(assay(tse, assay.type))) and this is applied only when n.iter>0

This procedure shuld be explained in the argument description and in @details field

This would be intuitive for a user?

ChouaibB: Everything is fine by me. Initially the default for rarefaction.depth was that way.
Alright, I re-implement it.

antagomir:
Can we utilize the rarefyAssay function directly (see #536) - if not done yet?

R/addAlpha.R
antagomir: Yes - but this does not seem to require full reimplementation, mainly changing the argument default values and how to deal with them.

TuomasBorman: `n.iter` controls whether rarefaction is applied. `sample` is passed with `...` to `rarefyAssay()` so it is not handled in this function

R/addAlpha.R
TuomasBorman: We can do this to all manual pages at the same time

TuomasBorman:
I think this is good to go. @ake123 Do you have time to check that there are necessary unit tests (there might be already)? Check Leo's comments about testing whether rarefying was disabled (n.iter = NULL and n.iter = 0)

ake123:
Sure I can check that

> I think this is good to go. @ake123 Do you have time to check that there are necessary unit tests (there might be already)? Check Leo's comments about testing whether rarefying was disabled (n.iter = NULL and n.iter = 0)

R/addAlpha.R
```diff
@@ -0,0 +1,632 @@
...
+#'   \item \code{aggregate}: Optional and only used for the \code{Absolute},
+#'   \code{dbp}, \code{Relative}, and \code{dmn} dominance indices.
+#'   Aggregate the values for top members selected by \code{ntaxa} or not. If
+#'   \code{TRUE}, then the sum of relative abundances is returned. Otherwise the
+#'   relative abundance is returned for the single taxa with the indicated rank
+#'   (default: \code{aggregate = TRUE}).
+#'   
+#'   \item \code{detection}: A \code{numeric} value for selecting detection
+#'   threshold for the abundances for richness indices. (Default: \code{0})
+#' }
```
TuomasBorman: TODO: What these parameters do? Which parameter affects which index?

TuomasBorman (changes requested):

TuomasBorman:
Fixed some issues and updated documentation. Would be nice to have someone else's thoughts (if everything works) but I think this is good now

TuomasBorman:
One last thing: I changed `n.iter` to `niter` to harmonize the paramater names. `nsomething` is used to denote number of something. We have already `nclasses` and `ndimred`.

If someone is against this change, I'm happy to discuss about this

antagomir (approved):
Looks good!

ake123:
It seems the unit test have been updated with niter=NULL, niter=0 as default and with no errors.

TuomasBorman (approved):

TuomasBorman:
@ake123 Can you check that OMA and mia* packages are updated?

ake123:
Yes sure! Should I add section addAlpha in the diversity section?

Issue #450, opened 2023-09-22
Clean up the manpages


antagomir:
The html manpages do not always render well, see e.g.
https://microbiome.github.io/mia/reference/getExperimentCrossAssociation.html

0xMuluh:
Do we have pending tasks to clean up rendering in manpages?

antagomir:
Could you for instance browse through all reference pages in mia, miaViz, miaTime and see if you notice anything that could be easily improved / clarified? 

If examples throw warnings it might be good to either modify the example to get rid of warnings, or comment on why those warnings appear and show how to deal with them.

This task is secondary priority if more urgent tasks pop up.

0xMuluh:
Sure will take a look while pending feedback on some commits I pushed earlier.

antagomir:
Status of this one @Daenarys8 ?

0xMuluh:
mia and  miaViz  are addressed in #603 and https://github.com/microbiome/miaViz/pull/140 respectively.
Will add a PR for miaTime.

antagomir:
Ok can we close this one or better to keep open still?

0xMuluh:
I think we can close.

antagomir:
Great. Please actively close all issues as they become ready.

PR #465, opened 2023-10-18
Manpage clean-up

jepasan:
I've gone through the reference pages and cleaned up some unwieldy example outputs as suggested in #450. I mainly spotted three big issues.

- In 41688b42620ef319909e51bb08dff6ddf13b3f49, I've used `head()` to avoid printing the whole probability matrix in https://microbiome.github.io/mia/reference/calculateDMN.html.
- In 13a55fb98987868c67f8522195f4e81a4f6f54f9, I've removed rows with zero standard deviation from the assay to eliminate the large warning chunks in https://microbiome.github.io/mia/reference/getExperimentCrossAssociation.html. The second assay used in the example also has some NA values, but I haven't changed those since they don't produce too many warnings, and one of the examples demonstrates the use of `show_warnings=FALSE`.
- In ecf30870a57b1715ecc8f9323e8d362efe22f587, I've reduced the number of artificially created groups to reduce repetitive warnings in https://microbiome.github.io/mia/reference/splitOn.html. These warnings ultimately stem from `addTaxonomyTree`, which is called for each split object created from the original when `update_rowTree = TRUE`. I don't think this is the best solution, perhaps these could be suppressed instead?

In addition, I've fixed some grammar mistakes here and there.

### DIFF
--- a/R/agglomerate.R
+++ b/R/agglomerate.R
@@ -1,7 +1,7 @@
 #' Agglomerate data using taxonomic information
 #'
-#' \code{agglomerateByRank} can be used to sum up data based on the association
-#' to certain taxonomic ranks given as \code{rowData}. Only available
+#' \code{agglomerateByRank} can be used to sum up data based on associations
+#' with certain taxonomic ranks, as defined in \code{rowData}. Only available
 #' \code{\link{taxonomyRanks}} can be used.
 #'
 #' @param x a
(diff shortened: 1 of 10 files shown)


antagomir:
Thanks!

Could @TuomasBorman check what to do with "n https://github.com/microbiome/mia/commit/ecf30870a57b1715ecc8f9323e8d362efe22f587, I've reduced the number of artificially created groups to reduce repetitive warnings in https://microbiome.github.io/mia/reference/splitOn.html. These warnings ultimately stem from addTaxonomyTree, which is called for each split object created from the original when update_rowTree = TRUE. I don't think this is the best solution, perhaps these could be suppressed instead?"

antagomir (approved):
Overall, seems very good. I don't really have anything to add here. We can merge when checks complete.

PR #476, opened 2023-11-28
taxonomyranks

ake123:
Add setTaxonomyRanks and getTaxonomyRanks functions

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -47,6 +47,7 @@ reference:
   - getPrevalentAbundance
   - getTopTaxa
   - getExperimentCrossCorrelation
+  - hierarchy-tree
 - subtitle: Dominance
 - contents:
   - perSampleDominantTaxa
@@ -80,3 +81,6 @@ reference:
 - subtitle: Package
 - contents:
   - mia-package
+- subtitle: Deprecated functions
+- contents:
+  - deprecate


R/taxonomy.R
```diff
@@ -209,6 +209,15 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
 
+# Function to set taxonomy ranks
+.setTaxonomyRanks <- function(rank = TAXONOMY_RANKS) {
+        assign("TAXONOMY_RANKS", rank, envir = .GlobalEnv)
+}
+
+# Function to get taxonomy ranks
+.getTaxonomyRanks <- function(x) {
+    return(TAXONOMY_RANKS)
+}
```
TuomasBorman: Add @export tag
Function names without the dot. Dot means that the function is internal and is not exported.

R/taxonomy.R
```diff
@@ -134,7 +134,7 @@ NULL
 #' @format a \code{character} vector of length 8 containing the taxonomy ranks
 #'   recognized. In functions this is used as case insensitive.
 #' @export
-TAXONOMY_RANKS <- c("domain","kingdom","phylum","class","order","family",
+TAXONOMY_RANKS <- c("domain","superkingdom","kingdom","phylum","class","order","family",
```
TuomasBorman: I'm not sure if we want to add superkingdom here since domain and superkingdom mean same thing

TuomasBorman:
Can you also update loadFromBiom? I have seen lots of BIOM files that have "sk__Bacteria" for instance. Currently, column that have this kind of values is not detected as superkingdom. (For example, a column with "k__kingdom" value is detected as "KIngdom" column)

R/taxonomy.R
ake123: corrected

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -243,7 +243,7 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
     col = x[ , colname]
     # List prefixes
     prefixes <- c(
-        "^d__",
+        "^sk__",
```
TuomasBorman: Just add sk__, don't remove d__

TuomasBorman: Ahh, now I know why you have done this

Hmmm, I think solution is to replace TAXONOMY_RANKS with your own variable taxonomy_ranks <- c("domain", "superkingdom", "kingdom"...)

That is because now user can modify TAXONOMY_RANKS, but those prefixes are still hardcoded

TuomasBorman: Or maybe better

prefixes <- c(domain = "d__", superkingdom = "sk__"...)

And based on matches, get the corresponding name

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -243,14 +243,15 @@ makeTreeSummarizedExperimentFromBiom <- function(obj, ...){
...
+        kingdom = "^k__",
+        phylum = "^p__",
+        class = "^c__",
+        order = "^o__",
+        family = "^f__",
+        genus = "^g__",
+        species = "^s__"
     )
     # Find which prefix is found from each column value, if none.
     found_rank <- lapply(prefixes, FUN = function(pref){
```
TuomasBorman: This is not working yet. You have to replace

colname <- TAXONOMY_RANKS[found_rank]

with

colname <- names(prefix)[found_rank]

R/taxonomy.R
```diff
@@ -209,6 +208,19 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+# Function to set taxonomy ranks
+setTaxonomyRanks <- function(rank, namespace = "mia") {
+    rank <- tolower(rank)
+    assignInNamespace("TAXONOMY_RANKS", rank, ns = asNamespace(namespace))
+    cat("Taxonomy ranks set to:", rank, "\n")
+}
+
+#'@export
+# Function to get taxonomy ranks
+getTaxonomyRanks <- function(namespace = "mia") {
```
TuomasBorman: User should not be able to modify namespace parameter, so remove it

R/taxonomy.R
```diff
@@ -209,6 +208,19 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
     }
 )
 
+#'@export
+# Function to set taxonomy ranks
+setTaxonomyRanks <- function(rank, namespace = "mia") {
+    rank <- tolower(rank)
+    assignInNamespace("TAXONOMY_RANKS", rank, ns = asNamespace(namespace))
```
TuomasBorman: Instead of namespace parameter, hardcode "mia"

R/taxonomy.R
```diff
@@ -209,6 +208,19 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+    cat("Taxonomy ranks set to:", rank, "\n")
```
TuomasBorman: Printing is not necessary, because user knows that he/she has changed the ranks

(moreover printing (print(), cat()) is not recommended way to print, because user cannot disable those. Preferable way is to use message(), warning(), or stop()

R/taxonomy.R
```diff
@@ -128,12 +128,22 @@
 #' x <- GlobalPatterns
 #' x <- addTaxonomyTree(x)
 #' x
+#' #setTaxonomyRanks and getTaxonomyRanks
+#' tse <- GlobalPatterns
+#' colnames(rowData(tse))[1] <- "TAXA1"
+#' 
+#' setTaxonomyRanks(colnames(rowData(tse)))
+#' #Taxonomy ranks set to: taxa1 phylum class order family genus species
```
TuomasBorman: First the comment and then the code. Also, Add space after #

R/taxonomy.R
```diff
@@ -128,12 +128,22 @@
...
 #' x <- addTaxonomyTree(x)
 #' x
+#' #setTaxonomyRanks and getTaxonomyRanks
+#' tse <- GlobalPatterns
+#' colnames(rowData(tse))[1] <- "TAXA1"
+#' 
+#' setTaxonomyRanks(colnames(rowData(tse)))
+#' #Taxonomy ranks set to: taxa1 phylum class order family genus species 
+#' 
+#' getTaxonomyRanks()
```
TuomasBorman: Add explanation what it does. Line 139 is not necessary --> remove

R/taxonomy.R
```diff
@@ -208,7 +218,19 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
         ans
     }
 )
-
+#' @rdname taxonomy-methods
+#'@export
+# Function to set taxonomy ranks
+setTaxonomyRanks <- function(rank) {
+    rank <- tolower(rank)
```
TuomasBorman: Add here input test. rank should be a character vector. Length >= 1.  Check other function for help. I think .is_a_character is a correct function for checking

R/taxonomy.R
```diff
@@ -208,7 +218,19 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+    assignInMyNamespace("TAXONOMY_RANKS", rank)
+}
+#' @rdname taxonomy-methods
+#'@export
+# Function to get taxonomy ranks
+getTaxonomyRanks <- function() {
+    return(TAXONOMY_RANKS)
+}
```
TuomasBorman: Add space after a functiion for consistency

R/taxonomy.R
```diff
@@ -209,6 +219,25 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
     }
 )
 
+#' @rdname taxonomy-methods
+#'@export
+# Function to set taxonomy ranks
+setTaxonomyRanks <- function(rank) {
+    rank <- tolower(rank)
+    # Check if rank is a character vector with length >= 1
+    if (!is.character(rank) || length(rank) < 1  || any(rank == "" | rank == " " | rank == "\t" | rank == "-" | rank == "_") || any(grepl("\\s{2,}", rank))) {
```
TuomasBorman: I think it is enough to check that the vector is character and length is positive. Or does these give any addtional value?

R/taxonomy.R
```diff
@@ -209,6 +219,25 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+        stop("Input 'rank' should be a character vector with non-empty strings, no spaces, tabs, hyphens, underscores, and non-continuous spaces.")
```
TuomasBorman: Add here call. = FALSE

TuomasBorman: Run also BiocCheck --> indentation should be multiplication of 4 spaces and the width of line should be less than 80 characters.

TuomasBorman:
What do you think? When user imports data with loadFromMetaphlan or similar imported, should we automatically set the taxonomyRanks based on colnames(rowData(tse))? There could be parameter for that... The downside is that there might be some taxonomy ranks that importer is not detecting...

ake123:
As you noted, there's a risk that the importer might not detect some taxonomy ranks, especially if the data contains unconventional or mislabeled column names. This could lead to incomplete or incorrect taxonomy assignment, potentially affecting downstream analysis.

TuomasBorman:
Well how about a new internal function something like `.set_taxonomy_ranks` --> it has parameter `detect.ranks` (TRUE/FALSE) --> it sets the ranks and give warning "Ranks are set to rank1, rank2... (Use setTaxonomyRanks() to overwrite ranks)."

We can use it hidden and see if it is good or not.

ake123:
Added the `.set_taxonomy_ranks` internal function.

antagomir:
Yes I think this is a perfect solution.

R/taxonomy.R
```diff
@@ -209,6 +219,42 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+    if (!is.character(rank) || length(rank) < 1 
+        || any(rank == "" | rank == " " | rank == "\t" | rank == "-" | rank == "_")
+        || any(grepl("\\s{2,}", rank))) {
+        stop("Input 'rank' should be a character vector with non-empty strings,
+             no spaces, tabs, hyphens, underscores, and non-continuous spaces."
+             , call. = FALSE)
+    }
+    .set_taxonomy_ranks(rank, detect.ranks = TRUE)
+    #assignInMyNamespace("TAXONOMY_RANKS", rank)
+}
```
TuomasBorman: Hmm, remove line 235 and uncomment line 236. I think it is more clear that way

Then .set_taxonomy_ranks is used only in importing functions

R/taxonomy.R
```diff
@@ -209,6 +219,42 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+
+.set_taxonomy_ranks <- function(rank, detect.ranks = FALSE) {
+    if (detect.ranks) {
+        # Assuming 'rank1, rank2, ...' are placeholders for actual rank values in 'rank'
+        detected_ranks <- paste(rank, collapse=", ")
+        warning(sprintf("Ranks are set to %s. (Use setTaxonomyRanks() to overwrite ranks).", detected_ranks))
+    }
+    # Assign validated rank to TAXONOMY_RANKS in a specific namespace
+    assignInMyNamespace("TAXONOMY_RANKS", rank)
```
TuomasBorman: Instead of this line, call setTaxonomyRanks(ranks)

R/taxonomy.R
```diff
@@ -209,6 +219,42 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
     }
 )
 
+#' @rdname taxonomy-methods
+#'@export
+# Function to set taxonomy ranks
+setTaxonomyRanks <- function(rank) {
```
TuomasBorman: Instead of rank --> ranks?

R/taxonomy.R
```diff
@@ -209,6 +219,42 @@ setMethod("checkTaxonomy", signature = c(x = "SummarizedExperiment"),
...
+        || any(grepl("\\s{2,}", rank))) {
+        stop("Input 'rank' should be a character vector with non-empty strings,
+             no spaces, tabs, hyphens, underscores, and non-continuous spaces."
+             , call. = FALSE)
+    }
+    .set_taxonomy_ranks(rank, detect.ranks = TRUE)
+    #assignInMyNamespace("TAXONOMY_RANKS", rank)
+}
+
+.set_taxonomy_ranks <- function(rank, detect.ranks = FALSE) {
```
TuomasBorman: Maybe set.ranks instead of detect.ranks? --> The ranks are not "detected", they are just set based on colnames(rowData(x))...

This function could get TreeSE, ranks and set.ranks as input.
1. TreeSE: if ranks are not specified by user, get ranks from colnames(rowData(x))
2. ranks: User can specify ranks manually.
3. set.ranks: Set ranks based on user-specified ranks parameter or colnames(rowData(x))

TuomasBorman (changes requested):

TuomasBorman:
Hi! Can you do PR again but not in mia (it is easier to review since then I can also run it locally without cloning your fork). The changes in this PR were good

ake123:
I seems I lost it in this PR I have made a new one

PR #477, opened 2023-11-28
cluster: fixes

TuomasBorman:


TuomasBorman:
In this PR:

1. Simplifies the code.
2. Get `altExp` and `MARGIN` with common internal functions.
3. Store the result with common internal functions.
4. Overwrite the results if the name of column and metadata is the same. (Same behaviour as in other functions)

PR #478, opened 2023-11-30
Fix critical issue with mergeFeatures

0xMuluh:
Ping #419 @antagomir @TuomasBorman 

- [X] The default value of rank in mergeFeaturesByRank should be NULL
- [X]  Add na.rm parameter to mergeFeatures
- [X]  Change the default to onRankOnly=TRUE in mergeFeaturesByRank
- [X]  Respect the rowData order

TODO
- [ ]  check other functions if they have similar sorting issues
- [ ]  Add fix on reviews from this PR

### DIFF
--- a/R/agglomerate.R
+++ b/R/agglomerate.R
@@ -1,21 +1,21 @@
 #' Agglomerate or merge data using taxonomic information
-#' 
-#' Agglomeration functions can be used to sum-up data based on specific criteria 
+#'
+#' Agglomeration functions can be used to sum-up data based on specific criteria
 #' such as taxonomic ranks, variables or prevalence.
 #'
 #' \code{agglomerateByRank} can be used to sum up data based on associations
 #' with certain taxonomic ranks, as defined in \code{rowData}. Only available
-#' \code{\link{taxonomyRanks}} can be used. 
-#' 
-#' \code{agglomerateByVariable} merges data on rows or columns of a 
-#' \code{SummarizedExperiment} as defined by a \code{factor} alongside the 
-#' chosen dimension. This function allows agglomeration of data based on other 
+#' \code{\link{taxonomyRanks}} can be used.
+#'
+#' \code{agglomerateByVariable} merges data on rows or columns of a
+#' \code{SummarizedExperiment} as defined by a \code{factor} alongside the
+#' chosen dimension. This function allows agglomeration of data based on other
 #' variables than taxonomy ranks.
-#' Metadata from the \code{rowData} or \code{colData} are 
+#' Metadata from the \code{rowData} or \code{colData} are
 #' retained as defined by \code{archetype}.
-#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} are 
-#' agglomerated, i.e. summed up. If the assay contains values other than counts 
-#' or absolute values, this can lead to meaningless values being produced. 
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} are
+#' agglomerated, i.e. summed up. If the assay contains values other than counts
+#' or absolute values, this can lead to meaningless values being produced.
 #'
 #' @param x a \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}} or
 #'   a \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
@@ -39,14 +39,14 @@
 #' @param agglomerate.tree \code{TRUE} or \code{FALSE}: should
 #'   \code{rowTree()} also be agglomerated? (Default:
 #'   \code{agglomerate.tree = FALSE})
-#' 
+#'
 #' @param agglomerateTree alias for \code{agglomerate.tree}.
 #'
 #' @param ... arguments passed to \code{agglomerateByRank} function for
 #'   \code{SummarizedExperiment} objects,
 #'   to \code{\link[=agglomerate-methods]{agglomerateByVariable}} and
 #'   \code{\link[scuttle:sumCountsAcrossFeatures]{sumCountsAcrossFeatures}},
-#'   to \code{getPrevalence} and \code{getPrevalentTaxa} and used in 
+#'   to \code{getPrevalence} and \code{getPrevalentTaxa} and used in
 #'   \code{agglomeratebyPrevalence}
 #'   \itemize{
 #'        \item \code{remove_empty_ranks}: A single boolean value for selecting 
@@ -74,7 +74,7 @@
 #'   \code{strip_altexp = TRUE})
 #'
 #' @param MARGIN A character value for selecting if data is merged
-#'   row-wise / for features ('rows') or column-wise / for samples ('cols'). 
+#'   row-wise / for features ('rows') or column-wise / for samples ('cols').
 #'   Must be \code{'rows'} or \code{'cols'}.
 #'
 #' @param f A factor for merging. Must be the same length as
@@ -87,7 +87,7 @@
 #'   This can be single integer value or an integer vector of the same length
 #'   as \code{levels(f)}. (Default: \code{archetype = 1L}, which means the first
 #'   element encountered per factor level will be kept)
-#'   
+#'
 #' @param mergeTree \code{TRUE} or \code{FALSE}: Should
 #'   \code{rowTree()} also be merged? (Default: \code{mergeTree = FALSE})
 #'
@@ -127,7 +127,7 @@
 #' optionally-pruned object of the same class as \code{x}.
 #'
 #' @name agglomerate-methods
-#' 
+#'
 #' @seealso
 #' \code{\link[=splitOn]{splitOn}}
 #' \code{\link[=unsplitOn]{unsplitOn}}
@@ -138,9 +138,9 @@
 #' \code{\link[SingleCellExperiment:splitAltExps]{splitAltExps}}
 #'
 #' @examples
-#' 
+#'
 #' ### Agglomerate data based on taxonomic information
-#' 
+#'
 #' data(GlobalPatterns)
 #' # print the available taxonomic ranks
 #' colnames(rowData(GlobalPatterns))
@@ -151,16 +151,16 @@
 #' ## How many taxa before/after agglomeration?
 #' nrow(GlobalPatterns)
 #' nrow(x1)
-#' 
+#'
 #' # agglomerate the tree as well
 #' x2 <- agglomerateByRank(GlobalPatterns, rank="Family",
 #'                        agglomerate.tree = TRUE)
 #' nrow(x2) # same number of rows, but
 #' rowTree(x1) # ... different
 #' rowTree(x2) # ... tree
-#' 
-#' # If assay contains binary or negative values, summing might lead to 
-#' # meaningless values, and you will get a warning. In these cases, you might 
+#'
+#' # If assay contains binary or negative values, summing might lead to
+#' # meaningless values, and you will get a warning. In these cases, you might
 #' # want to do agglomeration again at chosen taxonomic level.
 #' tse <- transformAssay(GlobalPatterns, method = "pa")
 #' tse <- agglomerateByRank(tse, rank = "Genus")
@@ -170,20 +170,20 @@
 #' sum(is.na(rowData(GlobalPatterns)$Family))
 #' x3 <- agglomerateByRank(GlobalPatterns, rank="Family", na.rm = TRUE)
 #' nrow(x3) # different from x2
-#' 
-#' # Because all the rownames are from the same rank, rownames do not include 
-#' # prefixes, in this case "Family:". 
+#'
+#' # Because all the rownames are from the same rank, rownames do not include
+#' # prefixes, in this case "Family:".
 #' print(rownames(x3[1:3,]))
-#' 
+#'
 #' # To add them, use getTaxonomyLabels function.
 #' rownames(x3) <- getTaxonomyLabels(x3, with_rank = TRUE)
 #' print(rownames(x3[1:3,]))
-#' 
+#'
 #' # use 'remove_empty_ranks' to remove columns that include only NAs
-#' x4 <- agglomerateByRank(GlobalPatterns, rank="Phylum", 
+#' x4 <- agglomerateByRank(GlobalPatterns, rank="Phylum",
 #'                         remove_empty_ranks = TRUE)
 #' head(rowData(x4))
-#' 
+#'
 #' # If the assay contains NAs, you might want to consider replacing them,
 #' # since summing-up NAs lead to NA
 #' x5 <- GlobalPatterns
@@ -195,28 +195,28 @@
 #' assay(x5)[ is.na(assay(x5)) ] <- 0
 #' x6 <- agglomerateByRank(x5, "Kingdom")
 #' head( assay(x6) )
-#' 
+#'
 #' ## Look at enterotype dataset...
 #' data(enterotype)
 #' ## Print the available taxonomic ranks. Shows only 1 available rank,
 #' ## not useful for agglomerateByRank
 #' taxonomyRanks(enterotype)
-#' 
+#'
 #' ### Merge TreeSummarizedExperiments on rows and columns
-#' 
+#'
 #' data(esophagus)
 #' esophagus
 #' plot(rowTree(esophagus))
 #' # get a factor for merging
 #' f <- factor(regmatches(rownames(esophagus),
 #'                        regexpr("^[0-9]*_[0-9]*",rownames(esophagus))))
-#' merged <- agglomerateByVariable(esophagus, MARGIN = "rows", f, 
+#' merged <- agglomerateByVariable(esophagus, MARGIN = "rows", f,
 #'                                 mergeTree = TRUE)
 #' plot(rowTree(merged))
 #' #
 #' data(GlobalPatterns)
 #' GlobalPatterns
-#' merged <- agglomerateByVariable(GlobalPatterns, MARGIN = "cols", 
+#' merged <- agglomerateByVariable(GlobalPatterns, MARGIN = "cols",
 #'                                 colData(GlobalPatterns)$SampleType)
 #' merged
 NULL
@@ -242,7 +242,7 @@ setGeneric("agglomerateByVariable",
 #'
 #' @export
 setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
-    function(x, rank = taxonomyRanks(x)[1], onRankOnly = FALSE, na.rm = FALSE,
+    function(x, rank = taxonomyRanks(x)[1], onRankOnly = TRUE, na.rm = FALSE,
         empty.fields = c(NA, "", " ", "\t", "-", "_"), ...){
         # input check
         if(nrow(x) == 0L){
@@ -250,7 +250,7 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
                 call. = FALSE)
         }
         if(!.is_non_empty_string(rank)){
-            stop("'rank' must be an non empty single character value.",
+            stop("'rank' must be a non-empty single character value",
                 call. = FALSE)
         }
         if(!.is_a_bool(onRankOnly)){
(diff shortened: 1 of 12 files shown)


antagomir:
up

antagomir:
up -  @Daenarys8

R/agglomerate.R
```diff
@@ -155,16 +155,22 @@ setGeneric("mergeFeaturesByRank",
...
+        if (!is.null(rank)) {
+            if (!.is_non_empty_string(rank)){
+                stop("'rank' must be a non-empty single character value.", call. = FALSE)
+            }
+            if (!(rank %in% taxonomyRanks(x))) {
+                stop("'rank' must be a value from 'taxonomyRanks()'", call. = FALSE)
+            }
+        } else {
+            rank <- taxonomyRanks(x)[1]
         }
```
TuomasBorman: INsteqad of this, change default behavior of rank in agglomerateByPrevalence. --> there is no need to change the behavior here, NULL (i.e. no agglomeration) does nothing

However, you should change the default rank to NULL in agglomeratebyPrevalence --> when NULL, agglomerateByRank is skipped.

TuomasBorman: Even though, we discussed in the issue otherwise.... sorry about that

R/merge.R
```diff
@@ -158,19 +163,34 @@ setGeneric("mergeSamples",
...
+    
+    # if na.rm is TRUE, remove the empty, white-space, NA values from
+    # tree will be pruned later.
+    if( na.rm ){
+        x <- .remove_with_empty_taxonomic_info(x, tax_cols[col],
+                                               empty.fields)
+    }
     if(is.character(f) && length(f)==1 && f %in% colnames(rowData(x))){
         f <- factor(as.character(rowData(x)[, f]))
     }
```
TuomasBorman: The intention of na.rm was to control the merging. Currently, if grouping factor contains NA values, these rows/cols with NA values are completely dropped because there cannot be a factor group NA.

The idea of na.rm was to control those behavior. If na.rm =TRUE, the functions should work as it does currently. If na.rm=FALSE, the function should add a group for those values that are NA. So something like this -->

```

if na.rm is FALSE
   From f vector, convert NA values to character "NA"
Convert f vector to factor
Perform grouping/merging
```

TuomasBorman:
This does not yet take into account the row order. In mergeRows, the order of data is in alphabetical order. In agglomerateByRanks the order follows the order of rowData. As discussed, the order based on rowData might be  the best, but the implementation is harder without big advantages.

So I suggest that we modify agglomerateByRank so that it outputs the data in alphabetical order. I think that is the most straightforward and requires only one additional line in the end of the function.

TuomasBorman (changes requested):
See comments

0xMuluh:
> alphabetical order

by single line, are we referring to?
x <- sort(x)

R/agglomerate.R
```diff
@@ -226,6 +226,7 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
         # adjust rownames
         rownames(x) <- getTaxonomyLabels(x, empty.fields, ...,
                                         with_rank = FALSE, resolve_loops = FALSE)
+        rownames(x) <- sort(rownames(x))
```
TuomasBorman: Basically what this does is that it just mixes up the rownames. The data inside is not ordered, only the names are changed.

To get the desired behavior, add to the end of this function something like this

`x <- x[ order(rownames(x), ]`

TuomasBorman:
To recap:

1

Behavior of functions that are using agglomerateByRank internally is unexpected. Why? Because if user does not provide rank, the data is agglomerated to the highest rank by default. However, that should not be the case. For example, agglomerateByPrevalence should not do rank-agglomeration by default.

I think our approach is now incorrect. Instead of changing the default value of rank in agglomerateByRank to NULL, we should modify agglomerateByPrevalence.

-->

In agglomererateByPrevalence, catch rank parameter from `...`, If rank is specified, call agglomerateByRank.

2.

mergeRows and mergeCols drop off those instances that have NA values in grouping variable. That is because grouping vector is converted to factor. To include instances with NA values as "NA group", NA values could be converted to "NA". -->

this behavior should be controlled somehow, `na.rm` parameter could be perfect for that, but we should check if na.rm is used somewhere else in the function (fed with `...` to some other function)

3.

The default value of onRankOnly in agglomerateByRank should be TRUE by default.

4.

In the end of the agglomerateByRank function, order the data alphabetically based on rownames.

TuomasBorman:
Sorry for hassle. I think most of the points are not yet implemented.

R/agglomerate.R
0xMuluh: ACK

R/agglomerate.R
```diff
@@ -229,7 +229,7 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
         # Remove those columns from rowData that include only NAs
         x <- .remove_NA_cols_from_rowdata(x, ...)
         x <- .add_values_to_metadata(x, "agglomerated_by_rank", rank)
-        x
+        x <- x[ order(rownames(x)), ]
```
TuomasBorman, line 311: Good to add comment. What is this line doing? --> easier to debug and read the code afterwards

R/merge.R
```diff
@@ -158,32 +163,46 @@ setGeneric("mergeSamples",
...
         stop("'average' must be TRUE or FALSE.", call. = FALSE)
     }
     if( !.is_a_bool(check.assays) ){
         stop("'check.assays' must be TRUE or FALSE.", call. = FALSE)
     }
     #
+    # Make a vector from the taxonomic data.
+    rank <- taxonomyRanks(x)[1]
+    col <- which( taxonomyRanks(x) %in% rank )
+    tax_cols <- .get_tax_cols_from_se(x)
```
TuomasBorman: What is the intention of these lines? They are not even used?

R/merge.R
```diff
@@ -158,32 +163,46 @@ setGeneric("mergeSamples",
...
+
+    # if na.rm is TRUE, convert NA values to "NA" in f.
+    if (na.rm) {
+        f[is.na(f)] <- "NA"
+    }
```
TuomasBorman: Use similar coding style than in other parts of the function

`if( na.rm )`

Also add comment, why this is done. --> This is done because otherwise we lose NA values.

It is not clear from the code what is the intention

R/getPrevalence.R
```diff
@@ -629,18 +629,23 @@ setGeneric("agglomerateByPrevalence", signature = "x",
 #' @rdname agglomerate-methods
 #' @export
 setMethod("agglomerateByPrevalence", signature = c(x = "SummarizedExperiment"),
-    function(x, rank = taxonomyRanks(x)[1L], other_label = "Other", ...){
```
TuomasBorman, line 560: Sorry, I did not notice that agglomerateByPrevalence has also rank parameter. In this case, things get even more simpler.

--> Just change the default value of rank to NULL. It should work --> test

R/agglomerate.R
0xMuluh: ACK.

R/merge.R
0xMuluh: You're right. Sorry for the noise

R/merge.R
0xMuluh: ACK.

R/estimateDiversity.R
```diff
@@ -258,7 +258,7 @@ setMethod("estimateDiversity", signature = c(x="SummarizedExperiment"),
                              "inverse_simpson", "log_modulo_skewness", "shannon")
         index_string <- paste0("'", paste0(supported_index, collapse = "', '"), "'")
         if ( !all(index %in% supported_index) || !(length(index) > 0)) {
-            stop("'", paste0("'index' must be from the following options: '",supported_types), "'", call. = FALSE)
```
TuomasBorman: Good, you found bug. --> This should be index_string


Check if that works

R/merge.R
```diff
@@ -52,32 +52,41 @@
...
     }
     if( !.is_a_bool(check.assays) ){
         stop("'check.assays' must be TRUE or FALSE.", call. = FALSE)
     }
-    #
+
+    # This is done otherwise we lose NA values
+    if ( na.rm ) {
+        f[is.na(f)] <- "NA"
+    }
```
TuomasBorman: This is in wrong place. Check the function. f can also be name specifying variable from rowData/colData

R/merge.R
```diff
@@ -52,32 +52,41 @@
...
     if(is.character(f) && length(f)==1 && f %in% colnames(rowData(x))){
         f <- factor(as.character(rowData(x)[, f]))
     }
     else if(is.character(f) && length(f)==1 && f %in% colnames(colData(x))){
+        f[is.na(f)] <- "NA"
```
TuomasBorman: This does na.rm = TRUE without asking from user

TuomasBorman: Also, I am wondering what was the intention of having colData here. This is merge_rows.... We should probably fix this

R/merge.R
```diff
@@ -52,32 +52,41 @@
...
         f <- factor(as.character(colData(x)[, f]))
-    } else 
+    } else
     {
-        f <- .norm_f(nrow(x), f)  
+        f <- .norm_f(nrow(x), f)
     }
```
TuomasBorman: 1. Remove na.rm from .merge_rows and .merge_cols
2. Modify these if sentences that check whether f is name from col/rowData or vector. Something like this:
```

if( .is_a_string(f) && f %in% colnames(rowData(x)) ){
    f <- rowData(x)[[ f ]]
}
f <- .norm_f(nrow(x), f)
```
3. Add na.rm to .norm_f function

R/merge.R
```diff
@@ -121,16 +130,24 @@
...
     else if(is.character(f) && length(f)==1 && f %in% colnames(colData(x))){
+        f[is.na(f)] <- "NA"
         f <- factor(as.character(colData(x)[, f]))
-    } else 
+    } else
     {
-        f <- .norm_f(ncol(x), f, "columns")  
+        f <- .norm_f(ncol(x), f, "columns")
     }
```
TuomasBorman: Modify this also similarly to .merge_rows

R/estimateDiversity.R
0xMuluh: it works:)

R/merge.R
0xMuluh: ACK

R/merge.R
0xMuluh: ACK

R/merge.R
0xMuluh: ACK

R/merge.R
0xMuluh: ACK

R/merge.R
```diff
@@ -123,15 +124,11 @@
...
-    if(is.character(f) && length(f)==1 && f %in% colnames(rowData(x))){
-        f <- factor(as.character(rowData(x)[, f]))
-    }
-    else if(is.character(f) && length(f)==1 && f %in% colnames(colData(x))){
-        f <- factor(as.character(colData(x)[, f]))
-    } else 
-    {
-        f <- .norm_f(ncol(x), f, "columns")  
+    if( .is_a_string(f) && f %in% colnames(rowData(x)) ){
+      f <- rowData(x)[[ f ]]
```
TuomasBorman: This should be colData

R/merge.R
```diff
@@ -9,6 +12,10 @@
         stop("'f' must have the same number of ",dim.type," as 'x'",
             call. = FALSE)
     }
+    # This is done otherwise we lose NA values
+    if ( na.rm ) {
```
TuomasBorman: This should be !na.rm --> By making NA values to characters, we are preserving the observations. Otherwise they are lost. --> NA values are not removed --> !na.rm

TuomasBorman: Also, f can be a factor vector, in that case this will not work. Try something like this

if( !na.rm && any(is.na(f)) ){
    f <- as.character(f)
    f[is.na(f)] <- "NA"
}

0xMuluh: ACK

R/merge.R
```diff
@@ -9,6 +12,11 @@
         stop("'f' must have the same number of ",dim.type," as 'x'",
             call. = FALSE)
     }
+    # This is done otherwise we lose NA values
+    if ( !na.rm && any(is.na(f)) ) {
+      f <- as.character(f)
+      f[is.na(f)] <- "NA"
```
TuomasBorman: Indentation should be multiplication of 4 spaces. Run BiocCheck::BiocCheck()

(Also if sentence is not in similar format than others, not big deal but we should follow similar style)

R/merge.R
```diff
@@ -123,15 +125,11 @@
...
     # input check
-    if(is.character(f) && length(f)==1 && f %in% colnames(rowData(x))){
-        f <- factor(as.character(rowData(x)[, f]))
-    }
-    else if(is.character(f) && length(f)==1 && f %in% colnames(colData(x))){
-        f <- factor(as.character(colData(x)[, f]))
-    } else 
-    {
-        f <- .norm_f(ncol(x), f, "columns")  
+    if( .is_a_string(f) && f %in% colnames(rowData(x)) ){
```
TuomasBorman: Should be colnames of colData

TuomasBorman:
Can you run examples of Leo that initiated this https://github.com/microbiome/mia/issues/419 and print the output to here?

TuomasBorman:
Also bump the versions and add NEWS

TuomasBorman (approved):

0xMuluh:
To confirm fix, we will recall the example @antagomir used to raise the issue #419 .
Let's first prepare example data.
```
library(mia)
tse <- GlobalPatterns
tse <- transformAssay(tse, assay.type="counts", method="relabundance")
```

Here we merge features by prevalence, and return the result at the family level.

```
nrow(mergeFeaturesByPrevalence(tse, rank="Family", assay.type="relabundance", detection  = 0.5/100, prevalence  = 20/100))
[1] 21
```
Here we merge features first by Family level grouping, then by prevalence.

```
altExp(tse, "Family") <- mergeFeaturesByRank(tse, rank="Family")
nrow(mergeFeaturesByPrevalence(altExp(tse, "Family"), assay.type="relabundance", detection  = 0.5/100, prevalence  = 20/100))
[1] 21
```

Same happens when we treat Family as a group, rather thank rank:

```
altExp(tse, "Family2") <- mergeFeatures(tse, f="Family")
nrow(mergeFeaturesByPrevalence(altExp(tse, "Family2"), assay.type="relabundance", detection  = 0.5/100, prevalence  = 20/100))
[1] 21
```

Finally, checking the prevalences manually yields the same numbers.

```
sum(rowMeans(assay(altExp(tse, "Family"), "relabundance") > 0.5/100) > 0.2)
[1] 20
```
```
sum(rowMeans(assay(altExp(tse, "Family2"), "relabundance") > 0.5/100) > 0.2)
[1] 20
```

antagomir:
Thanks! 

The last two examples yield a different number (20) - shouldn't it be the same?

Can we have these in unit tests, to make sure it will remain stable also in future releases?

TuomasBorman:
I believe mergeFeaturesByPrevalence() creates a group called others. "others" have all the features under the thresholds. This might explain the behavior.

Can you @Daenarys8 check and create unit tests. They should explain this behavior. For instance if you compare these, you could add comment "other group was removed. It is added by agglomerateByPrevalence function to collect features under threshold." or something like that so that we know in the future what is happening and that it is desired behavior

antagomir:
The code should also show how altExp(tse, "Family") was created. There are different options (onRankOnly etc).

0xMuluh:
```
    tse <- GlobalPatterns
    tse <- transformAssay(tse, assay.type="counts", method="relabundance")
```

Other group not present
```
    altExp(tse, "Family") <- agglomerateByRank(tse, rank="Family")
    altExp(tse, "Family1") <- agglomerateByRank(tse, rank="Family", onRankOnly = TRUE)
    altExp(tse, "Family2") <- agglomerateByRank(tse, rank="Family", onRankOnly = FALSE)
    altExp(tse, "Family3") <- agglomerateByRank(tse, rank="Family", onRankOnly = TRUE, na.rm = TRUE)
    altExp(tse, "Family4") <- agglomerateByRank(tse, rank="Family", onRankOnly = TRUE, na.rm = FALSE)
    altExp(tse, "Family5") <- agglomerateByVariable(tse, f="Family", MARGIN = 'row')
```
In the following, other group is added by agglomerateByPrevalence function to collect features under threshold

```
    actual <- agglomerateByPrevalence(tse, rank="Family", assay.type="relabundance", 
                                         detection  = 0.5/100, prevalence  = 20/100)
    actual0 <- agglomerateByPrevalence(altExp(tse, "Family"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
    actual1 <- agglomerateByPrevalence(altExp(tse, "Family1"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
    actual2 <- agglomerateByPrevalence(altExp(tse, "Family2"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
    actual3 <- agglomerateByPrevalence(altExp(tse, "Family3"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
    actual4 <- agglomerateByPrevalence(altExp(tse, "Family4"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
    actual5 <- agglomerateByPrevalence(altExp(tse, "Family5"), assay.type="relabundance", 
                                       detection  = 0.5/100, prevalence  = 20/100)
```

```
> nrow(actual)
[1] 21
> nrow(actual1)
[1] 21
> nrow(actual2)
[1] 27
```
actual2 create groups based on the full taxonomic hierarchy up to family level while the previous creates the factor group with groups at the family level.

```
> nrow(actual3)
[1] 20
> nrow(actual4)
[1] 21
> nrow(actual5)
[1] 21
> 
```

```
> sum(rowMeans(assay(altExp(tse, "Family"), "relabundance") > 0.5/100) > 0.2)
[1] 20
> sum(rowMeans(assay(altExp(tse, "Family1"), "relabundance") > 0.5/100) > 0.2)
[1] 20
> sum(rowMeans(assay(altExp(tse, "Family2"), "relabundance") > 0.5/100) > 0.2)
[1] 26
> sum(rowMeans(assay(altExp(tse, "Family3"), "relabundance") > 0.5/100) > 0.2)
[1] 19
> sum(rowMeans(assay(altExp(tse, "Family4"), "relabundance") > 0.5/100) > 0.2)
[1] 20
> sum(rowMeans(assay(altExp(tse, "Family5"), "relabundance") > 0.5/100) > 0.2)
[1] 20
> 
```

tests/testthat/test-3agglomerate.R
```diff
@@ -120,6 +120,40 @@ test_that("agglomerate", {
     not_uniq <- mergeFeaturesByRank(se, rank = "Species", make_unique = FALSE)
     expect_true( !any( duplicated(rownames(uniq)) ) )
     expect_true( any( duplicated(rownames(not_uniq)) ) )
+    
+
```
TuomasBorman: One additional line

tests/testthat/test-3agglomerate.R
```diff
@@ -120,6 +120,40 @@ test_that("agglomerate", {
...
+                                       detection  = 0.5/100, prevalence  = 20/100)
+    actual5 <- agglomerateByPrevalence(altExp(tse, "Family5"), assay.type="relabundance", 
+                                       detection  = 0.5/100, prevalence  = 20/100)
+    
+    expect_equal(nrow(actual), nrow(actual0))
+    expect_equal(nrow(actual), nrow(actual1))
+    expect_equal(nrow(actual2), 27)
+    expect_equal(nrow(actual3), 20)
+    expect_equal(nrow(actual), nrow(actual4))
+    expect_equal(nrow(actual), nrow(actual5))
```
TuomasBorman, line 160: Add comments: why they are same, and why there are 20 and 27 featues

tests/testthat/test-3agglomerate.R
0xMuluh: ACK

tests/testthat/test-3agglomerate.R
0xMuluh: ACK

TuomasBorman (approved):

Issue #480, opened 2023-12-21
MARGIN and altexp parameters


TuomasBorman:
Certain functions do not yet support `MARGIN` and `altexp` parameters. `MARGIN` is for specifying rows or columns, and `altexp` is for specifying alternative experiment from `altExp` slot.

We have decided to not limit functions to work only in row or column-wise. Allowing functions to work both row and column-wise might help future maintenance. Additional benefit for this approach is additional generalization; functions work more similarly using general internal functions. Moreover, there could be an application where `colTree` is needed but we do not really support colTrees yet in our functions.

Proposal is the following structure:

```
addValues <- function(tse, ...){
    # Get assay
    mat <- .check_and_get_assay(tse, ...)
    # Calculate some values
    result <- calculate(mat)
    # Store values to colData
    tse <- .add_values_to_colData(tse, result, ...)
}

.check_and_get_assay <- function(tse, MARGIN = 1, altexp = NULL, ...){
    # Do something
}

.add_values_to_colData <- function(tse, values, MARGIN = 2, altexp = NULL, ...){
    # Do something
}
```

I made a draft to branch https://github.com/microbiome/mia/tree/getter_and_setters --> Work with this branch.

See utils.R and estimateDiversity.R --> abundance matrix and trees are now fetched with generic getters from utils.R. Also the result is stored to `colData` with this generic setter. 

Tasks to do:
1. Modify functions so that they fetch and store the values by using these generic functions from utils.R. No need to add visible `MARGIN` and `altexp` parameter, we can add them later if we think they are good to have.
2. Check that the functions work when `MARGIN` and `altexp` is specified.
3. Report any possible problems (there might be in some cases. For instance columns do not have `referenceSeq` slot as rows have. Also, `estimateDIversity(tse, MARGIN = 2, index = "fisher")` is not working but the issue is not in our methods.)

0xMuluh:
I added a commit, some initial feedback would help speed up the changes to be added.

PR #484, opened 2024-01-25
Getter and setters

0xMuluh:
Ping #480

### DIFF
--- a/R/agglomerate.R
+++ b/R/agglomerate.R
@@ -33,10 +33,10 @@
 #'   \code{\link[=merge-methods]{mergeRows}} and
 #'   \code{\link[scuttle:sumCountsAcrossFeatures]{sumCountsAcrossFeatures}}.
 #'   \itemize{
-#'        \item{\code{remove_empty_ranks}}{A single boolean value for selecting 
+#'        \item{\code{remove_empty_ranks}}{A single boolean value for selecting
 #'        whether to remove those columns of rowData that include only NAs after
 #'        agglomeration. (By default: \code{remove_empty_ranks = FALSE})}
-#'        \item{\code{make_unique}}{A single boolean value for selecting 
+#'        \item{\code{make_unique}}{A single boolean value for selecting
 #'        whether to make rownames unique. (By default: \code{make_unique = TRUE})}
 #'    }
 #'
@@ -54,7 +54,7 @@
 #' your results. If no loops exist (loops meaning two higher ranks containing
 #' the same lower rank), the results should be comparable. You can check for
 #' loops using \code{\link[TreeSummarizedExperiment:detectLoop]{detectLoop}}.
-#' 
+#'
 #' Agglomeration sums up the values of assays at the specified taxonomic level. With
 #' certain assays, e.g. those that include binary or negative values, this summing
 #' can produce meaningless values. In those cases, consider performing agglomeration
@@ -79,16 +79,16 @@
 #' ## How many taxa before/after agglomeration?
 #' nrow(GlobalPatterns)
 #' nrow(x1)
-#' 
+#'
 #' # agglomerate the tree as well
 #' x2 <- agglomerateByRank(GlobalPatterns, rank="Family",
 #'                        agglomerateTree = TRUE)
 #' nrow(x2) # same number of rows, but
 #' rowTree(x1) # ... different
 #' rowTree(x2) # ... tree
-#' 
+#'
 #'  # If assay contains binary or negative values, summing might lead to meaningless
-#'  # values, and you will get a warning. In these cases, you might want to do 
+#'  # values, and you will get a warning. In these cases, you might want to do
 #'  # agglomeration again at chosen taxonomic level.
 #'  tse <- transformAssay(GlobalPatterns, method = "pa")
 #'  tse <- agglomerateByRank(tse, rank = "Genus")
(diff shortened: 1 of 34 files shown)


R/agglomerate.R
```diff
@@ -154,7 +154,7 @@ setGeneric("mergeFeaturesByRank",
 #' @export
 setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
     function(x, rank = taxonomyRanks(x)[1], onRankOnly = FALSE, na.rm = FALSE,
-        empty.fields = c(NA, "", " ", "\t", "-", "_"), ...){
+        empty.fields = c(NA, "", " ", "\t", "-", "_"), altexp = NULL, ...){
         # input check
```
TuomasBorman: There should not be reason for adding `altexp` parameter. It should be passed with `...`

R/utils.R
```diff
@@ -1,299 +1,529 @@
...
-.is_numeric_string <- function(x){
-  x <- as.character(x)
-  suppressWarnings({x <- as.numeric(x)})
-  !is.na(x)
-}
-
-.is_function <- function(x){
-  typeof(x) == "closure" && is(x, "function")
-}
-
```
TuomasBorman: Why it seems that all these functions have changed

R/agglomerate.R
TuomasBorman: I will get back to this tomorrow. This is little bit more complicated thing.

R/calculateOverlap.R
```diff
@@ -82,27 +82,25 @@ setMethod("calculateOverlap", signature = c(x = "SummarizedExperiment"),
...
+
         # All the sample pairs
         sample_pairs <- as.matrix(expand.grid(colnames(x), colnames(x)))
-        
+
         # Loop through all sample pairs
         result <- apply(sample_pairs, 1, FUN = function(sample_pair){
           # Get samples
-          sample1 <- assay[ , sample_pair[1]]
-          sample2 <- assay[ , sample_pair[2]]
```
TuomasBorman, line 95: You could just use variable name "assay" so it does not look that everything has changed

R/cluster.R
```diff
@@ -131,13 +131,13 @@ setMethod("cluster", signature = c(x = "SummarizedExperiment"),
     if (.is_non_empty_string(MARGIN)) {
         MARGIN <- tolower(MARGIN)
     }
-    if (length(MARGIN) != 1L 
-        || !(MARGIN %in% c(1, 2, "features", "samples", "columns",
```
TuomasBorman, line 135: There is lots of removed spaces. It is good that there is no empty spaces, but it looks like these all lines have changed, so it is better to not remove those spaces

TuomasBorman (changes requested):
Looks good. Now add also .add_values_to_reducedDim and other .add* methods when the returned value is TreeSUmmairzedExperiment and the value is added to colData/rowData, reducedDim...

Also, there are lots of lines where only spaces are removed. Can you revert those changes? It is really hard to keep up with what has changed now. (I know what is done, but it is super hard for user...)

R/agglomerate.R
TuomasBorman: Because we have to allow MARGIN and altExp. It might be necessary to work with the components of TreeSE instead of the whole TreeSE.

Meaning that:
1. First fetch rowData, tree, assays...
2. agglomerate values
3. Add values back to TreeSE

Or then we can just create another function for MARGIN = 2

TuomasBorman: Currently, we do not have use for agglomerating data column-wise but there might be in the future...

TuomasBorman:
So to summarize: it is better to modify only those lines where it is relevant. You can do another PR for removing additional spaces if we see that necessary

R/agglomerate.R
0xMuluh: ACK

R/utils.R
0xMuluh: Oh its because to run GHA, I put commits against `mia:master`. So it includes all the earlier commits you pushed in `getter_and_setters`

TuomasBorman: These lines have not changed in my branch. Would it be possible if you make changes directly to mia (without using fork). I can give you an access if you have not already

R/agglomerate.R
TuomasBorman: I can make a suggestion of this. I have to try different approaches and let's discuss about this after that

R/utils.R
0xMuluh: okay. I don't have access.

R/agglomerate.R
0xMuluh: okay sure.

TuomasBorman:
Hello!

Sorry that it took so long. I made changes to mia:getters_and_setters branch. Please see those. They are related to agglomerateByRank function. Now, MARGIN is supported meaning that we can utilize this function also to columns (altexp was already working...).

```
devtools::load_all()

data("GlobalPatterns")

tse <- GlobalPatterns

# Agglomerate to all ranks
altExps(tse) <- splitByRanks(tse)

# Agglomerate altExp
altExp(tse, "test") <- tse[1:10, ]
tse_altExp <- agglomerateByRank(tse, rank = "Kingdom", altexp = "test")
tse_altExp

# Create ranks for columns
colData(tse)[["Kingdom"]] <- sample(c("king1", "king2", "king3"), size = ncol(tse), replace = TRUE)
colData(tse)[["Class"]] <- sample(c("class1"), size = ncol(tse), replace = TRUE)
colData(tse)[["Species"]] <- sample(c("species1", "species2", "species3", "species4"), size = ncol(tse), replace = TRUE)
colData(tse)

# Agglomerate columns
tse_col <- agglomerateByRank(tse, rank = "Kingdom", MARGIN = 2)
tse_col

tse_col <- agglomerateByRank(tse, rank = "Class", MARGIN = 2)
colData(tse_col)
```

Can you give feedback on that? Is there anything that could break? Is it too complex solution / is it easy to follow what happens? If everything is good, I think we can proceed again.

antagomir:
@Daenarys8 any chance to resolve the conflicts and finalize this one?

0xMuluh:
ACK. Will push an update asap.

antagomir:
This is old PR, still open. What's the status of this one @Daenarys8 ?

TuomasBorman:
This might be something that we should check in the future, not now. This would be good addition, but requires lots of work and I am not sure if it is worth to do at least now when we have other tasks

antagomir:
Ok. I was just worried about having too many active open PRs.

TuomasBorman:
Closing for now

PR #487, opened 2024-02-10
agglomerateByRank: agglomerate tree fix

TuomasBorman:
Hi,

I noticed that tree was not agglomerated correctly in `agglomeratedByRanks` function. Instead of agglomeration of an existing tree, new tree was calculated. This is now fixed.

-Tuomas


antagomir (commented):
Looks good!

Issue #488, opened 2024-02-12
Clarify `pseudocounts` argument in `transformAssay`


artur-sannikov:
The documentation for the `pseudocounts` argument of the `transformAssay` function is not clear. Saying "should the minimum value of assay.type be added to assay values"  implies that the user should add the minimum value themselves. However, there is an automatic calculation when `pseudocounts = TRUE`.

Can be rewritten as: 

> TRUE, FALSE, or a numeric value. When TRUE, automatically adds the minimum value of  `assay.type`. When FALSE, does not add any pseudocount. Alternatively, a specified numeric value is added as pseudocount.

antagomir:
Hi @artur-sannikov - in fact, would you even like to try & make a pull request on this? 

Would be great but if not, noprob & we will handle with some delay. Many thanks for the suggestion. 

I would slightly modify the text as follows:

"TRUE, FALSE, or a numeric value. When TRUE, automatically adds the minimum positive value of assay.type. When FALSE, does not add any pseudocount (pseudocount = 0). A specified numeric value is added as pseudocount."

artur-sannikov:
Hi @antagomir, I'll open a pull request. I think this can be the final version:

> TRUE, FALSE, or a numeric value. When TRUE, automatically adds the minimum positive value of assay.type. When FALSE, does not add any pseudocount (pseudocount = 0). Alternatively, a user-specified numeric value can be added as pseudocount.

The last sentence tells the user to add their own number if they want.

PR #490, opened 2024-02-13
Clarify pseudocount argument

artur-sannikov:
Make `pseudocount` argument of `transformAssay` documentation clearer. #488


antagomir (approved):
Thanks!

PR #496, opened 2024-02-27
Deprecate taxonomyTree and addTaxonomyTree

theotimeee:
Replace `taxonomyTree` with `getHierarchyTree` and `addTaxonomyTree` with `addTree`. 

Create new `deprecate` R script and manual page.

Create new `get-values` manual page.


R/taxonomy.R
```diff
@@ -382,15 +379,32 @@ setMethod("getTaxonomyLabels", signature = c(x = "SummarizedExperiment"),
...
+#' 
+#' \code{getHierarchyTree} generates a hierarchy tree from the available
+#'   taxonomic information. Internally it uses
+#'   \code{\link[TreeSummarizedExperiment:toTree]{toTree}} and
+#'   \code{\link[TreeSummarizedExperiment:resolveLoop]{resolveLoop}} to sanitize
+#'   data if needed.
+#' 
+#' @inheritParams taxonomy-methods
+#' 
+#' @name get-values
```
TuomasBorman: The name should be getHiearchyTree

R/taxonomy.R
```diff
@@ -425,17 +439,17 @@ setMethod("taxonomyTree", signature = c(x = "SummarizedExperiment"),
 )
 
 #' @rdname taxonomy-methods
-setGeneric("addTaxonomyTree",
+setGeneric("addTree",
```
TuomasBorman: This name should be addHierarchyTree

NEWS
```diff
@@ -97,4 +97,5 @@ Changes in version 1.9.x
 Changes in version 1.11.x
 + loadFromMetaphlan: support strain rank
 + agglomerateByRank: agglomerate tree fix
++ Deprecate taxonomyTree, addTaxonomyTree
```
TuomasBorman: Say what are the new names --> maybe that replaced functions x and y with y and z

TuomasBorman (changes requested):
Remember to run BiocCheck

R/taxonomy.R
```diff
@@ -415,41 +440,41 @@ setMethod("taxonomyTree", signature = c(x = "SummarizedExperiment"),
             if(any(td_NA[,i])){
                 to_drop <- paste0(colnames(td)[i],":",td[,i][td_NA[,i]])
                 tree <- ape::drop.tip(tree,
-                                      to_drop,
```
TuomasBorman: Add ape::drop.tip with importFrom tag (check other functions)

R/deprecate.R
```diff
@@ -0,0 +1,40 @@
...
+            signature = "x",
+            function(x, ...)
+                standardGeneric("addTaxonomyTree"))
+
+#' @rdname deprecate
+#' @export
+setMethod("addTaxonomyTree", signature = c(x = "SummarizedExperiment"),
+            function(x){
+                .Deprecated(msg = paste0("'addTaxonomyTree' is deprecated.\n",
+                                        "Use 'addHierarchyTree' instead."))
```
TuomasBorman: Do not add new line ("\n"). It is tidier without (because there might be multiple warnings

R/taxonomy.R
```diff
@@ -372,31 +369,59 @@ setMethod("getTaxonomyLabels", signature = c(x = "SummarizedExperiment"),
...
+#' 
+#' @name getHierarchyTree
+#' 
+#' @examples
+#' data(GlobalPatterns)
+#' GlobalPatterns
+#' getHierarchyTree(GlobalPatterns)
 
-#' @rdname taxonomy-methods
+#' @rdname get-values
```
TuomasBorman: This name should be hierarchy-tree. Also add the same name above addHierarchyTree -->

1. There should not be get-values page.
2. There should be hiearachy-tree page
3. The title of hiarchy-tree page should be something like ("calculate hireachy tree)
4. *HiearchyTree functions should be removed from taxonomy-methods page.


See man directory. There is manual pages. Check the corresponding pages that they are correct. (In RStudio, you can click the file and then preview.)

TuomasBorman (changes requested):
Almost there

TuomasBorman (approved):
I think this is good to go.

@antagomir

antagomir:
Some final things, sorry...

Is there a risk that `addTree()` gets mixed with other types of trees? Should it be `addHierarchyTree()`, or is it more generic so that you can use addTree for different things?

It is essential to update the OMA as well, or at least have a clear scheduled plan how that is done.

TuomasBorman:
Actually it is now `addHiearchyTree`. It is just for adding calculated tree (not generic function for adding any kind of trees)

Updating OMA and other packages is taken into account. We have planned with @thpralas that names are changed everywhere immediately when PR in mia is merged

antagomir:
Ah great! Then good to go I guess.

PR #502, opened 2024-03-05
Rename cluster to addCluster

theotimeee:
Rename `cluster` to `addCluster.` 
Fix indentation in calculateDMM and cluster files following Bioconductor best practices (indentation must be a multiple of 4 spaces).
Move deprecated functions from the calculateDMM file to deprecate file.


R/cluster.R
```diff
@@ -112,3 +113,49 @@ setMethod("cluster", signature = c(x = "SummarizedExperiment"),
...
+    if (name %in% names(metadata(x))) {
+        stop("The 'name' must not exist in the metadata of the object.", call. = FALSE)
+    }
+}
+
+.check_data_name <- function(x, clust.col, MARGIN) {
+    if (MARGIN == 1) {
+        if (clust.col %in% names(rowData(x))) {
+            stop("The 'clust.col' parameter must not exist in the names of the rowData of the object.", 
+                call. = FALSE)
```
TuomasBorman: Sorry, I made a commit just couple hours ago where these functions were removed. Sync with master.

TuomasBorman (changes requested):

R/deprecate.R
```diff
@@ -1,12 +1,52 @@
...
+#' @param clust.col A single character value indicating the name of the 
+#'   \code{rowData} (or \code{colData}) where the data will be stored.
+#'   
+#' @param exprs_values a single \code{character} value for specifying which
+#'   assay to use for calculation.
+#'   (Please use \code{assay.type} instead.)
+#'   
+#' @param transposed Logical scalar, is x transposed with cells in rows?
+#' 
+#' @param type the type of measure used for the goodness of fit. One of
```
TuomasBorman: This manual page should collect all deprecated functions. Some of them might have same parameter name but different meaning. So I'm wondering what happens then. --> I'm wondering if it is better to not list parameters for any of these functions since user's should not use them anyways (even though BIocCheck would tell otherwise)

R/deprecate.R
```diff
@@ -1,12 +1,52 @@
...
+#' @rdname deprecate
+#' @export
+#' @importFrom bluster clusterRows
+setMethod("cluster", signature = c(x = "SummarizedExperiment"),
+            function(x, BLUSPARAM, assay.type = assay_name, 
+                    assay_name = "counts", MARGIN = "features", full = FALSE, 
+                    name = "clusters", clust.col = "clusters", ...){
+                .Deprecated(msg = paste0("'cluster' is deprecated. ",
+                                        "Use 'addCluster' instead."))
+                addCluster(x)
```
TuomasBorman: This does not pass parameters. addCluster gets only x parameter. Not others. Moreover, you can skip listing parameters. Use ... and then pass them to addCluster with ...


Only x is needed to list since the method is chosen based on class of x

TuomasBorman: The idea is to keep these functions a hollow as possible. Just that they work but do not do anything but call the new function

TuomasBorman (changes requested):

R/calculateDMM.R
```diff
@@ -141,16 +141,16 @@ setMethod("calculateDMN", signature = c(x = "ANY"), .calculate_DMN)
...
+                   transposed = FALSE, ...){
+              .Deprecated(old="calculateDMN", new="cluster", 
+                          "Now calculateDMN is deprecated. Use cluster with DMMParam parameter instead.")
+              mat <- assay(x, assay.type)
+              if(!transposed){
+                  mat <- t(mat)
+              }
+              calculateDMN(mat, ...)
+          }
 )
```
TuomasBorman: It seems that the indentation is incorrect. Also lines are too long (80 characters is maximum)

Can you either fix these or revert the file? Maybe remove the changes, since this PR is about cluster function

Otherwise, this is ready to go. Thanks!

TuomasBorman (approved):
Looks good! @antagomir

antagomir:
There are unresolved comments, and one conflict. If those are resolved then this seems good to me at least?

TuomasBorman (approved):

PR #503, opened 2024-03-06
Initialise mediation functions

RiboRings:
This PR is not ready yet, but it was opened to begin discussion around mediation analysis for mia. Here, two new main functions are introduced: mediateColData and mediateAssay, both located in [R/mediate.R](https://github.com/microbiome/mia/blob/mediation/R/mediate.R). They depend on the mediation package and if this is not ideal these functions could be placed in an extension of mia.

mediateColData provides an easy way to run mediation analysis between 3 variables in the colData, while mediateAssay allows to run multiple mediation models for every taxon in an assay, or for every component of a reducedDim, and subsequently adjust significance for multiple comparison. They both leverage the mediation::mediate function under the hood.

mediateColData takes a tse object, three colData variables for outcome, treatment and mediator, respectively, plus other covariates from the colData and other arguments supported by mediation::mediate. It returns the output of the mediation model (similar to the output of lm or glm). Below is a self-contained example of how it works:

```
library(mia)
library(microbiomeDataSets)
library(mediation)
library(dplyr)
source("R/mediate.R")

tse <- OKeefeDSData()

tse <- transformAssay(tse,
                      method = "relabundance")

tse <- estimateDiversity(tse,
                         index = "shannon",
                         assay.type = "relabundance")

colData(tse)$bmi_group <- as.numeric(tse$bmi_group)

med_out <- mediateColData(tse,
                          outcome = "bmi_group",
                          treatment = "nationality",
                          mediator = "shannon",
                          covariates = "timepoint.within.group",
                          boot = TRUE, sims = 1000)

summary(med_out)

# Causal Mediation Analysis 
#
# Nonparametric Bootstrap Confidence Intervals with the Percentile Method
#
# (Inference Conditional on the Covariate Values Specified in `covariates')
#
#                Estimate 95% CI Lower 95% CI Upper p-value    
# ACME             0.1086       0.0344         0.19   0.004 ** 
# ADE             -0.4580      -0.6726        -0.24  <2e-16 ***
# Total Effect    -0.3493      -0.5484        -0.13  <2e-16 ***
# Prop. Mediated  -0.3110      -0.9781        -0.08   0.004 ** 
# ---
# Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
#
# Sample Size Used: 222 
#
#
# Simulations: 1000 

plot(med_out)
```

Screenshot 2024-03-06 at 18 15 50

mediateAssay iterates over the same code used for mediateColData. It takes the extra arguments assay.type and dim.type. If one of them is correctly specified, it runs mediation::mediate for every row of an assay or every column of a reducedDim slot, respectively. As output, it returns a dataframe with effect sizes and p-values for every taxon/component. Below is an example that requires variables from the previous example above:

```
tse <- transformAssay(tse,
                      method = "clr",
                      pseudocount = 1)

tse <- tse[ , tse$timepoint.within.group == 2]

med_res <- mediateAssay(tse,
                        outcome = "bmi_group",
                        treatment = "nationality",
                        assay.type = "clr",
                        boot = TRUE, sims = 300)

med_res %>%
  filter(ACME_adjpval > 0.05) %>%
  head() %>%
  knitr::kable()
```

Treatment | Mediator | Outcome | ACME_estimate | ADE_estimate | ACME_pval | ADE_pval | ACME_adjpval | ADE_adjpval | ACME_CI_lower | ADE_CI_lower
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
nationality | Uncultured Mollicutes | bmi_group | -0.1700134 | -0.1986718 | 0.0066667 | 0.2133333 | 0.2888889 | 0.2133333 | -0.0469092 | 0.1062497
nationality | Eubacterium ventriosum et rel. | bmi_group | -0.1087084 | -0.2599768 | 0.0200000 | 0.0866667 | 0.4814815 | 0.0901333 | -0.0100311 | 0.0607181
nationality | Eubacterium cylindroides et rel. | bmi_group | 0.1010478 | -0.4697330 | 0.0200000 | 0.0133333 | 0.4814815 | 0.0279570 | 0.2347390 | -0.1674882
nationality | Lactobacillus salivarius et rel. | bmi_group | 0.0816559 | -0.4503411 | 0.0266667 | 0.0066667 | 0.4814815 | 0.0222222 | 0.1957793 | -0.1609730
nationality | Bacteroides uniformis et rel. | bmi_group | 0.2371214 | -0.6058066 | 0.0266667 | 0.0133333 | 0.4814815 | 0.0279570 | 0.4490582 | -0.2087500
nationality | Uncultured Selenomonadaceae | bmi_group | 0.1084270 | -0.4771122 | 0.0333333 | 0.0133333 | 0.4814815 | 0.0279570 | 0.2127692 | -0.1225544
nationality | Parabacteroides distasonis et rel. | bmi_group | 0.2731733 | -0.6418585 | 0.0333333 | 0.0000000 | 0.4814815 | 0.0000000 | 0.5877062 | -0.2688401
nationality | Catenibacterium mitsuokai et rel. | bmi_group | -0.0810081 | -0.2876771 | 0.0466667 | 0.0800000 | 0.5515152 | 0.0845528 | -0.0009472 | 0.0257356

More examples can be found in [this repository](https://github.com/RiboRings/mediation/tree/main)


antagomir:
Great.

1. Input types: splitting functionality by colData and assay is one option. Another option is to split by univariate vs. multivariate. In such case you could do something like mediate(tse, outcome = ..., treatment = "nationality", assay.type = "clr", ...) and there the "outcome" could refer to colData variable, _or_ to one assay feature (taxon). This would be recognized automatically. It is important to have a chance to run mediation also for colData but then the question is, what is the added value compared to just running the basic mediate::mediation function from the original package on just a single variable from colData.. And why not allow situations where multiple comparisons could be run for colData.. (e.g. if there are different derived indices in addition to Shannon diversity).. This is a more general design aspect of mia and might deserve some thought. On the other hand perhaps we could just go with this colData / assay split first and improve later. 

2. Output: @TuomasBorman might comment but have been wondering if these functions should directly return augmented TreeSE objects as output. BUt I am not sure if it is feasible here. At least the examples should show how to add relevant pats of the results to the TreeSE object 

3. Support for custom mediation functions. If possible, it would be idea to allow users to input also custom mediation functions. This will greately facilitate later innovation and comparisons. In practice, we should define how the inputs and ouputs of such function should look like and as long as the function follows those standards it could be anything. For instance, we have allow the "FUN" argument to insert entirely new dissimilarity measures in functions such as runNMDS: https://microbiome.github.io/mia/reference/runNMDS.html 

4. We need an argument to choose from different methods, even if this would first provide only one method. Some other methods might like to take the entire assay rather than run for loop over features.

TuomasBorman:
Looks nice!

1. I think it is better to have one function for one task if possible. For example, now we have 2 functions for cross-correlation which might just confuse user.

In this case one options would be `getMediate(tse, outcome = "disease", mediator = "counts", treatment = "diet", outcome.type = NULL, mediator.type = NULL, treatment.type = NULL)`, where `outcome`,` mediator `and `treatment` parameters are automatically fetched from `colData`/`assay`/`reducedDim`. -> give error if overlapping name --> user can manually specify with *type where it is searched

2. Yes, it could be added to `metadata` but I don't really see additional value. However, the function name could be `getMediate` to highlight that it does not return `TreeSE`.  (`add*` would return)

RiboRings:
Thank you for the comments!

@TuomasBorman, if I understand correctly, you are suggesting that mediateColData and mediateAssay should be combined into a single function, right? It sounds like it would simplify the code, and make it possible to use single or multiple variables for any of the outcome, mediator and treatment arguments.

But then how should we standardise the output? Because when multiple analyses are run (mediateAssay), the most practical output is a data.frame of model statistics for every combination, whereas for a single analysis (mediateColData), the most informative output is the trained model itself (so you can do `summary` or `plot` on it). From this aspect, keeping the two functions separate might be more convenient.

RiboRings:
@antagomir, so you would consider this implementation with mediation::mediate like one option out of many, right? And then other mediation functions such as the hdmed ones could be specified with `FUN = ...`? This would definitely make this function more generic and complete.

In this case, how would you control the suitability of the custom method? For example, the hdmed (high-dimensional mediation) funs are only appropriate for high-dimensional data like the assay (mediateAssay), but not for colData variables (mediateColData). Also for this aspect having two separate functions might be more convenient(?)

Or actually what do you mean by custom mediation functions?

antagomir:
Well in fact also the arguments may be very different for different methods. That's why there is for instance runMDS, runNMDS, etc. and not a single runOrdination..  so perhaps it is not good to go very generic for now. Let's just finalize this for the mediation package.

Regarding comment from TB above: it is recommened in general that the function output should always be the same, regardless of the inputs. We could have that, and in addition the user could choose to include the resulting model/s in metadata slot (default FALSE but examples could show how to do this when necessary). This applies also to the assay case, it would be a list of output models.

Would that work? 

Default output could indeed be a data frame with effect sizes and (adjusted) p-values.

RiboRings:
Hi!

The `getMediation` function is ready for review. I still need to add more argument checks, comment code and add unit testing, but the main parts are there. Now there is a single function that can take mediators from the colData (`mediator` arg), the assays (`assay.type`) or reducedDims (`dim.type`). It can also return the original model in `attr(med_df, "metadata")[[model_number]]` when `add.metadata = TRUE`.

Caveats:

- It is probably not ideal to allow the user to input multiple outcomes and treatments, because the outcomes might have different distributions and the treatments different control/treatment conditions, which preferably should be specified as the `family`, `control.value` and `treat.value` arguments.
- Adding the option to select specific taxa/reduced dimensions as mediators would make the function more complicated. I think the user could simply subset the (Tree)SE object or specify number of dimensions when running ordination, and then feeding that object to getMediation.

R/mediate.R
```diff
@@ -0,0 +1,386 @@
...
+  
+  # Split CI lists into lower and upper limits and add them to dataframe
+  med_df[["ACME_CI_lower"]] <- results[["ACME_ci"]][names(results[["ACME_ci"]]) == "2.5%"]
+  med_df[["ACME_CI_upper"]] <- results[["ACME_ci"]][names(results[["ACME_ci"]]) == "97.5%"]
+  med_df[["ADE_CI_lower"]] <- results[["ADE_ci"]][names(results[["ADE_ci"]]) == "2.5%"]
+  med_df[["ADE_CI_upper"]] <- results[["ADE_ci"]][names(results[["ADE_ci"]]) == "97.5%"]
+  
+  if (add.metadata) {
+    # If desired, the models for every mediator are saved into the metadata attribute
+    attr(med_df, "metadata") <- results[["Model"]]
```
TuomasBorman: Use .add_values_to_metadata

R/mediate.R
```diff
@@ -0,0 +1,386 @@
...
+#' @rdname getMediation
+#' @export
+setGeneric("getMediation", signature = c("se"),
+           function(se, ...) standardGeneric("getMediation"))
+
+#' @rdname getMediation
+#' @export
+setMethod("getMediation", signature = c(se = "SummarizedExperiment"),
+      function(se, outcome, treatment,
+               mediator = NULL, assay.type = NULL, dim.type = NULL,
```
TuomasBorman: Other functions are using x parameter for TreeSE so that should be followed

R/mediate.R
```diff
@@ -0,0 +1,386 @@
...
+        i <- 0
+            
+        for (mediator in mediators) {
+             
+          # Update index 
+          i <- i + 1
+          
+          if (message) {
+            print(paste0("Mediator ", i, " out of ", length(mediators), ": ", mediator)) 
+          }
```
TuomasBorman: Use verbose parameter instead of message.

Also do not print, use message() instead. (User cannot stop printing, but messages can be hindered). --> Then you do not need to use paste0() either when using message()

R/mediate.R
```diff
@@ -0,0 +1,386 @@
...
+      if (any(sapply(kwargs[c("control.value", "treat.value")], is.null))) {
+        stop("Too many treatment levels. Consider specifing a treat.value and a control.value", call. = FALSE)
+      # but if they are specified
+      } else {
+        # and they also appear in the treatment variable
+        if (!all(kwargs[c("control.value", "treat.value")] %in% unique(df[[treatment]]))) {
+          stop("treat.value and/or control.value not found in the levels of the treatment variable.", call. = FALSE)
+        }
+        
+        # Find indices of samples that belong to either control or treatment
```
TuomasBorman: Can you still run BioCheck::BiocCheck() and fix everthing related to this function? I seems that line length and indentation is wrong.

R/mediate.R
```diff
@@ -0,0 +1,386 @@
...
+            rownames(mat) <- paste0(dim.type, seq(1, nrow(mat)))
+            # Use reducedDim for analysis
+            mediators <- rownames(mat)
+          }
+          
+        } else {
+          # Throw error if none or multiple mediation options are specified
+          stop("The arguments mediator, assay.type and dim.type are mutually exclusive",
+               ", but ", sum(med_opts), " were provided.", call. = FALSE)
+        }
```
TuomasBorman: I think it is more clear if you move this up and remove the code above from if statement

if( sum(med_opts) != 1){
    stop()
}

R/mediate.R
TuomasBorman: Also, this prints multiple rows which might be annoyig if console is full of these lines. You might want to consider to check mergeSEs function. It has similar loop but it updates the row so that only one row is printed.

TuomasBorman (changes requested):
I still have to run the code, but these are what I found by just looking

R/mediate.R
RiboRings: I tried to run `BiocCheck()` but it threw this error:
```
─ BiocCheckVersion: 1.38.2
─ BiocVersion: 3.18
─ Package: mia
─ PackageVersion: 1.11.8
─ sourceDir: /Users/giuliobene3000/Desktop/Projects/mia
─ installDir: /var/folders/5y/2glwvl6j3yv459vzzlfh7j8h0000gn/T//RtmpJxZMgN/file16b91db8fe15
─ BiocCheckDir: /Users/giuliobene3000/Desktop/Projects/mia.BiocCheck
─ platform: unix
─ isTarBall: FALSE

* Installing package...
* Checking for deprecated package usage...
adding rname 'https://bioconductor.org/checkResults/3.18/bioc-LATEST/meat-index.dcf'
Warning in .sql_add_resource(x, rname, rtype, fpath, ext, fname) :
  restarting interrupted promise evaluation
Warning in .sql_add_resource(x, rname, rtype, fpath, ext, fname) :
  internal error -3 in R_decompress1
Warning in value[[3L]](cond) :
  
trying to add rname 'https://bioconductor.org/checkResults/3.18/bioc-LATEST/meat-index.dcf' produced error:
  lazy-load database '/Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/library/curl/R/curl.rdb' is corrupt
Error in bfcrpath(bfc, rnames = url, exact = TRUE, download = TRUE, rtype = "web") : 
  not all 'rnames' found or unique.
```

RiboRings: I checked that function and indentation looks correct. Could you share the error you get so I can try to pinpoint the problem?

R/mediate.R
RiboRings: I moved it up to the first part of the conditional. However, I don't think there's a way to completely remove the if else statement, since input is different for assay, reducedDims and colData.

R/mediate.R
RiboRings: It looks to me that .add_values_to_metadata takes a (Tree)SE object as input. However, here the models are added to the output dataframe, not the original se object.

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#'   (default: \code{dim.type = NULL})
+#'
+#' @param family A specification for the outcome model link function.
+#'    (default: \code{family = gaussian("identity")})
+#' 
+#' @param covariates Name or list of names of colData variables used as covariates
+#'   in the model. (default: \code{covariates = NULL})
+#' 
+#' @param p.adj.method A single character value for selecting adjustment method
+#'   of p-values. Passed to `p.adjust` function. (default: \code{p.adj.method = "BH"})
```
antagomir: default in p.adjust function is "holm" I think (can be checked). I think we should not change the default if the argument is passed. Instead we can provide our own recommendation in the example section and show how to use.

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#' 
+#' @param add.metadata TRUE or FALSE, should the model metadata be returned.
```
antagomir, line 36: Also mention how it is returned; added to metadata slot of the SE object?

antagomir: Show in examples how that can be then retrieved from the object (if not yet shown)

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#'   \item{ADE_estimate}{the Average Direct Effect (ADE) estimate}
+#'   \item{ACME_pval}{the p-value for the ACME estimate}
+#'   \item{ADE_pval}{the p-value for the ADE estimate}
+#'   \item{ACME_adjpval}{the adjusted p-value for the ACME estimate}
+#'   \item{ADE_adjpval}{the adjusted p-value for the ADE estimate}
+#'   \item{ACME_CI_lower}{the lower limit of the confidence interval for the ACME estimate}
+#'   \item{ACME_CI_upper}{the upper limit of the confidence interval for the ACME estimate}
+#'   \item{ADE_CI_lower}{the lower limit of the confidence interval for the ADE estimate}
+#'   \item{ADE_CI_upper}{the upper limit of the confidence interval for the ADE estimate}
+#' }
```
antagomir: Just thinking if all of these are needed. We could simplify this and only provide the ones that are commonly used. Then the users can have all the rest in metadata, for the more dedicated users?

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#' 
+#' # Analyse mediated effect of nationality on BMI through NMDS components        
+#' med_df <- getMediation(tse,
+#'                        outcome = "bmi_group",
+#'                        treatment = "nationality",
+#'                        dim.type = "MDS",
+#'                        covariates = c("sex", "age"),
+#'                        treat.value = "Scandinavia",
+#'                        control.value = "CentralEurope",
+#'                        boot = TRUE, sims = 500)
```
antagomir: I think that the the "sims" argument is "niter" or "n.iter" in many packages. We should aim to harmonize argument names. I would go with n.iter - can we change? However if this comes directly from the mediation package, then we should keep the same and just pass the argument.

R/mediate.R
antagomir: Indentation should be multiples of 4 according to Bioc guidelines. Is it o now?

antagomir: The rnames error I am not familiar with. Any progress on that?

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+        }
+      }
+    }
+  }
+  return(x)
+}
+
+# Run mediation analysis
+#' @importFrom mediation mediate
+run.mediate <- function(x, outcome, treatment,
```
antagomir: If this is internal function then you could start it with period, so it would be .run.mediate

antagomir: same for other internals

R/mediate.R
antagomir: Do we have a reason not to add them to (Tree)SE? Is it necessary to return them to some dataframe instead of TreeSE metadataslot - or is this some intermediate step?

antagomir (commented):
See the review comments. Overall, looks nice!

Is there any useful way to visualize mediation results? For the examples.

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#' @examples
+#' # Import libraries
+#' library(mia)
+#' library(microbiomeDataSets)
+#' library(scater)
+#' 
+#' # Load dataset
+#' tse <- LahtiWAData()
+#'  
+#' # Convert BMI variable to numeric
```
TuomasBorman: Is there any other dataset that we can use? This loads the data and requires internet access. --> if there is no internet access, we get an error even though the code is working.

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+    
+    if (verbose) {
+      message(diff, " samples removed because of missing data.")
+    }
+  }
+  
+  # If the treatment variable has three or more levels
+  if (!is.numeric(df[[treatment]]) & length(unique((df[[treatment]]))) > 2) {
+    # and boot is TRUE
+    if (!is.null(kwargs[["boot"]])) {
```
TuomasBorman: Add this boot check inside the if statement of line 258. Try to avoid nested ifs

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+    # Remove missing data from se
+    x <- x[ , rownames(df)]
+    
+    if (verbose) {
+      message(diff, " samples removed because of missing data.")
+    }
+  }
+  
+  # If the treatment variable has three or more levels
+  if (!is.numeric(df[[treatment]]) & length(unique((df[[treatment]]))) > 2) {
```
TuomasBorman: Use && instead of &

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+    # and boot is TRUE
+    if (!is.null(kwargs[["boot"]])) {
+      # and control and treat value are not specified
+      if (any(sapply(kwargs[c("control.value", "treat.value")], is.null))) {
+        stop("Too many treatment levels. Consider specifing a treat.value and a control.value", call. = FALSE)
+      # but if they are specified
+      } else {
+        # and they also appear in the treatment variable
```
TuomasBorman: Can you drop this else? Either the run is stopped because of if or then it continues, so else is not doing much here. Also it makes code little bit harder to read (because of nested ifs)


ALso: this comment here seems to be incorrect / little bit unclear. "control and treat value must be in treatment variable"

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+        
+        # Check that arguments can be passed to mediate and remove unused samples
+        x <- check.mediate.args(x, outcome, treatment,
+                                mediator, covariates, verbose, ...)
+        
+        # Check which mediator was provided (colData, assay or reducedDim)
+        med_opts <- sapply(
+          list(mediator = mediator, assay.type = assay.type, dim.type = dim.type),
+          function(x) !is.null(x)
+        )
```
TuomasBorman: Replace sapply with vapply or lapply. (Bioconductor recommendations)

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+        
+        if (sum(med_opts) != 1) {
+          # Throw error if none or multiple mediation options are specified
+          stop("The arguments mediator, assay.type and dim.type are mutually exclusive",
+               ", but ", sum(med_opts), " were provided.", call. = FALSE)
+        
+        } else {
```
TuomasBorman: You can also remove this else. The run stop if the statement above is TRUE. This else does not give additional value but makes code little bit harder to read because of nested ifs

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#' @param treatment The name of the colData variable used as treatment in the model.
+#'
+#' @param mediator The name of the colData variable used as mediator in the model.
+#'   (default: \code{mediator = NULL})
+#'
+#' @param assay.type The name of the assay used for feature-wise mediation analysis.
+#'   (default: \code{assay.type = NULL})
+#' 
+#' @param dim.type The name of the reducedDim used for component-wise mediation analysis.
+#'   (default: \code{dim.type = NULL})
```
TuomasBorman: Is dimred used in other functions? For example plotReducedDim?

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+  if (!is.null(covariates)) {
+    for (covariate in covariates) {
+      # Fetch covariate from colData and store it in dataframe
+      df[[covariate]] <- colData(x)[[covariate]]
+      
+      # Add covariate to formula of mediation model
+      relation_m <- paste(relation_m, "+", covariate)
+      # Add covariate to formula of outcome model
+      relation_dv <- paste(relation_dv, "+", covariate)
+    }
```
TuomasBorman: Try to avoid loops

```
cbind(df, colData(x)[, covariates, drop = FALSE])

paste(relation_m, "+", paste(covariates, collapse = " + "))

paste(relation_mv, "+", paste(covariates, collapse = " + "))
```

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+  
+  return(results)
+}
+
+
+# Combine results into output dataframe
+make.output <- function(results, p.adj.method, add.metadata) {
+  
+  # Create dataframe with model variables, effect sizes and p-values
+  med_df <- do.call(data.frame, results[1:(length(results) - 3)])
```
TuomasBorman: Use seq_len instead of 1: --> BIoc recommendations

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+        # Set initial index  
+        i <- 0
+            
+        for (mediator in mediators) {
+             
+          # Update index 
+          i <- i + 1
+          
+          if (verbose) {
+            message("\rMediator ", i, " out of ", length(mediators), ": ", mediator, appendLF = FALSE)
```
TuomasBorman: It seems that because mediate gives messages, this is not working as expected and new line is created for every mediator. But is is fine

R/mediate.R
```diff
@@ -0,0 +1,394 @@
...
+#'
+#' # Analyse mediated effect of nationality on BMI through clr-transformed features     
+#' med_df <- getMediation(tse,
+#'                        outcome = "bmi_group",
+#'                        treatment = "nationality",
+#'                        assay.type = "clr",
+#'                        covariates = c("sex", "age"),
+#'                        treat.value = "Scandinavia",
+#'                        control.value = "CentralEurope",
+#'                        boot = TRUE, sims = 300)
```
TuomasBorman: There are 130 features which takes long time. You could agglomerate the data to phylum level for instance to speed up

TuomasBorman (changes requested):
I checked the code now in detail, seems good and working

R/mediate.R
TuomasBorman: There is some problem with installation of packages. This might be related to mismatch between Bioc packages. You could try to install R devel 4.4 (which is the version that should be developed against)

R/mediate.R
TuomasBorman: 1.

My mistake. I looked the code now better. The returned value is now df.


2.

We could return data.frame in `metadata(tse)`

cons
-  The resulting df is not linked anywhere i.e., the columns are not linked with `colnames(tse)` or `rownames(tse)`
- User only needs the df and fetching it from `metadata(tse)` might feel like additional step

pros
+ We have other functions that add the results to `metadata(tse)` (like `cluster` adds detailed clustering info)
+ Adding it to `metadata(tse)` is "correct" way to use the data container
+ Calculating might take long time and by storing it to `metadata(tse)` keeps it safe there

As this is new function, we can decide the output type freely. There could be `as.df` option, but I think results would be better to store in `metadata(tse)` by default

RiboRings: What if we have a function called `getMediation`, which returns the dataframe object, and a function called `runMediation`, which returns the TreeSE with df as metadata?

R/mediate.R
RiboRings: Yes, default of p.adjust is "holm". I saw "fdr"/"BH" is the default of `testExperimentCrossCorrelation`, so I used it. So should we use "holm"?

R/mediate.R
RiboRings: I think we can skip the columns Treatment and Outcome, because they are constant, and the last 4 that describe the confidence intervals, so we are left with 7 columns out of 13.

R/mediate.R
RiboRings: sims is an argument of the original mediate function, so we can keep it

R/mediate.R
RiboRings: I updated indentation. The BiocCheck works for me on R devel 4.4

R/mediate.R
TuomasBorman: I think that would be clear. However, runMediation should be called addMediation

R/mediate.R
antagomir: Ah right. It would be good for us to be consistent within mia at least. 

In my opinion it would be most clean to do as I proposed. We can have "BH"/"fdr" in all examples. But default can be invisible the user, then it is confusing if it changes the default value of p.adjust while we at the same time say that the function calls that. So yes, my suggestion is to change it also in testExperiment.. function.

antagomir: Btw isn't it it nowadays the real name `testExperimentCrossAssociation`?

R/mediate.R
antagomir: Also "ADE_pval" and "ACME_pval" are unnecessary imo, users should only look at adjpval

R/mediate.R
antagomir: Right. I agree, this is why we have used data sets from `data()` command in all examples. 

In fact, `LahtiWAData()` is available this way as:

```
library(miaTime)
data(hitchip1006)
```

I originally thought that the idea was to use O'Keefe diet swap data (`dietswap()`)? 

Which one works better for a demo?

antagomir: The O'Keefe data, if it is useful, is dietary intervention study. It might be a better example as intervention studies are standard targets for mediation analyses. Anyway, both will do in principle.

R/mediate.R
antagomir: Yes, in examples it is better to do so. But also could write in the example that it is done for the sake of the example to speed up. So that users do not robotically copypaste aggregation part and think that it is somehow necessary part of this.

antagomir:
@RiboRings if you can close the conversations one by one when they have been resolved that would be great, to keep track on what was solved and what not

antagomir:
Up - @RiboRings

RiboRings:
Hi! I finally got back to this

R/mediate.R
RiboRings: Now it is imported as suggested by @antagomir

In general, I used hitchip1006 because it comes with pre-calculated alpha diversity, it has more than one confounder so that addition of multiple confounders can be demonstrated (c("sex", "age")), it has more than 2 levels for the treatment variable so the treat.value and control.value functionality can be shown, and unlike okeefe it does not need filtering of only one timepoint

But if okeefe might have some advantage, I can also try that out. Is there a way to load okeefe with `data`?

R/mediate.R
RiboRings: Thanks! Now it is agglomerated to family level (22 taxa) and the number of permutations was reduced to 100. I commented that we use 100 for speed but around 1000 permutations is recommended.

DESCRIPTION
```diff
@@ -80,6 +81,7 @@ Suggests:
     reldist,
     ade4,
     microbiomeDataSets,
+    miaTime,
```
TuomasBorman: Is there reason for adding miaTime? This might cause a problem since packages are looked from Bioc or Cran. miaTime is only in Github so it cannot be installed here

(If miaTime is needed, there might be option for installing it with remotes package, but we have to check it)

R/mediate.R
```diff
@@ -0,0 +1,416 @@
...
+        # Combine results into dataframe
+        med_df <- .make.output(results, p.adj.method, add.metadata)
+        return(med_df)
+    }
+)
+
+
+# Check that arguments can be passed to mediate and remove unused samples
+.check.mediate.args <- function(x, outcome, treatment, mediator, covariates, verbose = TRUE, ...) {
+
```
TuomasBorman, line 302: Add input check for verbose

TuomasBorman: Never mind, it is already checked in main function

R/mediate.R
antagomir: Great. If it is still slow, you can just pick the top few most prevalent taxa and reduce to 50 permutations if it still works ok for a demo.

R/mediate.R
antagomir: Both are ok for a demo I think.  

I would say that an intervention study, like OKeefe, are more standard for mediation analyses generally speaking but I can see the reasoning.

DESCRIPTION
antagomir: I guess this is because of the demo data sets that were needed. 

We can reconsider demo data sets from mia or miaViz if this helps..?

RiboRings: Yes I added miaTime because the hitchip1006 data is loaded from that package. Are there better alternatives?

TuomasBorman: I think there is Remotes also available in DESCRIPTION

Remotes:
Github:microbiome/miaTme

or something like that --> can you check

TuomasBorman: Also it is better if you use data from mia package for instance. It is only demonstration for the functionality. We should avoid adding dependencies

RiboRings: Ok! I did some extensive research for a good dataset, but the ones in mia were not ideal and did not give any significant results. I'll think a little bit.

antagomir: We could consider moving hitchip data to mia if it helps. It is not critical to have in miaTime.

RiboRings: For now I could add miaTime to Remotes in DESCRIPTION. Or should we directly add hitchip to mia?

antagomir: What would be @TuomasBorman preference? I don't care too much which pkg hitchip data is in.

TuomasBorman: You can add miaTime to Remotes for now. I have heard that mia might be problematic to install because of all the dependencies. In the future, we have to check which dependencies can be dropped, and also what we do with this

TuomasBorman:
Looks good, I will go through this in detail by tomorrow. I ran BiocCheck. There are at least multiple lines that were over 80 characters long.

I think code that we add should follow as closely as possible Bioconductor recommendations (it is harder to fix the code in the future). --> Can you run BiocCheck and fix the issues related to the lines that are added with this PR?

RiboRings:
Ok, thank you for all the guidance!

I shortened the lines. Just as a reference, I also got this error when running BiocCheck:

```
* Checking for remote package usage...
    * ERROR: Package dependencies must be on CRAN or Bioconductor. Remove
      'Remotes:' from DESCRIPTION
```

antagomir:
Hmm - - perhaps we should move hitchip to mia then..

TuomasBorman:
There are also SilvermanAGutData from miaTime that is used in unit tests. At some point, we are planning to submit miaTime to Bioconductor so this is temporal thing.

Adding miaTime to Remotes in DESCRIPTION means that miaTime is installed when mia is built. This might cause problems. So it might be better to remove it from there.

Options
1. Move hitchip data to mia --> We have to then move it back to miaTime when it is in Bioconductor.
2. Run examples only when miaTime is installed --> This means that examples are not run (because miaTime is not installed by default)
3. In addition to step 2, add miaTime installation step in the maediate examples. --> miaTime is installed only when documentation is built (not when user installs the package, as far as I know). However, there should be no installation steps in the documentation.

 Maybe the option 3 is the best?

TuomasBorman (approved):
Top quality!

I forgot to mention that update NEWS file, It should include all the changes in easy, human-readable format. --> Added getMediation and addMediation functions.

antagomir:
All options ok to me. 

hitchip can move permanently to mia if this helps, there is no particular need to have it in miaTime. It was put there because we needed to limit the number of data sets in mia package and because hitchip had some time series in it

RiboRings:
Maybe we could comment out the examples and remove miaTime from DESCRIPTION, and then uncomment the examples when miaTime is submitted to BioConductor?

TuomasBorman:
> Maybe we could comment out the examples and remove miaTime from DESCRIPTION, and then uncomment the examples when miaTime is submitted to BioConductor?

That was actually same to where I ended up after I thought this over night. --> So let's do that.

Instead of commenting out do:

```
if( require("miaTime") ){
examples
}
```

miaTime is used some other places also, but they are not affected. They are already ran only if miaTime is installed. (vignette is not ran)

antagomir:
a shining idea

RiboRings:
R CMD check fails due to warning:

```
❯ checking for unstated dependencies in examples ... WARNING
  'library' or 'require' call not declared from: ‘miaTime’
```

Also, for some reason miaTime is being installed even when it is not in DESCRIPTION:

```
miaTime                    0.1.21     2024-04-17 [1] Github (microbiome/miaTime@9fe9[771](https://github.com/microbiome/mia/actions/runs/8753409443/job/24023056975#step:4:791))
```

antagomir:
For me running the latest version of the "mediation" branch through checks throws warning

> Warning in data(hitchip1006) : data set ‘hitchip1006’ not found
> Error: object 'hitchip1006' not found

Also

* checking tests ...
  Running ‘testthat.R’^[[5~^[[5;5~^[[5;5~^[[6;5~
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
---
  [9]  2.5067 -  2.5100 == -0.00331
  ...
  Backtrace:
      ▆
   1. └─mia (local) testTransformations(tse) at test-5transformCounts.R:331:5
   2.   └─testthat::expect_equal(actual, compare) at test-5transformCounts.R:293:9

TuomasBorman:
I added \donttest which should fix the issue of @RiboRings. 

@antagomir I cannot reproduce your error. It runs fine in my local machine, and the same tests are also run in GHA. Although, it is good to investigate this further to get answer on why you get different results

TuomasBorman:
Did not run, I disabled the examples

antagomir:
If this goes through automated tests then I think should be ok

RiboRings:
Ubuntu:
```
* checking whether package ‘mia’ can be installed ... [30s/30s] WARNING
Found the following significant warnings:
  Warning: package ‘matrixStats’ was built under R version 4.5.0
  Warning: package ‘GenomicRanges’ was built under R version 4.5.0
  Warning: package ‘Biobase’ was built under R version 4.5.0
  Warning: package ‘SingleCellExperiment’ was built under R version 4.5.0
```

Mac:
```
Error: Error: Failed to install 'mia' from local:
  unable to load shared object '/Users/runner/work/_temp/Library/cli/libs/cli.so':
  dlopen(/Users/runner/work/_temp/Library/cli/libs/cli.so, 0x0006): tried: '/Users/runner/work/_temp/Library/cli/libs/cli.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e' or 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/work/_temp/Library/cli/libs/cli.so' (no such file), '/Users/runner/work/_temp/Library/cli/libs/cli.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e' or 'arm64'))
```

TuomasBorman:
That will be hopefully fixed soon when new Bioc is released. However R 4.5.0 is little bit odd since next Bioc devel is developed against 4.4.0. That is something to do with rworkflows, I think

RiboRings:
Hi @TuomasBorman! Should I do anything to fix this? How should we proceed to merge this PR?

TuomasBorman:
I haven't been able to fix the problem, however, everything should be ok if at least one of these runs work and there are no errors related to this PR. --> so we can merge without passing all these rules

(I asked from rworkflows maintainers if they know what is the problem, and they have not answered yet)

TuomasBorman:
Seems ok, I will merge (I just updated NEWS). Thanks!

PR #505, opened 2024-03-14
rename importers to standardize "import" terminology

theotimeee:
rename following importers : 

- loadFromBiom --> importBIOM

- loadFromQIIME2 --> importQIIME2

- readQZA --> importQZA

- loadfromMothur --> importMothur

- loadFromMetaphlan --> importMetaphlan

- loadFromHumann --> importHumann


TuomasBorman (changes requested):
SImilarly to other functions, add the olds function to deprecate.R

Also, I am thinking if importHumann and importMetaphlan should be importHUMAnN and importMetaPhlAn @antagomir 

I think we should follow their naming convention even though it is harder to remember the correct naming. It is more clear that the function imports the HUMAnN data

TuomasBorman (commented):
Have run devtools::document()

It seems that yu have changed the name of importHUMAnN document but the file name is not updated in man folder

NEWS
```diff
@@ -104,4 +104,9 @@ Changes in version 1.11.x
 + cluster: Overwrite old results instead of failing
 + getPrevalence: bugfix, if assay contains NA values, it does not end up to NA anymore.
 + getExperimentCrossCorrelation fix: enable using of sampleMap in MAE.
+<<<<<<< HEAD
++ rename importers loadFromBiom, loadFromQIIME2, readQZA, loadFromMothur, 
+loadFromMetaphlan, loadFromHumann
+=======
 + Implemented the setTaxonomyRanks function to specify which ranks are recognized as taxonomy ranks.
+>>>>>>> 32d85c6bfe311c9160a72bf718d5f36987bdc261
```
TuomasBorman: Not merged correctly

"+ Implemented..." comes from the devel branch. It should be first. Then there should be your additions.  Also those additonal git marks should be removed -->
```

+ rename importers loadFromBiom, loadFromQIIME2, readQZA, loadFromMothur, 
loadFromMetaphlan, loadFromHumann
```

TuomasBorman (approved):
@antagomir

man/deprecate.Rd
```diff
@@ -21,6 +27,18 @@ addTaxonomyTree(x, ...)
 taxonomyTree(x, ...)
 
 \S4method{taxonomyTree}{SummarizedExperiment}(x, ...)
+
+loadFromBiom(...)
+
+loadFromQIIME2(...)
+
+loadFromQZA(...)
```
TuomasBorman: This should be readQZA

TuomasBorman (approved):
Change loadFromQZA to readQZA, then good to go

TuomasBorman (approved):
@antagomir

NEWS
antagomir: is this resolved?

man/deprecate.Rd
antagomir: close when resolved

antagomir (approved):

antagomir:
Ok to me when the above points are resolved.

PR #506, opened 2024-03-20
create new generic converter

theotimeee:
Create a new generic function `convert` to replace the following converters : 

- makeTreeSEFromPhyloseq
- makeTreeSEFromBiom 
- makeTreeSEFromDADA2 
- makePhyloseqFromTreeSE


R/convert.R
```diff
@@ -0,0 +1,182 @@
...
+#'   (By default: \code{assay.type = "counts"})
+#'   (x class : SummarizedExperiment)
+#'   
+#' @param assay_name a single \code{character} value for specifying which
+#'   assay to use for calculation.
+#'   (Please use \code{assay.type} instead. At some point \code{assay_name}
+#'   will be disabled.)
+#'   (x class : SummarizedExperiment)
+#'   
+#' @param tree_name a single \code{character} value for specifying which
```
TuomasBorman: SHould be tree.name

R/convert.R
```diff
@@ -0,0 +1,182 @@
...
+#' - Coerce a \code{TreeSummarizedExperiment} object to a \code{phyloseq} object
+#' 
+#' The appropriate conversion is automatically selected based on the class of 
+#' the input data.
+#'
+#' @return An object of class TreeSummarizedExperiment or Phyloseq.
+#'
+#' @importFrom Biostrings DNAStringSet
+#' @importFrom S4Vectors SimpleList DataFrame make_zero_col_DFrame
+#' @importFrom SummarizedExperiment colData colData<-
```
TuomasBorman: Add these just above the function. Easier to find and see which function requires them

R/convert.R
```diff
@@ -0,0 +1,182 @@
...
+            }
+)
+
+#' @rdname convert
+#' @export
+setMethod("convert", signature = c(x = "TreeSummarizedExperiment"),
+            function(x, tree_name = "phylo", ...){
+                .make_phyloseq_from_TreeSE(x, tree_name, ...)
+            }
+)
```
TuomasBorman: These all other functions are clear:

- We want to convert X to TreeSE

However, in this case there mgiht be more options in the future. (Now we have only TreeSE --> phylyseq)

--> Add to paramater.

- The default of it is "phyloseq".
- Check that the value is a single character value and that it equals to supported types, i.e., "phyloseq"
- Use switch to get the correct function based on to parameter

TuomasBorman (changes requested):
When dada2, phyloseq and biom ackages are not loaded, the functions says that these classes cannot be found --> I think that is ok. (just to note)

R/makeTreeSummarizedExperimentFromBiom.R
```diff
@@ -1,85 +1,14 @@
...
-#' (default \code{remove.artifacts = FALSE})
-#' 
-#' @param ... additional arguments 
-#'   \itemize{
-#'        \item{\code{patter}}{\code{character} value specifying artifacts
-#'        to be removed. If \code{patterns = "auto"}, special characters
-#'        are removed. (default: \code{pattern = "auto"})}
-#'    }
-#' 
-#' @return An object of class
```
TuomasBorman: This documentation was little bit unclear (reflecting also the naming). But loadFromBiom is importer. So it cannot be added ton convert documentation.

Can you return the documentation that is related to loadFromBiom (I know that we have another PR open that renames importers)


The convert documentation is good

TuomasBorman: loadFromBiom imports data from BIOM file. convert converts the class type

NAMESPACE
```diff
@@ -49,11 +51,7 @@ export(loadFromHumann)
 export(loadFromMetaphlan)
 export(loadFromMothur)
 export(loadFromQIIME2)
-export(makePhyloseqFromTreeSE)
 export(makePhyloseqFromTreeSummarizedExperiment)
-export(makeTreeSEFromBiom)
-export(makeTreeSEFromDADA2)
-export(makeTreeSEFromPhyloseq)
```
TuomasBorman: Also these functions should still be available (although they should be in deprecate.R file)

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -1,219 +1,135 @@
...
-            # Adds to the list
-            args[["sample_data"]] <- sample_data
-        }
+    # Creates a phyloseq object
+    phyloseq <- do.call(phyloseq::phyloseq, args)
+    return(phyloseq)
+}
+
+
+.make_phyloseq_from_TreeSE <- function(x, tree.name = "phylo", ...){
```
TuomasBorman: For users still using old name, modify this line to:

.make_phyloseq_from_TreeSE <- function(x, tree.name = tree_name, tree_name = "phylo", ...){

R/convert.R
```diff
@@ -0,0 +1,200 @@
...
+#' (x class : biom)
+#' 
+#' @param remove.artifacts \code{TRUE} or \code{FALSE}: If file have
+#' some taxonomic character naming artifacts, should they be removed.
+#' (default \code{remove.artifacts = FALSE}) 
+#' (x class : biom)
+#' 
+#' @param ... additional arguments 
+#'   \itemize{
+#'        \item{\code{patter}}{\code{character} value specifying artifacts
```
TuomasBorman: patter --> typo

TuomasBorman:
With these, I think good to go

NAMESPACE
```diff
@@ -152,6 +157,11 @@ exportMethods(isNotContaminant)
 exportMethods(left_join)
 exportMethods(makePhyloseqFromTreeSE)
 exportMethods(makePhyloseqFromTreeSummarizedExperiment)
+exportMethods(makeTreeSEFromBiom)
+exportMethods(makeTreeSEFromDADA2)
+exportMethods(makeTreeSEFromPhyloseq)
+exportMethods(makeTreeSummarizedExperimentFromBiom)
+exportMethods(makeTreeSummarizedExperimentFromDADA2)
```
TuomasBorman: Why there are these new functions? Only cluster should be added.

Did you check that the functions in deprecate.R matches with the ones that are removed from "main" files?

(The reason seems to be that those functions was not exported. If that is the reason, users have not used these functions --> we can just remove them without deprecation)

R/makephyloseqFromTreeSummarizedExperiment.R
```diff
@@ -1,219 +1,136 @@
...
-            sample_data <- phyloseq::sample_data(sample_data)
-            # Adds to the list
-            args[["sample_data"]] <- sample_data
-        }
+    # Creates a phyloseq object
+    phyloseq <- do.call(phyloseq::phyloseq, args)
+    return(phyloseq)
+}
+
+
```
TuomasBorman: Remove additional line

TuomasBorman:
Works nicely. @antagomir

TuomasBorman:
> When dada2, phyloseq and biom ackages are not loaded, the functions says that these classes cannot be found --> I think that is ok. (just to note)

I am still wondering this. It is not good that these warning messages occur when the package is loaded. Can you try to figure out fix for this?

One possibility is to add packages to dependencies but we don't want to do that. Other option is to make non-generic fuctions (like convertPhyloseq and convertToPhyloseq), but this generic function is much more tidier solution

theotimeee:
> > When dada2, phyloseq and biom ackages are not loaded, the functions says that these classes cannot be found --> I think that is ok. (just to note)
> 
> I am still wondering this. It is not good that these warning messages occur when the package is loaded. Can you try to figure out fix for this?
> 
> One possibility is to add packages to dependencies but we don't want to do that. Other option is to make non-generic fuctions (like convertPhyloseq and convertToPhyloseq), but this generic function is much more tidier solution

I added import tags for these classes above each method and the namespace what modified accordingly. I think the warning messages do not occur anymore when loading the package

theotimeee:
> > > When dada2, phyloseq and biom ackages are not loaded, the functions says that these classes cannot be found --> I think that is ok. (just to note)
> > 
> > 
> > I am still wondering this. It is not good that these warning messages occur when the package is loaded. Can you try to figure out fix for this?
> > One possibility is to add packages to dependencies but we don't want to do that. Other option is to make non-generic fuctions (like convertPhyloseq and convertToPhyloseq), but this generic function is much more tidier solution
> 
> I added import tags for these classes above each method and the namespace what modified accordingly. I think the warning messages do not occur anymore when loading the package

Actually this creates dependencies errors with R CMD check, I am going to look for another solution

TuomasBorman:
https://github.com/microbiome/OMA/pull/484

TuomasBorman:
Any updates on this? Should we go with generic function or specific functions for each type

convertFromPhyloseq/convertToPhyloseq

antagomir:
I like the idea of generic function if it can be reasonably implemented.

It would be good to get this case closed soon in any case.

TuomasBorman:
It works nicely but the problem is that the package loading gives warning. This is caused because the class in generic. For example phyloseq class is not detected if phyloseq package is not loaded, and mia does not load it by default.

antagomir:
Ah right.. is there any way to deal with this automagically somehow..

theotimeee:
I have removed the import tags and it seems like the warnings do not appear anymore. However, there is a new R CMD check note : `Warning: <anonymous>: ... may be used in an incorrect context: ‘convert(...)’`
I am going to look into this and try to fix it.

antagomir:
Might be that "convert" is too generic name, reserved for other purposes?

TuomasBorman:
If you cannot find solution for this, we can also rename the existing functions and move the documentation under shared "convert" page.

-  makeTreeSEFromPhyloseq --> convertFromPhyloseq
-  makeTreeSEFromBiom --> convertFromBIOM
-  makeTreeSEFromDADA2 --> convertFromDADA2
-  makePhyloseqFromTreeSE --> convertToPhyloseq

theotimeee:
> If you cannot find solution for this, we can also rename the existing functions and move the documentation under shared "convert" page.
> 
> * makeTreeSEFromPhyloseq --> convertFromPhyloseq
> * makeTreeSEFromBiom --> convertFromBIOM
> * makeTreeSEFromDADA2 --> convertFromDADA2
> * makePhyloseqFromTreeSE --> convertToPhyloseq

Should I do that even though it's a note and not a warning? I'm still trying to understand the note better and determine if it's something to be concerned about.

TuomasBorman:
This is what happens

```
> devtools::load_all()
ℹ Loading mia
in method for ‘convert’ with signature ‘x="dada"’: no definition for class “dada”
in method for ‘convert’ with signature ‘x="phyloseq"’: no definition for class “phyloseq”
in method for ‘convert’ with signature ‘x="biom"’: no definition for class “biom”
> library(dada2)
Loading required package: Rcpp
> devtools::load_all()
ℹ Loading mia
in method for ‘convert’ with signature ‘x="phyloseq"’: no definition for class “phyloseq”
in method for ‘convert’ with signature ‘x="biom"’: no definition for class “biom”
```

So it it caused, because dada2:dada class is not found. But when dada2 package is loaded, it is found. We do not want to add new dependencies, that is why those packages are not loaded when mia is loaded.

Previously these functions were implemented as "basic" functions, so they were not this kind of generic functions that choose method based on class. 

In theory, you could probably add a script to mia.R that loads these packages if available, but that is suboptimal solution. I think it is not good that there are this kind of notes when mia is loaded. Some users might think that these are warnings.

TuomasBorman:
Any thoughts @antagomir

antagomir:
If we go with mia then I am OK with either solution. One generic function would be nice but if this is not possible, then specific functions.

Another solution could be a separate utility package. We might have other misc utility needs too. But that will be more maintenance overhead.

TuomasBorman:
\+ convert() is very neat solution
\- We cannot have more dependencies, we have too many already
\- We cannot have notes, when package is loaded. It does not look good.

--> I think we have to go with separate functions. You can just rename the functions from their original files and then just combine the documentation under convert

NEWS
```diff
@@ -135,6 +135,9 @@ Changes in version 1.13.x
   includes the newly added subsampled assay.
 + Fix bug in mergeFeaturesByPrevalence
 + new aliases calculateDPCoA to getDPCoA, calculateNMDS to getNMDS, calculateRDA to getRDA, 
-calculateCCA to getCCA
+  calculateCCA to getCCA
++ add informative error message in rarefyAssay on assays with strictly-negative values
```
TuomasBorman, line 139: Problem with news, same line is added

R/makeTreeSummarizedExperimentFromPhyloseq.R
```diff
@@ -1,90 +0,0 @@
...
-    } else {
-        rowTree <- NULL
-    }
-    if (!is.null(obj@refseq)) {
-        referenceSeq <- obj@refseq
-    } else {
-        referenceSeq <- NULL
-    }
-    TreeSummarizedExperiment(assays = assays,
-                            rowData = rowData,
```
TuomasBorman: Not sure, if the link is just broken between files, but it seems now that you have copy.-pasted code to another file. Instead, it should only look like that the name has changed.

Also, as this is not anymore big change (only the name is changed and documentation is moved under common page), I am wondering if you should make new PR with new branch if that is easier.

We have parameter naming PR that will be soon merged, and it affects also these files. It was good that we explored the possibility of common convert function even though it will not be implemented.

I am wondering what is the easiest approach.

TuomasBorman (changes requested):

R/makeTreeSummarizedExperimentFromPhyloseq.R
theotimeee: Yes I think it may be easier to open a new PR because I have probably made some unnecessary changes in this PR when I tried to implement `convert` generic function.

theotimeee: Should I make new methods for the functions that do not have one (e.g. `makeTreeSEFromDADA2`) ?

Also should we keep both names makeTreeSE* and makeTreeSummarizedExperiment* for every converters or not?

TuomasBorman: Now this bigger PR was merged https://github.com/microbiome/mia/pull/582, so you can continue with this.

Those names will be deprecated. All the functions that were available for user must be deprecated and added to deprecate.R file.

Each functionality will have only one function that is convertTo* or convertFrom*

R/convertFromBIOM.R
```diff
@@ -69,31 +84,30 @@
...
+#' @rdname convert
+#' @export
+convertFromBIOM <- function(
+        obj, prefix.rm = removeTaxaPrefixes, 
         removeTaxaPrefixes = FALSE, rank.from.prefix = rankFromPrefix, 
         rankFromPrefix = FALSE,
         artifact.rm = remove.artifacts, remove.artifacts = FALSE, ...){
     # input check
     .require_package("biomformat")
-    if(!is(x,"biom")){
```
TuomasBorman, line 95: The name of object is changed from obj to x indicating that your branch is out of sync

R/convertToPhyloseq.R
```diff
@@ -0,0 +1,264 @@
...
+      
+      # Check assay.type
+      .check_assay_present(assay.type, x)
+      
+      # phyloseq object requires nonduplicated rownames. If there are 
+      # duplicated rownames, they are converted so that they are unique
+      if( any(duplicated(rownames(x))) ){
+        rownames(x) <- getTaxonomyLabels(x)
+      }
+      # List of arguments
```
TuomasBorman, line 88: These look like they are new functions even though they are not. The name has only changed. This cannot be merged since this breaks the history of the functions. One option is to create new branch and continue from there

TuomasBorman (changes requested):

R/convertToPhyloseq.R
theotimeee: OK I think I am going to create a new branch then, it will be easier

TuomasBorman:
https://github.com/microbiome/mia/pull/591

Issue #407, opened 2023-07-27
agglomerateByPrevalence manpage lacks explanation


antagomir:
Looking at `?agglomerateByPrevalence` does not sufficiently explain what this function is intended to do.

Suggested todos:
- provide a more complete description at the roxygen manpage
- consider using similar function grouping also for other function families in the mia package (if not already)

PR #509, opened 2024-04-08
Add explanation for agglomerateByPrevalence

theotimeee:
This PR aims to solve the issue (https://github.com/microbiome/mia/issues/407) by making the function `agglomerateByPrevalence` easier to understand : 

- Specify the returned value
- Add details on the way the function works
- Clarify the examples


TuomasBorman:
Very nice. I think we could still move it under agglomerate-methods with agglomerateByRank and agglomerateByVariable

I think this is more "agglomeration" than "prevalence" function, so that would be the best place for it

R/getPrevalence.R
```diff
@@ -594,23 +585,22 @@ setMethod("getPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
...
+#' @rdname agglomerate-methods
+#' 
+#' @details
+#' \code{agglomerateByPrevalence} sums up the values of assays at the taxonomic 
+#' level specified by \code{rank} (by default the highest taxonomic level 
+#' available) and selects the summed results that exceed the given population 
+#' prevalence at the given detection level. The other summed values (below the 
+#' threshold) are agglomerated in an additional row taking the name indicated by
+#' \code{other_label} (by default "Other").
+#'
```
TuomasBorman: Is it possible to add examples here instead of agglomerate.R? I am not sure if it works like that. Are the examples then included in agglomerate.methods.Rd along with examples from agglomerate.R?

R/agglomerate.R
```diff
@@ -49,6 +49,9 @@
 #'   experiments be removed prior to agglomeration? This prevents to many
 #'   nested alternative experiments by default (default:
 #'   \code{strip_altexp = TRUE})
+#'   
+#' @param other_label A single \code{character} valued used as the label for the
+#'   summary of non-prevalent taxa. (default: \code{other_label = "Other"})
```
TuomasBorman: This could be added to getPrevalence.R above agglomerateByPrevalence function if possible.

R/agglomerate.R
```diff
@@ -62,8 +65,11 @@
 #' can produce meaningless values. In those cases, consider performing agglomeration
 #' first, and then applying the transformation afterwards.
 #'
-#' @return A taxonomically-agglomerated, optionally-pruned object of the same
-#'   class as \code{x}.
+#' @return 
+#' \code{agglomerateByRank} returns a taxonomically-agglomerated, 
+#'   optionally-pruned object of the same class as \code{x}.
+#' \code{agglomerateByPrevalence} returns a taxonomically-agglomerated object
+#' of the same class as x and based on prevalent taxonomic results.
```
TuomasBorman: Also these things related to agglomerateByPrevalence could be moved if possible

R/getPrevalence.R
theotimeee: I moved it here and it looks like it works well.

TuomasBorman:
If you check the man page of agglomerate-methods, it does not describe detection and prevalence and as_relative parameters. (They are passed through with ...). You could add to agglomerate.R file those parameters under ... parameter.

" Passed to getPrevalence and getPrevalentTaxa functions. Used in agglomeratebyPrev function..." (Along with the current text.

As agglomerate.R already has ... parameter, you cannot ... above agglomerateByPrevalence function (otherwise it overwrites the one in agglomerate.R).

TuomasBorman:
See the man page. You could modify description so that it describes the agglomeration functions in more general (it now only says about agglomerateByRank) --> Agglomeration functions can be used to sum-up to rank or variable or prevalence...

NAMESPACE
```diff
@@ -50,7 +50,6 @@ export(importQIIME2)
 export(importQZA)
 export(inner_join)
 export(left_join)
-export(loadFromBiom)
```
TuomasBorman: Always check. Now loadFromBiom is removed. That is because you forgot to add

```
#' @rdname deprecate
#' @export

```
above loadFromBiom in deprecate.R

TuomasBorman (approved):

PR #513, opened 2024-04-15
Rename splitByRanks to agglomerateByRanks and add new option as.list

theotimeee:
This PR renames `splitByRanks` to `agglomerateByRanks` and add a new argument `as.list`.

- If `as.list = TRUE` : The `SimpleList` of agglomerated `SummarizedExperiment` objects is returned by `agglomerateByRanks`
- If `as.list = FALSE` : The `SimpleList` of `SummarizedExperiment` objects is added to `altExps` of the `SummarizedExperiment` passed as a parameter


R/splitByRanks.R
```diff
@@ -121,28 +115,48 @@ setGeneric("splitByRanks",
...
+        if( !.is_a_bool(as.list) ){
+            stop("'as.list' must be TRUE or FALSE.", call. = FALSE)
+        }
+        if( as.list ){
+            .split_by_ranks(x, ranks, args)
+        }
+        else{
+            altExps(x) <- .split_by_ranks(x, ranks, args)
+            x
+        }
```
TuomasBorman: SummarizedExpriment does not have altExps slot. You could convert it to TreeSE and give warning.

Also: you are replacing existing altExps with new list (old alternative experiments are lost). Just add these new ones to the list

R/splitByRanks.R
```diff
@@ -121,28 +115,48 @@ setGeneric("splitByRanks",
...
 )
 
-#' @rdname splitByRanks
+#' @rdname agglomerate-methods
 #' @export
-setMethod("splitByRanks", signature = c(x = "TreeSummarizedExperiment"),
+setMethod("agglomerateByRanks", signature = c(x = "TreeSummarizedExperiment"),
           function(x, ranks = taxonomyRanks(x), na.rm = TRUE, ...){
               callNextMethod()
           }
```
TuomasBorman: I think method for TreeSE is not needed because TreeSE is SCE (and that function is called)

TreeSE --> SCE (does have altExp) --> SE

R/splitByRanks.R
```diff
@@ -121,28 +115,48 @@ setGeneric("splitByRanks",
...
 )
 
-#' @rdname splitByRanks
+#' @rdname agglomerate-methods
 #' @export
-setMethod("splitByRanks", signature = c(x = "SingleCellExperiment"),
-          function(x, ranks = taxonomyRanks(x), na.rm = TRUE, ...){
+setMethod("agglomerateByRanks", signature = c(x = "SingleCellExperiment"),
+          function(x, ranks = taxonomyRanks(x), na.rm = TRUE, as.list = TRUE,
+                    ...){
```
TuomasBorman: as.list = FALSE is better option (recommended way). In deprecate.R, you can have agglomerateByRanks(as.list = FALSE) hard-coded inside splitByRanks function since the function did not have that parameter

R/splitByRanks.R
```diff
@@ -121,37 +116,53 @@ setGeneric("splitByRanks",
...
 }
 
-#' @rdname splitByRanks
+#' @rdname agglomerate-methods
 #' @export
-setMethod("splitByRanks", signature = c(x = "SummarizedExperiment"),
-    function(x, ranks = taxonomyRanks(x), na.rm = TRUE, ...){
+setMethod("agglomerateByRanks", signature = c(x = "SummarizedExperiment"),
+    function(x, ranks = taxonomyRanks(x), na.rm = TRUE, as.list = FALSE, ...){
+        browser()
```
TuomasBorman: browser()

--> Have you run tests? They should notify you about this. (devtools::check() or BiocCheck)

R/deprecate.R
```diff
@@ -379,7 +406,7 @@ setMethod("full_join", signature = c(x = "ANY"),
             }
 )
 
-#' @rdname deprecate
+#' @rdname deprecate
```
TuomasBorman: Can you revert this line

R/deprecate.R
```diff
@@ -362,6 +362,33 @@ loadFromHumann <- function(...) {
...
+#' @export 
+setMethod("splitByRanks", signature = c(x = "SingleCellExperiment"),
+          function(x,...){
+              .Deprecated(msg = paste0("'splitByRanks' is deprecated. ",
+                                       "Use 'agglomerateByRanks' instead."))
+              agglomerateByRanks(x, as.list = FALSE,...)
+          }
+)
+
```
TuomasBorman: After considering this, you could leave splitByRanks as an alias.

1. SingleCellExperiment has fnction splitAltExps which is simialr to this
2. We have still unsplitByRanks

--> Add splitByRanks function back, but like it is here: just empty function calling agglomerateByRanks. (Also, only SummarizedExperiment method is needed since SCE is SE)

R/splitByRanks.R
```diff
@@ -55,42 +63,29 @@
...
-#' altExps(GlobalPatterns)
-#' altExp(GlobalPatterns,"Kingdom")
-#' altExp(GlobalPatterns,"Species")
+#' # agglomerateByRanks
+#' # 
+#' x <- agglomerateByRanks(GlobalPatterns)
+#' altExps(x)
+#' altExp(x,"Kingdom")
+#' altExp(x,"Species")
 #'
```
TuomasBorman: x could be tse

R/splitByRanks.R
```diff
@@ -55,42 +63,29 @@
...
-#' x <- unsplitByRanks(GlobalPatterns)
+#' x <- unsplitByRanks(x)
 #' x
-NULL
-
-################################################################################
-# splitByRanks
-
-#' @rdname splitByRanks
+#'
```
TuomasBorman, line 83: Add NULL

R/splitByRanks.R
```diff
@@ -121,37 +116,53 @@ setGeneric("splitByRanks",
...
+        if( !.is_a_bool(as.list) ){
+            stop("'as.list' must be TRUE or FALSE.", call. = FALSE)
+        }
+        if( as.list ){
+            .split_by_ranks(x, ranks, args)
+        }
+        else{
+            altExps(x) <- .split_by_ranks(x, ranks, args)
+            x
+        }
```
TuomasBorman: Store result to variable and end the function with return(x) --> tidier

TuomasBorman: Also do this:

x <- .split_by_ranks(x, ranks, args)
if( !as.list ){
    # Add to altexps
}
return(x)

TuomasBorman: That is simpler

TuomasBorman: Moreover, you are overwriting existing altExps which is problem. Instead, append the altExps list

R/splitByRanks.R
```diff
@@ -121,37 +116,53 @@ setGeneric("splitByRanks",
...
+                }
+                if( as.list ){
+                    .split_by_ranks(x, ranks, args)
+                }
+                else{
+                    x <- as(x, "TreeSummarizedExperiment")
+                    warning("SummarizedExperiment does not have altExps slot. ",
+                            "Therefore, it has been converted to ",
+                            "TreeSummarizedExperiment.")
+                    altExps(x) <- c(altExps(x),.split_by_ranks(x, ranks, args))
```
TuomasBorman: Instead, can you just call callNextMethod() --> to not copy paste same code

TuomasBorman: SCE already has altExps slot. SE does not have

R/splitByRanks.R
TuomasBorman: This is method for SE. It does not have altExps so you have to convert it to TreeSE

NEWS
```diff
@@ -106,7 +106,8 @@ Changes in version 1.11.x
 + getExperimentCrossCorrelation fix: enable using of sampleMap in MAE.
 + Implemented the setTaxonomyRanks function to specify which ranks are recognized as taxonomy ranks.
 + Rename cluster to addCluster
-+ rename importers loadFromBiom, loadFromQIIME2, readQZA, loadFromMothur, loadFromMetaphlan, loadFromHumann
++ rename importers loadFromBiom, loadFromQIIME2, readQZA, loadFromMothur, 
+loadFromMetaphlan, loadFromHumann
```
TuomasBorman: Can you revert this line

TuomasBorman:
To summarize:

1. Leave splitByRanks as alias. (just empty function calling agglomerateByRanks)
2. In SE method, add TreeSE conversion if the class is SE. (Give warning about the conversion)
3. Add SCE/TreeSE specific functionality to SCE method (TreeSE is SCE). Then call SE method.
4. TreeSE method is just calling SCE method.

R/splitByRanks.R
theotimeee: > Also do this:
> 
> x <- .split_by_ranks(x, ranks, args) if( !as.list ){ # Add to altexps } return(x)

If I do that, won't the method return a SimpleList no matter the value of `as.list`? The value returned by `.split_by_ranks` internal function is a SimpleList if I am not mistaken. Do we want to return a SimpleList even when the result is added to altExps?

R/splitByRanks.R
theotimeee: Here I added the agglomerateByRanks information to agglomerate-methods man page. If I add NULL, I think it does not add all the information written above to the man page and only add agglomerateByRanks methods to Usage part

TuomasBorman: If you add #' @rdname agglomerate-methods to line 83 it should work. If not, then this is ok. This NULL resets roxygen settings --> that way we have full control on where these document chunks are going.

R/splitByRanks.R
TuomasBorman: Yes, you are correct. How about

```
res <- .split_by_ranks()
# Add to altExp if user has specified to do so
if( !as.list ){
    # Create a general function that adds results to altExps (That could be used in other functions also in the future)
    res <- .add_to_altExps(x, res)
}
return(res)
```

That might be easier to read

R/splitByRanks.R
theotimeee: I tried but it did not work, I am not sure but NULL may work as intended only with document chunks using #' @name xyz

TuomasBorman: Might be. Thanks!

PR #515, opened 2024-04-16
Deprecate functions *_join

theotimeee:
This PR deprecates following functions :

- `full_join`
- `inner_join`
- `left_join`
- `right_join`

These functions were aliases for `mergeSEs` with parameter `join`, thus they do not need to be replaced.

### DIFF
--- a/R/deprecate.R
+++ b/R/deprecate.R
@@ -104,3 +104,71 @@ loadFromHumann <- function(...) {
                             " Use 'importHUMAnN' instead."))
     importHUMAnN(...)
 }
+
+#' @rdname deprecate
+#' @export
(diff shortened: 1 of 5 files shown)


antagomir (approved):

antagomir:
Could you check that we have at least one example in OMA Ch. 3 about using the SEtools if these are removed from here?

TuomasBorman:
We still have mergeSEs function. The intention was to remove these aliases; they make things more complicated without adding value

antagomir:
right - and in fact we already have examples in oma as well so I guess it is Ok

TuomasBorman (approved):

PR #517, opened 2024-04-17
deprecate plotNMDS

theotimeee:
This PR deprecates `plotNMDS` in the mia package .
`plotNMDS` is a wrapper for `scater::plotReducedDim` and will be added to the miaViz package (see https://github.com/microbiome/miaViz/pull/116)


TuomasBorman (approved):
Yes, this is in wrong place. Plotting functions should be in miaViz

I will merge this soon since there are one PR in merging queue

PR #518, opened 2024-04-17
deprecate subset* functions

theotimeee:
This PR deprecates the following functions : 

- `subsetSamples`
- `subsetFeatures`
- `subsetTaxa`

These functions are deprecated because using brackets is more convenient (works on both dimension at the same time, is more flexible and is used throughout R to subset data with two or more dimension).


TuomasBorman (commented):
Since the whole file will be deleted. It is easier if you just put deprecation messages and leave the functions where they are. We find them from here and we can remove their internal functions with them (check that internal functions are not used in functions)

theotimeee:
> Since the whole file will be deleted. It is easier if you just put deprecation messages and leave the functions where they are. We find them from here and we can remove their internal functions with them (check that internal functions are not used in functions)

I checked and the internal functions in this file are only used for these subset functions

TuomasBorman (approved):

Issue #519, opened 2024-04-18
Fix estimateDiversity input check


TuomasBorman:
Related to this PR https://github.com/microbiome/mia/pull/514

I noticed that `match.args` is little bit problematic in `estimateDiversity` functions' input check. It subsets indices that are supported without giving any notification if user inputted unsupported indices along with supported ones. Also, `name` parameter gets `index` as default value and it is not updated after `match.args` call which leads to missmatch and uninformative error.

```
library(mia)

data(GlobalPatterns)
tse <- GlobalPatterns

index <- c("shannon", "test")
# Uninformative error
res <- estimateDiversity(tse, index = index)

# "shannon" passes through in match.args call. Now this works
index <- c("shannon", "test")
name <- c("shannon")
res <- estimateDiversity(tse, index = index, name = name)
```

-->

Fix input check for `index` parameter. Instead of using `match.args`, check that all the indices are supported ones.

```
supported_index <- c("shannon", "index2")
if( !all( index %in% supported_index ) ){
    stop("'index' contains unsupported indices.", call. = FALSE)
}
```

PR #514, opened 2024-04-16
Issue 334 estimate diversity

ndegunst:
This should do the trick for #334. I've made it so that, if there's no tree and someone tries to calculate a faith index, the function will not calculate it and throws a warning instead.

PR #520, opened 2024-04-18
changed diversity input check

ndegunst:
I changed around the input check for the index as suggested in #519 . Sorry for the headache caused by #514.

### DIFF
--- a/R/estimateDiversity.R
+++ b/R/estimateDiversity.R
@@ -258,7 +258,7 @@ setMethod("estimateDiversity", signature = c(x="SummarizedExperiment"),
                              "inverse_simpson", "log_modulo_skewness", "shannon")
         index_string <- paste0("'", paste0(supported_index, collapse = "', '"), "'")
         if ( !all(index %in% supported_index) || !(length(index) > 0)) {
-            stop("'", paste0(supported_types, collapse = "', '"), "'", call. = FALSE)
+            stop("'", paste0("'index' must be from the following options: '",supported_types), "'", call. = FALSE)
         }
         
         if(!.is_non_empty_character(name) || length(name) != length(index)){
@@ -292,7 +292,7 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
                              "inverse_simpson", "log_modulo_skewness", "shannon")
         index_string <- paste0("'", paste0(supported_index, collapse = "', '"), "'")
         if ( !all(index %in% supported_index) || length(index) == 0) {
-            stop(paste("Unsupported index:", index_string), call. = FALSE)
+            stop(paste("'index' must be from the following options: '", index_string), call. = FALSE)
         }
         
         # Check tree_name
@@ -303,7 +303,6 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
         if (!is.null(assay_name)) {
             .Deprecated(old="assay_name", new="assay.type", "Now assay_name is deprecated. Use assay.type instead.")
         }
-        
         
         if(!.is_non_empty_character(name) || length(name) != length(index)){
             stop("'name' must be a non-empty character value and have the ",
--- a/R/estimateDiversity.R
+++ b/R/estimateDiversity.R
@@ -266,7 +266,7 @@ setMethod("estimateDiversity", signature = c(x="SummarizedExperiment"),
         
         if(!.is_non_empty_character(name) || length(name) != length(index)){
             stop("'name' must be a non-empty character value and have the ",
-                "same length than 'index'.",
+                "same length as 'index'.",
                 call. = FALSE)
         }
         .check_assay_present(assay.type, x)
@@ -315,7 +315,7 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
         
         if(!.is_non_empty_character(name) || length(name) != length(index)){
             stop("'name' must be a non-empty character value and have the ",
-                "same length than 'index'.",
+                "same length as 'index'.",
                 call. = FALSE)
         }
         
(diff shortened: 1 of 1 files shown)


TuomasBorman:
No problem at all, it is maintainers job to evaluate the code, and I don't assume that you know yet all the things in mia. The main goal is to develop good things with the best possible quality.

1. Can you modify also TreeSE method (same thing there, but now with additional faith index)
2. Run BiocCheck::BiocCheck() (Check that the code added by you passes Bioconductor recommendations. There are warnings related to other lines, skip those.)
3. Run manual tests that the code does what it should do. (Can you run the code from the issue? Can you print the code and output here?)

R/estimateDiversity.R
```diff
@@ -254,7 +254,11 @@ setMethod("estimateDiversity", signature = c(x="SummarizedExperiment"),
         }
 
         # input check
-        index<- match.arg(index, several.ok = TRUE)
+        supported_index <- c("coverage", "fisher", "gini_simpson", 
+                             "inverse_simpson", "log_modulo_skewness", "shannon")
+        if( !all( index %in% supported_index ) ){
+            stop("'index' contains unsupported indices.", call. = FALSE)
+        }
```
TuomasBorman: Looks good! Also check that the length of index is over 0 (it is not checked currently. NULL might cause problems --> test `index = NULL`

TuomasBorman (changes requested):

ndegunst:
Thanks for the tips !

I ran the BiocCheck with no warnings for the lines I modified and I ran some manual tests on top of the already existing tests, here are the results :

```R

tse <- estimateDiversity(tse, 
                              assay.type = "counts",
                              index = "test", 
                              name = "shannon")

```

Output : 
```
Error: 'index' contains unsupported indices.

```

```R
tse <- estimateDiversity(tse, 
                         assay.type = "counts",
                         name = "shannon")
```

Output : 
```
Error: 'name' must be a non-empty character value and have the same length as 'index'.

```

```R

tse <- estimateDiversity(tse, assay.type = "counts"
                         , index = NULL,
                         name = "test")

```

Output : 
```
Error in .local(x, assay.type, assay_name, index, name, ...) : 
  No 'index' specified.

```

Would you rather have it so that it throws an index error or a name error ?

R/estimateDiversity.R
```diff
@@ -283,19 +291,31 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
             name = index, tree_name = "phylo", 
             ..., BPPARAM = SerialParam()){
         # input check
+        supported_index <- c("coverage", "fisher", "gini_simpson", 
+                             "inverse_simpson", "log_modulo_skewness",
+                             "shannon","faith")
+        
+        if(!length(index)>0){
+            stop("No 'index' specified.")
```
TuomasBorman: Remember to add call. = FALSE (test what it does)

R/estimateDiversity.R
```diff
@@ -283,19 +291,31 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
...
+        }
```
TuomasBorman: Also you could remove this check and modify the test below

R/estimateDiversity.R
```diff
@@ -283,19 +291,31 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
...
+        
+        if( !all( index %in% supported_index ) ){
```
TuomasBorman: This check could test

1. length > 0
2. All values are supported

The error message could be:

"'index' must be a character value specifying alpha diversity index to be calculated. Choose from the following options: ..."

In ... you could print supported types. Something like thia

`stop("'", paste0(supported_types, collapse = "', '"), "'", call. = FALSE)`

TuomasBorman:
Looks good! Run `BiocCheck::BiocCheck()` and fix everything related to these added lines. (There are some other things related to other lines, but let's keep them unchanged at this point)

ndegunst:
That should do it, we shouldn't use collapse within a paste call and if we do past it shouldn't be within a stop call. Therefore, I assembled the string before using it in the stop function.

ndegunst:
It should be good now, bioconductor doesn't allow a paste statement within a stop so i created a new variable that pastes the allowed indices before passing it to the stop function.

R/estimateDiversity.R
```diff
@@ -283,19 +288,26 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
             name = index, tree_name = "phylo", 
             ..., BPPARAM = SerialParam()){
         # input check
+        supported_index <- c("coverage", "fisher", "gini_simpson", "faith",
+                             "inverse_simpson", "log_modulo_skewness", "shannon")
+        index_string <- paste0("'", paste0(supported_index, collapse = "', '"), "'")
+        if ( !all(index %in% supported_index) || length(index) == 0) {
+            stop(paste("Unsupported index:", index_string), call. = FALSE)
```
TuomasBorman: The error message could still be more informative.

"'index' must be from the following options:"

R/estimateDiversity.R
```diff
@@ -283,19 +288,26 @@ setMethod("estimateDiversity", signature = c(x="TreeSummarizedExperiment"),
...
                  call. = FALSE)
         }
         if (!is.null(assay_name)) {
             .Deprecated(old="assay_name", new="assay.type", "Now assay_name is deprecated. Use assay.type instead.")
-        }	
-        # Check indices
-        index <- match.arg(index, several.ok = TRUE)
+        }
+        
+
```
TuomasBorman: Remove extra line (one is enough)

With these, good to go

TuomasBorman (approved):

PR #524, opened 2024-04-19
rename calculateDPCoA, calculateNMDS, calculateRDA, calculateCCA

theotimeee:
This PR makes the following naming modifications :

- `calculateDPCoA` --> `getDPCoA`

- `calculateNMDS` --> `getNMDS`

- `calculateRDA` --> `getRDA`

- `calculateCCA` --> `getCCA`


TuomasBorman:
Thanks! Certain ordination methods that we use come from scater package. They use calculate* and run*, for instance, calculatePCA and runMDS. That is why we have to think about if we

1. Keep these unchanged
2. Keep calculate* and run* aliases
3. Replace calculate* and run* with get* and add* like done in this PR
4. Add get* and run* aliases also for scater ordination methods.

I have to think about this little bit

R/deprecate.R
```diff
@@ -362,6 +365,118 @@ loadFromHumann <- function(...) {
...
+            function(x,y,...) {
+                .Deprecated(msg = paste0("'calculateDPCoA' is deprecated. ",
+                                        "Use 'getDPCoA' instead."))
+                getDPCoA(x,y,...)
+            }
+)
+
+#' @rdname deprecate
+#' @export
+setGeneric("calculateNMDS", function(x, ...) standardGeneric("calculateNMDS"))
```
TuomasBorman: Yes. As scater still has the calculate* and run* naming scheme, we should still keep these as aliases.

You could do that by just adding the "regular" R function

```
calculateNMDS <- function(x, ...){
    getNMDS(x, ...)
}
```

Also do same for run* functions --> rename them to addRDA for instance and then create similarly alias runRDA

NEWS
```diff
@@ -106,14 +106,18 @@ Changes in version 1.11.x
 + getExperimentCrossCorrelation fix: enable using of sampleMap in MAE.
 + Implemented the setTaxonomyRanks function to specify which ranks are recognized as taxonomy ranks.
 + Rename cluster to addCluster
++ rename importers loadFromBiom, loadFromQIIME2, readQZA, loadFromMothur, 
+loadFromMetaphlan, loadFromHumann
```
TuomasBorman: Can you fix this? There is extra line break. I'm not sure if NEWS file works correctly if there are breaks

R/deprecate.R
thpralas: Is it OK to add them as regular R function if I do not add the methods to the deprecate file? Here it deletes the methods names from the namespace file since they are not in the deprecate file but replaced with a regular function

TuomasBorman (approved):

PR #529, opened 2024-04-23
Rename getPrevalence to addPrevalence and getPrevalentAbundance to addPrevalentAbundance

theotimeee:
This PR makes the following naming modifications : 

- `getPrevalence` --> `addPrevalence`
- `getPrevalentAbundance` --> `addPrevalentAbundance`


TuomasBorman:
getPrevalence and getPrevalentAbundance return vectors. They do not add results to TreeSE. That is why they should be kept as get*

Issue #445, opened 2023-09-09
Optimize mapTaxonomy


antagomir:
Currently `mapTaxonomy` is slow for longer taxa vectors.

This seems be because it maps every entry along taxa vector separately, instead of mapping unique entries and then mapping back to original input.

This could be considerably optimized in terms of speed by using this trick.

Could also check at the same go if there are other similar issues in other methods..

PR #533, opened 2024-04-25
Issue 445 map taxonomy

ndegunst:
I optimized the mapTaxonomy function as indicated in #445 by mapping over unique taxa and then mapping back to the original data. Even though the time difference by optimizing doesn't seem staggering, it still improved a little bit.


R/taxonomy.R
```diff
@@ -561,78 +561,84 @@ setGeneric("mapTaxonomy",
...
+              }
+              #
+              td <- rowData(x)[,taxonomyRanks(x)]
+              # If no taxa are provided, return unique taxonomy identifiers
+              if(is.null(taxa)){
+                  return(unique(td)) 
+              }
+              r_fs <- NULL
+              c_f <- rep(TRUE,ncol(td))
+              unique_taxa <- unique(taxa)
```
TuomasBorman, line 639: Is this variable used?

TuomasBorman: --> yes in line 623

However, it would be more simple to use it also in line 615 and 618 instead of calculating unique values again

R/taxonomy.R
```diff
@@ -561,78 +561,84 @@ setGeneric("mapTaxonomy",
...
+              if(!is.null(from)){
+                  r_fs <- lapply(unique(taxa), .get_taxa_row_match, td = td, from = from,
+                                 use_grepl = use_grepl)
+              } else {
+                  r_fs <- lapply(unique(taxa), .get_taxa_any_match, td = td,
+                                 use_grepl = use_grepl)
+              }
+              names(r_fs) <- unique(taxa)
+              # Assign names to the resulting list
+              names(r_fs) <- unique_taxa
```
TuomasBorman: These lines do the same thing?

R/taxonomy.R
```diff
@@ -561,78 +561,84 @@ setGeneric("mapTaxonomy",
...
+              # Map back the results to original input
+              ans <- lapply(r_fs, .get_map_result, td = td, c_f = c_f)
+              names(ans) <- names(r_fs)
+              u_len <- unique(lengths(ans))
+              # If all results have the same length, unlist the result
+              if(length(u_len) == 1L && u_len == 1L){
+                  ans <- unlist(ans) 
+              }
+              #
+              ans
```
TuomasBorman, line 663: use return(ans)

TuomasBorman (approved):
Seems working.

Couple of notes:

1. Have you run BiocChec::BiocCheck() to check that these additions follow the guidelines of Bioconductor? --> You could state that you have done that

2.

There are lots of additional lines. This same could be done with only one additional line `taxa <- unique(taxa)`. You could state in PR why you have modified all these lines. Probably there is good reason but reviewing is easier if we know why something has been done. (Correcting the code based on BiocCheck is good reason).

Also, important thing to note is that people are using these methods. We should not modify them without good justification; rather we should avoid making changes. If user checks the code, it seems that everything has changed, and that does not give good impression. (of course we should continuously improve the code)

R/taxonomy.R
TuomasBorman: Or maybe the best solution would be to modify line to `taxa <- unique(taxa)`. By doing that, only one line needs to be changed

TuomasBorman (changes requested):
Also was this the intention of the issue @antagomir 

It seems that this fixes the problem described in the issue, but the speed improvement seems to be very limited. Since I don't know the use case. With `taxa`, user can specify pattern that is fetched from the `rowData`. This modification would improve the performance when

`mapTaxonomy(tse, taxa = c("Escherichia", "Escherichia")) `

--> only one output is returned

ndegunst:
I implemented the requested changes and ran the biocCheck which didn't seem to have any issue with the new lines. The taxa have to be mapped back to the originial data. Hence the reason it didn't considerably decrease the computation time.

ndegunst:
However, I noticed that the issue was closed just after I made the PR. Was this already fixed ?

antagomir:
Perhaps my mistake - I reopened it.

TuomasBorman:
> I implemented the requested changes and ran the biocCheck which didn't seem to have any issue with the new lines. The taxa have to be mapped back to the originial data. Hence the reason it didn't considerably decrease the computation time.

OK, cool! Hmmm, I ran the BiocCheck and it seems that these new lines are incorrectly indented (should be multiplication of 4 spaces). The indentation in old lines were correct, so I think you should revert the changes related to lines that are not really fixing the problem mentioned in the issue. Sorry about extra work

https://contributions.bioconductor.org/r-code.html?q=indentation#indentation

E: BiocCheck generates file mia.BiocCheck/00BiocCheck.log that has more detailed info than what is printed on console

TuomasBorman:
The indentation is still off. If there is juts couple lines that you are planning to change, you can create new PR and link this with that

antagomir:
@Insaynoah what's the status of this PR? Can we get it fixed.

TuomasBorman:
Closing. The following PR will fix the issue https://github.com/microbiome/mia/pull/600

Issue #532, opened 2024-04-24
R CMD Check contains NOTES


TuomasBorman:
Hi,

currently check contains NOTES. Especially it seems that syntax of list in roxygen has changed.

1. Run devtools::check()
2. List probelmatic things to this issue
3. Let's discuss what we can do about them. (Everything cannot be solved)
4. Fix

ake123:
Here is the result after running `devtools::check()`
```r
❯ checking package dependencies ... NOTE
  Imports includes 21 non-default packages.
  Importing from so many packages makes the package vulnerable to any of
  them becoming unavailable.  Move as many as possible to Suggests and
  use conditionally.

❯ checking top-level files ... NOTE
  Non-standard file/directory found at top level:
    ‘CONTRIBUTING.md’

❯ checking Rd cross-references ... NOTE
  Unknown package ‘https’ in Rd xrefs
```

TuomasBorman:
Do you have the latest versions? In addition to those NOTEs that you get, I get also NOTEs about roxygen

See the latest workflow run

https://github.com/microbiome/mia/actions/runs/8849563656/job/24301848155

antagomir:
The status of this issue?

TuomasBorman:
This is solved. All the checks (cmd and Bioc) are as polished as possible. (Let's keep it that way in the future also)

PR #540, opened 2024-05-08
Temporal fix for rworkflow bug

0xMuluh:
This commit provides a temporal fix to rworkflow by bumping the r version of MacOS and Ubuntu and
fixing a  bug in R/estimateDiversity.R.


0xMuluh:
Bumping r version for Mac workflow tends to work. I think the same for Ubuntu at least, until the fix in 
R/estimateDiversity.R is merged.

TuomasBorman:
Alright, maybe we could go like this as a temporary fix. Can you fix estimateDiversity.R in this PR (change the title also according to that)

TuomasBorman:
Roxygen2 changed how list is created. It seems that it throws now a warning. This is related to this issue: https://github.com/microbiome/mia/issues/532

Check the workflow run, you will see these warnings:
```

 Warning in readLines(conn, line - oldline, warn = FALSE) :
    invalid input found on input connection './man/estimateDominance.Rd'
  checkRd: (-1) estimateDominance.Rd:99-101: Lost braces in \itemize; meant \describe ?
```

Can you fix them?

See for example from [here](https://github.com/EBI-Metagenomics/HoloFoodR/blob/devel/R/getResult.R) how lists are now created

0xMuluh:
@TuomasBorman I think we can finalize this now. Unless, you have more changes in mind

TuomasBorman:
Looks good, but still not working. Update manual pages, let's see if that works

0xMuluh:
I updated man pages already. The fail is as at the stage of deployment. Something about the bot not having permission to deploy to the repo.  See log [here](https://github.com/microbiome/mia/actions/runs/9059118621/job/24886047784?pr=540#step:4:5706)

TuomasBorman:
Now there is another error. Can you remove the rworkflows file and then create a new one with rworkflows::use_workflow(). That should update the workflow. Remember to specify R version to 4.4.0 (auto seems not to work)

In the future, create branch directly to mia. It is simpler since then I can also contribute much easier.

TuomasBorman:
https://github.com/neurogenomics/rworkflows/issues/130

Try to put R back to auto. It should work

TuomasBorman:
From the rworkflows issue

> One thing that might help is removing the ubuntu-latest rspm line, or replacing it with rspm: ~

Can you try that fix?

0xMuluh:
> From the rworkflows issue
> 
> > One thing that might help is removing the ubuntu-latest rspm line, or replacing it with rspm: ~
> 
> Can you try that fix?

Yeah, I tried this. Still failing. I think the issue here is permission to microbiome/mia.git is denied to github-actions[bot]. I'm not quite sure but did we alter the secrets?

TuomasBorman:
> > From the rworkflows issue
> > > One thing that might help is removing the ubuntu-latest rspm line, or replacing it with rspm: ~
> > 
> > 
> > Can you try that fix?
> 
> Yeah, I tried this. Still failing. I think the issue here is permission to microbiome/mia.git is denied to github-actions[bot]. I'm not quite sure but did we alter the secrets?

Yes, I noticed that too. However, nothing should have changed. I will check

PR #551, opened 2024-05-20
Fix agglomeration of trees

TuomasBorman:


TuomasBorman:
Related to https://github.com/microbiome/mia/pull/548

PR #556, opened 2024-05-23
Fixes to agglomeration

TuomasBorman:


TuomasBorman:
Related to this PR https://github.com/microbiome/mia/issues/419.

While polishing the code, I realized that our modifications related to na.rm and onRanksOnly were not good.

onRanksOnly in agglomerateByRank controls whether to check higher ranks or not. Basically, when onRanksOnly=TRUE, the agglomeration is done similarly to agglomerateByVariable.

Also, I think added na.rm to agglomerateByVariable was is too complicated, and does not really bring benefits.

So I suggest that we remove:

1. Switch back to onRankOnly=TRUE by default
2. Remove na.rm from agglomerateByVariable


This is to show behavior in this PR

```
library(mia)
tse <- GlobalPatterns

# Works similarly to agglomerateByVarible
tse1 <- agglomerateByRank(tse, rank = "Family", onRankOnly = TRUE)
tse2 <- agglomerateByVariable(tse, MARGIN = 1, f = "Family")
dim(tse1) # 334  26
dim(tse2) # 334  26

# Now it works differently than agglomerateByVariable since it takes into accounts also higher ranks
# Now it might be that there are duplicated rownames. Same family names might have different phyla rank.
tse1 <- agglomerateByRank(tse, rank = "Family")
dim(tse1) # 341  26

# If we set, na.rm == FALSE, we get also taxa that are in higher rank (rows might include phyla for instance)
tse1 <- agglomerateByRank(tse, rank = "Family", na.rm = FALSE)
dim(tse1) # 603  26

# This is how to get similar result by using agglomerateByPrevalence
tse1 <- agglomerateByRank(tse, rank = "Family")
tse1 <- agglomerateByPrevalence(tse1, assay.type="counts", detection  = 0.5/100, prevalence  = 20/100)
tse2 <- agglomerateByPrevalence(tse, assay.type="counts", detection  = 0.5/100, prevalence  = 20/100, rank = "Family")
dim(tse1) # 269  26
dim(tse2) # 269  26

# Or if we want to also include higher ranks
tse1 <- agglomerateByRank(tse, rank = "Family", na.rm = FALSE)
tse1 <- agglomerateByPrevalence(tse1, assay.type="counts", detection  = 0.5/100, prevalence  = 20/100)
tse2 <- agglomerateByPrevalence(tse, assay.type="counts", detection  = 0.5/100, prevalence  = 20/100, rank = "Family", agg.na.rm = FALSE)
dim(tse1) # 446  26
dim(tse2) # 446  26
```

@antagomir 

@Daenarys8 can you solve errors in unit tests? Also, can you check that the documentation reflects these new modifications?

antagomir:
1. onRankOnly default - sounds reasonable
2. "Remove na.rm from agglomerateByVariable": might make sense but I didnt find the reasoning why it is not good there but it would be however needed in agglomerateByRank; ideally the treatment would be as unified as possible. But deviations are ok when it is justified, of course.
3. "agglomerateByVariable(tse, MARGIN = 1, f = "Family")" in fact the use of MARGIN is in some sense nonconventional; usually margin=1 in R means that operations are done _per row_ but in this case the operations are done _per column_ even if the idea is that the function operates "on the rows".. I might be in favor of changing this so that the MARGIN meaning is switched. In the end, we *are* here aggregating variables _by sample_ , correct? It would be crucial to be very clear about this in the documentation and examples.

TuomasBorman:
2.

I think it complicates things. The merging is done with factor and factor works like that; NA is handles as "no group" so these NA values are not grouped. Instead, now we are creating artificial "NA" group.

In the previous PR, we only added na.rm to agglomerateByVariable, agglomerateByRank already had that. However, I'm ok with that if we decide to keep na.rm also for *Variable, but there were bugs in the code. When *Rank is calling *Variable internally, na.rm must be hard-coded to TRUE (which was the behavior earlier before these modifications).

3.

I am not sure of the formal definition of MARGIN, but for me it is clear that MARGIN = 1 means that a) the agglomeration is applied to rows b) the variable can be found from rowData. Of course I am biased. The documentation should be modified so that it states this clearly enough. Also if that does not follow the formal definition, I guess we could stretch it little bit.

antagomir:
2. It would be useful, in principle, if user can choose whether to include "NA" group or just exclude NA cases. Being blissfully ignorant of some technicalities, I would be by default in favor of keeping na.rm as a standard argument in all those methods (perhaps with default na.rm=FALSE, which would then generate the "NA" group? But if na.rm option is too complicated then one could just include "NA" group always, and show how to remove it as a separate post processing step. However that is not equally neat solution. 

3. This seems like an important point. Consider for instance: `apply(x, 1, sum) == rowSums(x)`. Clearly, this does the operation per each row. In our case, `agglomerateByVariable(tse, MARGIN = 1, f = "Family")` would operate per column. The apply family of methods is very central to the R ecosystem, I am not sure if we want to convert the margin meaning here. The other option is that we systematically align with the same conventions than apply methods have, and try to be very clear about this in all documentation.

TuomasBorman:
2. This is ok. There was a bug that I fixed.

3. I understand your point. I think MARGIN = 2 when we want to agglomerate based on rowData is also problematic since it might be hard to understand. I don't have any clear ideas on how to solve this.

btw, @Daenarys8 can you make those tests simpler. Those that you made in last PR. They are quite hard to follow. You could check that when you agglomerateByRank, it matches with rowData(tse)$family for instance. Also same with agglomerateByVariable. Then you could check that you get same result with agglomerateByPrevalence + agglomeration and agglomerateByRank

TuomasBorman:
3.
 I think we could change the name. MARGIN is used in other functions also so we could switch them too to harmonize. For instance, `direction` might be clear

antagomir:
I am still not certain whether 1 vs. 2 is confusing in the longer run. I think it is also a matter of convention & getting used to it. When it aligns with other R functions (apply family of functions), then it should be relatively easy to memorize. Examples are also available for those who need to check. Message could be thrown, something like "Agglomerating rows" so users will also see what is happening. 

The advantage of margin 1/2 is a more direct compatibility with other methods/packages.

I agree that at least we should not use MARGIN in a different meaning than other common functions do, therefore  `direction` could be an option if we must change (which I am not sure how critical it is). Such argument could then potentially have values "rows"/"cols" instead of 1/2 if that is more clear. But I am not sure if that is any more clear because the same argumentation applies there too..

TuomasBorman:
Points 1 and 2 are now fixed. The third point (MARGIN) is still open, but we have PR open about renaming. We could move the discussion there. https://github.com/microbiome/mia/pull/562

Then we could proceed with this. --> fix unit tests

antagomir:
Good to me - but if this is now in conflict with the established use of MARGIN then this is urgent imo.

TuomasBorman:
I think MARGIN conflicts, but 1 and 2 not. 1 and 2 are established way to specify rows and columns, but they do not conflict with our way to use them since they just mean "row" and "columns" Moreover, "rows" and "cols" are currenrtly supported by MARGIN (or the parameter name that we decide), although they are not used in examples

antagomir:
If we allow something like `direction` then could we implicitly also allow MARGIN for users who prefer that compatibility? In the same way that 1 vs. "rows" is now allowed?

We could potentially cross-check also in Bioc Slack what the broader consensus seems to be (if any).

TuomasBorman:
It is much simpler to allow different values for direction/MARGIN; it all happens under the hood. But the extra parameter is visible for user so I would not support multiple parameters for same thing.

--> I would go with direction="row" (but 1 and "rows", would still be supported but not officially)

Yes, we could ask that

R/agglomerate.R
```diff
@@ -333,16 +333,14 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
 #' @export
 setMethod("agglomerateByVariable",
             signature = c(x = "TreeSummarizedExperiment"),
-            function(x, MARGIN, f, archetype = 1L, mergeTree = FALSE,
-                     mergeRefSeq = FALSE, ...){
```
TuomasBorman, line 319: As mergeRefSeq is related only for rows direction, I am wondering if that should be hidden parameter

R/agglomerate.R
```diff
@@ -333,16 +333,14 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
...
+            function(x, MARGIN, f, archetype = 1L, mergeTree = FALSE, ...){
                 # Check MARGIN
                 MARGIN <- .check_MARGIN(MARGIN)
                 # Get function based on MARGIN
                 FUN <- switch(MARGIN, .merge_rows_TSE, .merge_cols_TSE)
                 # Agglomerate
                 x <- FUN(
-                    x, f, archetype = archetype, mergeTree = mergeTree,
-                    mergeRefSeq = mergeRefSeq, ...)
```
TuomasBorman: But this disables the usage of mergeRefSeq; user cannot use that anymore, so this is wrong

TuomasBorman: This is caused because this parameter is passed thorugh with "...".

The column-related functions do not have this parameter so this parameter flows with "...". Every time certain function is called, it check that the parameters match. Those paramaters that do not match, can flow with "...". If there is no "..." in function definition, the function throws error "unused argument"

TuomasBorman: Sorry, my mistake. There is already mergeRefSeq in internal function, so this should actually work...

0xMuluh: ah yes. However,  mergeRefSeq is documented in agglomerat-methods and issues usage warning with the change I made. Unless it is ok to have mergeRefSeq undocumented. is it?

0xMuluh: mergeRefSeq is restored with  temporal switch just in case we need it documented.

TuomasBorman: You can put it under "..."

TuomasBorman: In agglomerate-methods documentation

TuomasBorman: Or one option is to add mergeRefSeq to agglomerateByRank. That function is only for rows so it would make sense. The function calls agglomerateByVariable internally

0xMuluh: okay. I add to agglomerateByRank then.

R/agglomerate.R
```diff
@@ -63,6 +69,7 @@
 #'        \item \code{as.relative}: Logical scalar: Should the detection 
 #'        threshold be applied on compositional (relative) abundances? 
 #'        (default: \code{FALSE})
+#'        \item \code{}
```
TuomasBorman: Why there is empty item

R/agglomerate.R
```diff
@@ -328,17 +337,14 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
 #' @export
 setMethod("agglomerateByVariable",
             signature = c(x = "TreeSummarizedExperiment"),
-            function(x, MARGIN, f, archetype = 1L, mergeTree = FALSE,
-                     mergeRefSeq = FALSE, ...){
+            function(x, MARGIN, f, archetype = 1L, mergeTree = FALSE, ...){
```
TuomasBorman: archetype is also related to mergeRefSeq so it should be moved also

TuomasBorman: I thought about this and maybe the cleanest solution is to move mergeRefSeq and archetype to hidden parameters also from agglomerateByRank(). All agglomerate* function are in same manual page, so I think it might be hard for user to find that the parameter is available also for agglomerateByVariable() if is listed as main parameter for agglomerateByRank()

-->

So move mergeRefSeq and archetype to hidden parameters and list them under "..." in documenration

tests/testthat/test-3agglomerate.R
```diff
@@ -29,123 +29,148 @@ test_that("agglomerate", {
...
+        prevalence  = 20/100)
+    actual3 <- agglomerateByPrevalence(
+        altExp(tse, "Family3"), assay.type="relabundance", detection  = 0.5/100,
+        prevalence  = 20/100)
+    actual4 <- agglomerateByPrevalence(
+        altExp(tse, "Family4"), assay.type="relabundance", detection  = 0.5/100,
+        prevalence  = 20/100)
+    actual5 <- agglomerateByPrevalence(
+        altExp(tse, "Family5"), assay.type="relabundance", detection  = 0.5/100,
+        prevalence  = 20/100)
```
TuomasBorman: Can you simplify these tests? It is very hard for developer to say what they are actually testing.

For example, agglomerateByPrevalence is just additional layer of complexity. Tests should:

1. Test that correct values are calculated
2. Be easy to read, so that we can easily debug what is the problem

The tests should be like this. This is easy to read what is the purpose:

```
test <- agglomerateByRank(tse, rank="Family", na.rm = TRUE)
ref <- unique(test$Family)
ref <- ref[ !is.na(ref) ]
ref <- length(ref)
expect_equal(nrow(test), ref)
```

R/agglomerate.R
0xMuluh: ACK.

R/agglomerate.R
0xMuluh: it's fixed now.

tests/testthat/test-3agglomerate.R
0xMuluh: ACK.

TuomasBorman:
So to summarize:

We have used MARGIN parameter in multiple functions.

1. getCrossAssociation(): determine if associations are calculated for rows (1) or columns (2).
2. addCluster(): Determines whether rows (1) or columns (2) are clustered.
3. agglomerateByRank(): Determines whether rows (1) or columns (2) are calculated.
4. transformAssay(): Determines whether transformation is applied row-wise (1) or column-wise (2).
5. splitOn(): Determines whether to split based on row (1) or column (2) variable.

The thing that connects them is that all parameters specify whether the operation is done on rows or columns. So I would have common name for these parameters. But as they have little bit meaning, there is no consensus of this in Bioc packages, that is what I found out.

MARGIN has specific meaning in R as you noted (if MARGIN = 1, then the number of rows will be unchanged), so we cannot use that. That is why I would go with `direction = "row"` or `direction = "col"`.

I would still keep 1, 2, "rows", and cols", as supported values as it does not cause any extra work https://github.com/microbiome/mia/blob/9198263a0f88f3776578629495b3c22fd5dfd722/R/utils.R#L139

TuomasBorman:
@antagomir

R/getPrevalence.R
```diff
@@ -275,7 +275,7 @@ setMethod("getPrevalence", signature = c(x = "SummarizedExperiment"),
 
         # check assay
         .check_assay_present(assay.type, x)
-        x <- .agg_for_prevalence(x, rank = rank, ...)
+        x <- .agg_for_prevalence(x, agg.na.rm = FALSE, rank = rank, ...)
```
TuomasBorman: This disables the usage of agg.na.rm. I reverted it

tests/testthat/test-3agglomerate.R
```diff
@@ -29,137 +29,113 @@ test_that("agglomerate", {
...
-    expect_equal(nrow(actual), nrow(actual0))
-    expect_equal(nrow(actual), nrow(actual1))
-    expect_equal(nrow(actual2), 27)
-    expect_equal(nrow(actual3), 20)
-    expect_equal(nrow(actual), nrow(actual4))
-    expect_equal(nrow(actual), nrow(actual5))
+    test1 <- agglomerateByRank(GlobalPatterns, rank="Family", na.rm = FALSE)
+    test2 <- agglomerateByVariable(GlobalPatterns, f="Family", MARGIN = "row", 
+       mergeTree = FALSE)
+    expect_equal(nrow(test1), nrow(test1))
```
TuomasBorman: This test tests only one point/perspective. Also it is not working correctly since there is type (test1 in both)

Instead, test that agglomerateByRank results to specific number of rows (just use rowData(tse)$family). And add  comments what you are trying to achieve.

Tests to implement
- agglomerateByRank and na.rm = FALSE
- agglomerateByRank and na.rm = TRUE
- agglomerateByVariable, and agg.na.rm = FALSE
- agglomerateByVariable, and agg.na.rm = TRUE

--> Test that these options outputs TreeSE whose nrow equals to specific number of families that you have fetched with rowData(tse)$family

0xMuluh: I am a bit confused. do I pass agg.na.rm to agglomerateByRank?

TuomasBorman: No, it is parameter of agglomerateByVariable

TuomasBorman: I rephrased my comment little bit

R/getPrevalence.R
0xMuluh: ACK.

tests/testthat/test-3agglomerate.R
```diff
@@ -145,7 +145,7 @@ test_that("agglomerate", {
     # Test that dimentionality is the same when NA values are removed.
     expect_equal(length((unique(rowData(tse)[,"Family"]))[ !is.na(unique(rowData(tse)[,"Family"])) ]),
                  nrow(agglomerateByRank(tse, rank="Family", na.rm = TRUE)))
-      expect_equal(length((unique(rowData(tse)[,"Family"]))[ !is.na(unique(rowData(tse)[,"Family"])) ]),
+    expect_equal(length((unique(rowData(tse)[,"Family"]))[ !is.na(unique(rowData(tse)[,"Family"])) ]),
                  nrow(agglomerateByVariable(tse, f="Family", MARGIN = 1, agg.na.rm = TRUE)))
```
TuomasBorman: Can you extract these values first and then check that they match? Also, agg.na.rm is not parameter of agglomerateByVariable, it is parameter of agglomerateByPrevalence.

agglomerateByVariable has parameter na.rm. So I am wondering if these functions work correctly.... Although it might be that there are no NA values in family level. Can you switch using the lowest rank in rowData? That should have NA values.

TuomasBorman: Looks good though. With these tests, good to go (when fixed)

0xMuluh: You are right. this is what i get: 
```
> data(GlobalPatterns, package="mia")
> tse <- GlobalPatterns
> actual <- unique(rowData(tse)[,"Species"])[ !is.na(unique(rowData(tse)[,"Species"]))]
> actual1 <- agglomerateByRank(tse, rank="Species", na.rm = TRUE)
> actual2 <- agglomerateByVariable(tse, f="Species", MARGIN = 1, na.rm = TRUE)
> expect_equal(length(actual), nrow(actual1))
> expect_equal(length(actual), nrow(actual2))
Error: length(actual) not equal to nrow(actual2).
1/1 mismatches
[1] 944 - 945 == -1
```
na.rm doesn't make a difference in agglomerateByVariable.

TuomasBorman:
Create this kind of checks
```

# Check that na.rm works
# Get all phyla
all_phyla <- unique( rowData(tse)$Phylum )
# When na.rm=TRUE, there should be as many rows as there are non-NA phyla
test <- agglomerateByVariable(tse, MARGIN = 1, f="Phylum", na.rm = TRUE)
nrow(test) == length( all_phyla[!is.na(all_phyla)] )
# When na.rm = FALSE, then phyla should also include NA --> one extra row
test <- agglomerateByVariable(tse, MARGIN = 1, f="Phylum", na.rm = FALSE)
nrow(test) == length( all_phyla )
```

tests/testthat/test-3agglomerate.R
TuomasBorman: Now it should work --> can you please test

0xMuluh:
> Create this kind of checks
> 
> ```
> 
> # Check that na.rm works
> # Get all phyla
> all_phyla <- unique( rowData(tse)$Phylum )
> # When na.rm=TRUE, there should be as many rows as there are non-NA phyla
> test <- agglomerateByVariable(tse, MARGIN = 1, f="Phylum", na.rm = TRUE)
> nrow(test) == length( all_phyla[!is.na(all_phyla)] )
> # When na.rm = FALSE, then phyla should also include NA --> one extra row
> test <- agglomerateByVariable(tse, MARGIN = 1, f="Phylum", na.rm = FALSE)
> nrow(test) == length( all_phyla )
> ```

na.rm works
```
> nrow(test) == length( all_phyla[!is.na(all_phyla)] )
[1] TRUE
> # When na.rm = FALSE, then phyla should also include NA --> one extra row
> test <- agglomerateByVariable(tse, MARGIN = 1, f="Phylum", na.rm = FALSE)
> nrow(test) == length( all_phyla )
[1] TRUE
```

R/agglomerate.R
```diff
@@ -49,20 +49,32 @@
...
+#'        \item \code{prevalence}: Prevalence threshold (in 0 to 1). The
+#'        required prevalence is strictly greater by default. To include the
 #'        limit, set \code{include_lowest} to \code{TRUE}.
-#'        \item \code{as.relative}: Logical scalar: Should the detection 
-#'        threshold be applied on compositional (relative) abundances? 
+#'        \item \code{as.relative}: Logical scalar: Should the detection
+#'        threshold be applied on compositional (relative) abundances?
 #'        (default: \code{FALSE})
+#'        \item \code{mergeRefSeq} \code{TRUE} or \code{FALSE}: Should a
+#'        onsensus sequence be calculated? If set to \code{FALSE}, the result
```
TuomasBorman: onsensus

tests/testthat/test-2merge.R
```diff
@@ -77,7 +77,8 @@ test_that("merge", {
     xtse <- TreeSummarizedExperiment(assays = list(mat = mat),
                                      rowRanges = unname(grl))
     FUN_check_x <- function(x,archetype=1){
-        actual <- agglomerateByVariable(x, MARGIN = "rows", f, archetype)
+        actual <- agglomerateByVariable(x, MARGIN = "rows", f, archetype, 
+            mergeTree = FALSE)
```
TuomasBorman, line 81: Has the behavior changed?

TuomasBorman: Why mergeTree was added?

tests/testthat/test-3agglomerate.R
```diff
@@ -29,137 +29,132 @@ test_that("agglomerate", {
...
-                                       detection  = 0.5/100, prevalence  = 20/100)
-    actual3 <- agglomerateByPrevalence(altExp(tse, "Family3"), assay.type="relabundance", 
-                                       detection  = 0.5/100, prevalence  = 20/100)
-    actual4 <- agglomerateByPrevalence(altExp(tse, "Family4"), assay.type="relabundance", 
-                                       detection  = 0.5/100, prevalence  = 20/100)
-    actual5 <- agglomerateByPrevalence(altExp(tse, "Family5"), assay.type="relabundance", 
-                                       detection  = 0.5/100, prevalence  = 20/100)
+    # Test that dimentionality is the same for merging object by agglomerateByRank
+    # and agglomerateByVariable.
+    expect_equal(nrow(test0), length(all_phyla))
```
TuomasBorman: Good. The add one more check that test that there are more taxa when agglomeration is done to "Species" level for instance.

nrow(agglomerateByRank)) == x

nrow(agglomerateByVariable)) == y

You can hardcode those values. So these checks will inform us if something has changed.

tests/testthat/test-5prevalence.R
```diff
@@ -314,71 +317,71 @@ test_that("subsetByRare", {
...
                                       as_relative=TRUE))
-    
-    prevalent <- rownames(subsetByPrevalent(GlobalPatterns, detection=d, prevalence=p, 
+
+    prevalent <- rownames(subsetByPrevalent(GlobalPatterns, detection=d, prevalence=p,
                                                 as_relative=TRUE))
-    
+
     all_taxa <- c(rare, prevalent)
-
```
TuomasBorman, line 346: I'm not sure if these lines should be modified like this.

TuomasBorman:
I think we are almost there, thanks!

tests/testthat/test-2merge.R
0xMuluh: Yes. when mergeTree is not passed, test fails.

TuomasBorman: Should be investigated why the behavior is changed. This is not intentional. I cannot see quickly why this happens. mergeTree is still visible parameter of agglomerateByVariable with default value FALSE

TuomasBorman: Can you check?

tests/testthat/test-5prevalence.R
0xMuluh: I only modified some deprecated function names and saved the file which might have removed white spaces.

TuomasBorman: One should always check what has been changed in the files. It makes the reviewing process easier and helps to maintain high-quality of code

0xMuluh: ACK.

tests/testthat/test-2merge.R
0xMuluh: the change occurred when we moved archetype to ...

R/agglomerate.R
0xMuluh: ACK.

TuomasBorman:
SHould be ok after resolving conflicts

Issue #498, opened 2024-03-01
addHierarchyTree replaces existing tree


TuomasBorman:
Currently, `addHierarchyTree` (`addTaxonomyTree`) replaces existing tree. Consider adding option for adding tree instead of replacing.

0xMuluh:
should we implement an addTree option to append the  rowTree and colTree slots ?

TuomasBorman:
Maybe there could be name parameter. By default, it could be "phylo". Then you could check if a tree named "phylo" is already existing and give warning if it is.

The created internal function could follow this structure https://github.com/microbiome/mia/blob/b627edce620807af20d3ed85c6667f4b8ef8f2ea/R/utils.R#L310. Also take into account the MARGIN in this internal function
.

antagomir:
Is this issue something we could close already @Daenarys8 @TuomasBorman

TuomasBorman:
Yes, I believe. Each row must be linked only once (rowLinks can have only one link per row). Based on this, it is not possible to not overwrite the existing tree

PR #557, opened 2024-05-23
Add function to add tree

0xMuluh:
ping #498


R/taxonomy.R
```diff
@@ -69,6 +69,10 @@
 #' @param use_grepl \code{TRUE} or \code{FALSE}: should pattern matching via
 #'   \code{grepl} be used? Otherwise literal matching is used.
 #'   (default: \code{FALSE})
+#'   
+#'@param MARGIN A character value for selecting if data is merged
+#'   row-wise / for features ('rows') or column-wise / for samples ('cols').
+#'   Must be \code{'rows'} or \code{'cols'}.
```
TuomasBorman, line 71: This could be hidden. There is not many use-case currently for MARGIN = 2, but good to have support

R/taxonomy.R
```diff
@@ -532,7 +536,7 @@ setGeneric("addHierarchyTree",
 #' @rdname hierarchy-tree
 #' @export
 setMethod("addHierarchyTree", signature = c(x = "SummarizedExperiment"),
-    function(x, ...){
+    function(x, replace = FALSE, ...){
```
TuomasBorman: Also these all could be moved to utils.R

R/utils.R
```diff
@@ -351,6 +351,46 @@
     altExps(x) <- values
     return(x)
 }
+
+# This function can be used to add tree to existing tree
+.add_tree <- function(x, tree, MARGIN, name = "phylo", ...) {
+  
+  MARGIN <- .check_MARGIN(MARGIN)
+  
+  if (MARGIN == 1L) {
```
TuomasBorman: Tidier way is to get function like this https://github.com/microbiome/mia/blob/b627edce620807af20d3ed85c6667f4b8ef8f2ea/R/agglomerate.R#L452

R/utils.R
```diff
@@ -351,6 +351,46 @@
...
+    # Add row tree
+    if (is.null(x@rowTree)) {
```
TuomasBorman: You should not use @. Instead, use rowTreeNames

R/utils.R
```diff
@@ -351,6 +351,46 @@
...
+      x@rowTree <- list(phylo = tree)
+    } else {
+      # Check if a tree with the given name already exists
+      if (name %in% names(x@rowTree)) {
+        warning(paste("A tree named", name, "already exists in the row tree. The new tree will be added with a different name."))
+        x@rowTree <- c(x@rowTree, list(tree))
+      } else {
+        x@rowTree[[name]] <- tree
```
TuomasBorman: I think this does not work

R/utils.R
```diff
@@ -351,6 +351,46 @@
     altExps(x) <- values
     return(x)
 }
+
+# This function can be used to add tree to existing tree
+.add_tree <- function(x, tree, MARGIN, name = "phylo", ...) {
```
TuomasBorman: So:

1. Add all the functionality to .add_tree() function.
2. Check MARGIN and name
3. Get correct functions based on MARGIN
4. Check if tree name. If it does exist, give warning
5. Add tree to object

changeTree() seems not to work when the tree does not already exist with that name.

TuomasBorman:
Can you check: it seems that row can have only one link in rowLinks. This means that no extra tree can be added (then there would be more than one linkage between single row and trees). There can be multiple trees if the dataset is divided into several trees. 

--> Can you check if that is true

0xMuluh:
Load sample data
```
> library(mia)
> data(GlobalPatterns, package="mia")
> tse <- GlobalPatterns
> tse <- transformAssay(tse, MARGIN = "samples", method="relabundance")
```
Check the rowLinks slot
`>row_links <- rowLinks(tse)`

Count the number of links per row
`> num_links_per_row <- sapply(row_links, length)`

Check if any row has multiple links
```
> has_multiple_links <- any(num_links_per_row > 1)
> has_multiple_links
TRUE
```

row seems to have multiple links.

TuomasBorman:
Your example does not show that. If you check rowLinks(tse):

1. Row in rowLinks() represent a row in TreeSE. For instance, row number 1 in TreeSE is a row number 1 in rowLinks
2. nodeNum tells which node represents the row in rowTree. In rowTree, multiple rows of TreeSE can be represented by a single node.

TuomasBorman:
But the question was:

1. If we add new rowTree, it must be linked with rows. 
2. Can there be more rows in rowLinks than there are rows in TreeSE. For instance in GlobalPatterns dataset there would be then almost 40,000 rows in rowLinks versus ~ 20,000 rows in TreeSE

--> I think this is not possible to do. I beliave the linkage between rows in TreeSE and rows in rowLinks is "hard" that way that each row in TreeSE is linked to corresponding index in rowLinks (but I might be wrong)

If that is the case, then we can just close this PR and issue --> the thing has be been checked

TuomasBorman:
Any updates on this?

TuomasBorman:
updates?

0xMuluh:
Not yet. Still unable to clarify if TreeSE can hold multiple *trees

PR #562, opened 2024-05-28
Organize parameter names

0xMuluh:
This commit organizes parameters used by user.

PR #570, opened 2024-06-05
Use agglomerateByVariable when onRankOnly = FALSE

0xMuluh:
#562 change behavior of agglomerateByRank when onRankOnly is FALSE

### DIFF
--- a/R/agglomerate.R
+++ b/R/agglomerate.R
@@ -23,10 +23,6 @@
 #' @param rank a single character defining a taxonomic rank. Must be a value of
 #'   \code{taxonomyRanks()} function.
 #'
-#' @param onRankOnly \code{TRUE} or \code{FALSE}: Should information only from
-#'   the specified rank be used or from ranks equal and above? See details.
-#'   (default: \code{onRankOnly = FALSE})
-#'
 #' @param na.rm \code{TRUE} or \code{FALSE}: Should taxa with an empty rank be
 #'   removed? Use it with caution, since empty entries on the selected rank
 #'   will be dropped. This setting can be tweaked by defining
@@ -98,14 +94,8 @@
 #'   returned. (Default: \code{mergeRefSeq = FALSE})
 #'
 #' @details
-#' When using \code{agglomerateByRank}, please note that depending on the 
-#' available taxonomic data and its structure, setting\code{onRankOnly = TRUE} 
-#' has certain implications on the interpretability of your results. If no loops
-#' exist (loops meaning two higher ranks containing the same lower rank), the 
-#' results should be comparable. You can check for loops using 
-#' \code{\link[TreeSummarizedExperiment:detectLoop]{detectLoop}}.
 #' 
-#' Also, agglomeration sums up the values of assays at the specified taxonomic level. With
+#' Agglomeration sums up the values of assays at the specified taxonomic level. With
 #' certain assays, e.g. those that include binary or negative values, this summing
 #' can produce meaningless values. In those cases, consider performing agglomeration
 #' first, and then applying the transformation afterwards.
@@ -242,7 +232,7 @@ setGeneric("agglomerateByVariable",
 #'
 #' @export
 setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
-    function(x, rank = taxonomyRanks(x)[1], onRankOnly = TRUE, na.rm = FALSE,
+    function(x, rank = taxonomyRanks(x)[1], na.rm = FALSE,
         empty.fields = c(NA, "", " ", "\t", "-", "_"), ...){
         # input check
         if(nrow(x) == 0L){
@@ -252,12 +242,6 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
         if(!.is_non_empty_string(rank)){
             stop("'rank' must be a non-empty single character value",
                 call. = FALSE)
-        }
-        if(!onRankOnly){
-            stop("Deprecated. Use 'agglomerateByVariable' instead.", call. = FALSE)
-        }
-        if(!.is_a_bool(onRankOnly)){
-            stop("'onRankOnly' must be TRUE or FALSE.", call. = FALSE)
         }
         if(!.is_a_bool(na.rm)){
             stop("'na.rm' must be TRUE or FALSE.", call. = FALSE)
@@ -288,7 +272,7 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
         }
 
         # get groups of taxonomy entries
-        tax_factors <- .get_tax_groups(x, col = col, onRankOnly = onRankOnly)
+        tax_factors <- .get_tax_groups(x, col = col, ...)
 
         # merge taxa
         x <- agglomerateByVariable(x, MARGIN = "rows", f = tax_factors, ...)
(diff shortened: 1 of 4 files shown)


R/agglomerate.R
```diff
@@ -253,6 +253,9 @@ setMethod("agglomerateByRank", signature = c(x = "SummarizedExperiment"),
             stop("'rank' must be a non-empty single character value",
                 call. = FALSE)
         }
+        if(!onRankOnly){
+            stop("'Use 'agglomerateByVariable' instead.", call. = FALSE)
```
TuomasBorman: Add "Deprecated. Use 'aggomerateByVariable..."

TuomasBorman:
You should add to parameter description that this parameter is deprecated. Moreover, what happens when we remove this parameter in the future --> it does not work anymore since the functionality is left in the code. This just blocks that it is not used anymore. The same variable is still there in the code.

I think you should

1. Remove onRankOnly completely from this file without deprecation.
2. In .get_tax_groups function, allow hidden parameter flow by adding ...
3. Add input check to .get_tax_groups
4. Pass parameters from agglomerateByRank to .get_tax_groups with ...

With that we can hide the functionality and prevent users to use it. Those that know that the option is available can still use it if necessary.

TuomasBorman:
This means also that no deprecation messages

tests/testthat/test-3agglomerate.R
```diff
@@ -91,20 +91,20 @@ test_that("agglomerate", {
...
-                 paste0("BCNMKCKTTVWYCKKMHTTMYTKKKYKTMMMKNKHDYKYMKDYKKNHNNNYMM",
-                        "KHHNDNNKTKMMMDNBHNBKKCTYMMCHNBNDDDNKSSHBNNRWDMYKKBNND",
-                        "NYTDRRKDVHNKNDRVGRNDRSBRRAWTBYNHRKKKWRSSRKKRAAWKSSKWR",
-                        "RWDWTNDBRVRRAMHHCMRDKKSSRARGSSVSYYHNYBRRVHNDNNHYKRMVV",
-                        "YKVRDNNNSRAARSBDKGGKK"))
+                 paste0("GCGAGCGTTATCCGGAATCATTGGGCGTAAAGGGTGCGTAGGCGGCGAGTTA",
+                        "AGTCTGAGGTAAAAGGTTGCAGCTCAACTGTAACAAGCCTTGGAAACTGACT",
+                        "AGCTAGAGTGCAGGAGAGGGCAGTGGAATTCCATGTGTAGCGGTAAAATGCG",
+                        "TAGATATATGGAGGAACACCAGTGGCGAAGGCGGCTGTCTGGCCTGTAACTG",
+                        "ACGCTGAGGCACGAAAGCGTGGGG"))
```
TuomasBorman: This result has changed? It should not be since the earlier onRankOnly default value was FALSE.

Probably if you use na.rm = FALSE it works with these sequences

tests/testthat/test-3agglomerate.R
```diff
@@ -91,20 +91,20 @@ test_that("agglomerate", {
...
+                        "AGTCGGATGTGAAATCCCCGGGCTCAACCTGGGAACTGCATTCGAAACTGGC",
+                        "AGGCTAGAGTCTTGTAGAGGGGGGTAGAATTCCAGGTGTAGCGGTGAAATGC",
+                        "GTAGAGATCTGGAGGAATACCGGTGGCGAAGGCGGCCCCCTGGACAAAGACT",
+                        "GACGCTCAGGTGCGAAAGCGTGGGG"))
     actual <- mergeFeaturesByRank(se,"Genus", mergeRefSeq = TRUE)
     expect_equal(as.character(referenceSeq(actual)[[1]]),
-                 paste0("BCNMKCKTTVWYCKKMHTTMYTKKKYKTMMMKNKHDYKYMKDYKKNHNNNYMM",
-                        "KHHNDNNKTKMMMDNBHNBKKCTYMMCHNBNDDDNKSSHBNNRWDMYKKBNND",
-                        "NYTDRRKDVHNKNDRVGRNDRSBRRAWTBYNHRKKKWRSSRKKRAAWKSSKWR",
-                        "RWDWTNDBRVRRAMHHCMRDKKSSRARGSSVSYYHNYBRRVHNDNNHYKRMVV",
```
TuomasBorman: Also can you move these tests that utilize data from miaTime package to the end of the tests? GHA has skipped all tests after this line since it does not install miaTime

tests/testthat/test-3agglomerate.R
0xMuluh: We had changed onRankOnly default to TRUE when we did the critical fix. I didn't catch the error back then since i hadn't miaTime. I will checkout the commit before the change and see if the sequence was valid then.

The sequence with na.rm TRUE/FALSE is also different.

0xMuluh: I think when we  ordered in alphabetical order it changed the sequence.

TuomasBorman: Instead of taking the first one, can you take one (by name) which is the correct taxon?

You can debug the function and check what is the first one before the sorting

0xMuluh: the first string before sorting is the older sequence.

```
> data(SilvermanAGutData)
> se <- SilvermanAGutData
> # checking reference consensus sequence generation
> actual <- mergeFeaturesByRank(se,"Genus", mergeRefSeq = FALSE)
Called from: .local(x, ...)
Warning message:
In mergeFeaturesByRank(se, "Genus", mergeRefSeq = FALSE) :
  'mergeFeaturesByRank' is deprecated. Use 'agglomerateByRank' instead.
Browse[1]> q <- as.character(referenceSeq(x)[[1]])
Browse[1]> q
[1] "TCAAGCGTTATCCGGATTTATTGGGTTTAAAGGGTGCGTAGGCGGTTTGATAAGTTAGAGGTGAAATCCCGGGGCTTAACTCCGGAACTGCCTCTAATACTGTTAGACTAGAGAGTAGTTGCGGTAGGCGGAATGTATGGTGTAGCGGTGAAATGCTTAGAGATCATACAGAACACCGATTGCGAAGGCAGCTTACCAAACTATATCTGACGTTGAGGCACGAAAGCGTGGGG"
Browse[1]> 
```

tests/testthat/test-3agglomerate.R
0xMuluh: ACK

tests/testthat/test-3agglomerate.R
TuomasBorman: Check the name of it and get its sequence in tests (instead of taking first one)

0xMuluh: I have allocated the new position to the former sequence.

tests/testthat/test-3agglomerate.R
```diff
@@ -86,20 +86,58 @@ test_that("agglomerate", {
...
+    expect_equal(nrow(actual), nrow(actual5))
+    
     # Load data from miaTime package
     skip_if_not(require("miaTime", quietly = TRUE))
     data(SilvermanAGutData)
     se <- SilvermanAGutData
     # checking reference consensus sequence generation
     actual <- mergeFeaturesByRank(se,"Genus", mergeRefSeq = FALSE)
-    expect_equal(as.character(referenceSeq(actual)[[1]]),
+    expect_equal(as.character(referenceSeq(actual)[[19]]),
```
TuomasBorman: Would it be possible to find with name? It is then easier to update in the future when we are searching specific name and not index

0xMuluh: Sure. It's updated now.

TuomasBorman (approved):

Issue #564, opened 2024-06-04
pseudocount


TuomasBorman:
If pseudocount=TRUE, the smallest positive value is added. Modify pseudocount so that pseudocount is half of the smallest positive value.

PR #571, opened 2024-06-05
Update pseudocount

0xMuluh:
See #564 
When pseudocount is TRUE, set it to half of non-zero minimum value.

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -422,7 +422,9 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
         }
         # If pseudocount TRUE, set it to half of non-zero minimum value, else set 
         # it to zero
-        pseudocount <- ifelse(pseudocount, min(mat[mat>0] / 2), 0)
+        # get min value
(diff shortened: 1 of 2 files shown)


R/transformCounts.R
```diff
@@ -420,8 +420,9 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
             stop("The assay contains missing or negative values. ",
                  "'pseudocount' must be specified manually.", call. = FALSE)
         }
-        # If pseudocount TRUE, set it to non-zero minimum value, else set it to zero
-        pseudocount <- ifelse(pseudocount, min(mat[mat>0]), 0)
+        # If pseudocount TRUE, set it to half of non-zero minimum value, else set 
+        # it to zero
+        pseudocount <- ifelse(pseudocount, min(mat[mat>0] / 2), 0)
```
TuomasBorman: This is correct, but it would be more clear that you first take the minimum positive value and then take half of it.

TuomasBorman (approved):

PR #574, opened 2024-06-08
WIP: Import taxpasta

Midnighter:
To do:

- [x] expand documentation
- [x] I need to import some biomformat internals, which I can only do with `biomformat:::`. I see no way around this at the moment.


R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+
+    # We read our own HDF5 array to later be able to read observation group
+    # metadata, which [biomformat::read_biom()] currently doesn't do.
+    raw <- h5read(filename, "/", read.attributes = TRUE)
+    biom <- .createBiom(raw)
+
+    # Without taxonomic information, we return a simple TreeSE.
+    if (is.null(raw$observation$`group-metadata`$ranks)) {
+        warning("The BIOM file does not contain taxonomy information; unable to generate a taxonomic tree.")
+        return(makeTreeSEFromBiom(biom))
```
TuomasBorman: importBIOM is the new name of the function

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    # We read our own HDF5 array to later be able to read observation group
+    # metadata, which [biomformat::read_biom()] currently doesn't do.
+    raw <- h5read(filename, "/", read.attributes = TRUE)
+    biom <- .createBiom(raw)
+
+    # Without taxonomic information, we return a simple TreeSE.
+    if (is.null(raw$observation$`group-metadata`$ranks)) {
+        warning("The BIOM file does not contain taxonomy information; unable to generate a taxonomic tree.")
```
TuomasBorman: Split long lines to multiple lines. 80 characters is the max (Bioconductor guidelines). Run BiocCheck::BiocCheck() and fix things related to added lines

TuomasBorman: Also add warning(call. = FALSE)

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+#' Create BIOM object from raw HDF5 array
+#'
+#' `.createBiom()` is an internal function used by [importTaxpasta()], that more or
+#' less replicates [biomformat::read_hdf5_biom()].
+#'
+#' @param h5array A raw HDF5 array read from a BIOM file.
+#'
+#' @return A [biomformat::biom()] object.
+#'
+.createBiom <- function(h5array) {
```
TuomasBorman: internal functions should be named as .create_biom

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+    date <- attr(h5array, "creation-date")
+    matrix_type <- "dense"
+    matrix_element_type <- "int"
+
+    return(
+        biomformat:::namedList(
+            id, format, format_url, type, generated_by, date, matrix_type,
+            matrix_element_type, rows, columns, shape, data
+        ) |> biomformat:::biom()
+    )
```
TuomasBorman: Create the object first and then return. It is easier to debug

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+#'
+#' @return A [TreeSummarizedExperiment::TreeSummarizedExperiment()] object.
+#' @export
+#'
+#' @examples
+#' tse <- importTaxpasta("profiles.biom")
+#'
+#' @importFrom rhdf5 h5read
+#' @importFrom SummarizedExperiment rowData
+#' @importFrom SingleCellExperiment altExps
```
TuomasBorman: Add NULL here as a last line after documentation. Then export functions "manually" by adding @export before the function description (check other functions/files).

Now there are also manual pages for internal functions which we do not want. 

Also, direct this function's documentation to same file where all importers are located. Check for example documentation of importBIOM

tests/testthat/data/biom_test_data.py
TuomasBorman, line 1: I am not sure what is the correct place for these files and scripts. That should be checked. Bioc guidelines might have info https://contributions.bioconductor.org/

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
+#' Import taxpasta-specific BIOM results to `TreeSummarizedExperiment`
+#'
+#' @param filename A character vector with one element that denotes the file
+#'   path to a BIOM file.
+#'
+#' @return A [TreeSummarizedExperiment::TreeSummarizedExperiment()] object.
+#' @export
+#'
+#' @examples
+#' tse <- importTaxpasta("profiles.biom")
```
TuomasBorman: When the package is built, manual pages are also run. These example does not work since there are no "profiles.biom" file. You can disable the run bu \dontrun (or something like that)

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+#' `.getRanks()` is an internal function used by [importTaxpasta()], that reads the
+#' string of semi-colon separated ranks from the group observation metadata.
+#'
+#' @param h5array A raw HDF5 array read from a BIOM file.
+#'
+#' @return A character vector of taxonomic ranks in order.
+.getRanks <- function(h5array) {
+    ranks <- strsplit(h5array$observation$`group-metadata`$ranks, ";", fixed = TRUE)[[1]]
+    return(ranks)
+}
```
TuomasBorman: This function is little bit overkill since the functionality itself is only one-liner. You could add that oneliner directly to code, and remove this function

R/importTaxpasta.R
```diff
@@ -0,0 +1,116 @@
...
+#' @return A [data.frame()] with observation metadata.
+#'
+#' @importFrom biomformat observation_metadata
+.createRowData <- function(biom, ranks) {
+    meta <- observation_metadata(biom)
+    column.names <- colnames(meta)
+    indeces <- startsWith(column.names, "taxonomy")
+
+    if (sum(indeces) != length(ranks)) {
+        stop("The number of generic taxonomy* columns differs from the number of ranks.")
```
TuomasBorman: call. = FALSE

man/mia-package.Rd
TuomasBorman, line 1: Why is this removed

TuomasBorman (changes requested):
Looks good, let me know if you need help with the documentation

R/importTaxpasta.R
TuomasBorman: Also, related to coding-style I would rather have one return() call per function. So instead of returning the object inside the if, you would add the taxonomy.

But that is more "matter of taste" kind of thing. (Sorry for little bit cryptic review, it is late here, and I wanted to comment so that I don't forget to do so)

R/importTaxpasta.R
Midnighter: I thought that `importBIOM` is for handling a file input, here `biom` is already a loaded BIOM object. At least, that's what the documentation suggests.

R/importTaxpasta.R
Midnighter: I've split lines and added `call. = FALSE`.

I disagree with you preferring a single return. The early return is a form of a guard clause which is good practice. It makes the code easier to read and reason about. See, for example, https://en.wikipedia.org/wiki/Guard_(computer_science) and https://www.oreilly.com/library/view/tidy-first/9781098151232/ch01.html. Maybe a relevant excerpt from the latter reference:

> (But what about MuLTipLe ReTuRns? The “rule” about having a single return for a routine came from the days of FORTRAN, where a single routine could have multiple entry and exit points. It was nearly impossible to debug such code. You couldn’t tell what statements were executed. Code with guard clauses is easier to analyze because the preconditions are explicit.)

R/importTaxpasta.R
Midnighter: That's a confusing style, but easy to adapt.

R/importTaxpasta.R
Midnighter: The only function that I want to export is `importTaxpasta` and it has the `@export` statement.

> Now there are also manual pages for internal functions which we do not want.

Right, I will exclude them.

> Also, direct this function's documentation to same file where all importers are located. Check for example documentation of importBIOM

I'll look into that.

tests/testthat/data/biom_test_data.py
Midnighter: If you are unhappy with the current placement, perhaps https://contributions.bioconductor.org/data.html#raw-data-and-the-instextdata-directory is an option?

Midnighter: Installing the data would then also allow using it in the examples.

Midnighter: On that note, what's the reasoning for mia using `LazyData: false`? It really takes a long time to load the package, partially because of that, I guess.

R/importTaxpasta.R
Midnighter: I find that the function name makes the purpose of the line much clearer.

man/mia-package.Rd
Midnighter: I must have made a mistake when generating documentation.

Midnighter: Oh, it's probably related to the following error when I generate documentation. I have no idea why that error occurs or how to fix it. The description seems correct to me.

```
Error in parse(text = dcf[, "Authors@R"], keep.source = TRUE) : 
  <text>:24:22: unexpected numeric constant
23: email = "midnighter@posteo.net", role = c("ctb),
24: comment = c(ORCID = "0000
                         ^
```

Midnighter:
Thank you for the detailed review @TuomasBorman! I have some open questions from your comments and will appreciate your assistance with the documentation. It's not clear to me why to break up the docstring like that for a single function and how to direct `importTaxpasta`'s docstring to the general file.

tests/testthat/data/biom_test_data.py
antagomir: inst/extdata and inst/extras and inst/scripts are valid options I think; perhaps test repo can also have its own data (according to guidelines) but I am not sure

R/importTaxpasta.R
antagomir: or, if the file is inserted in one of the inst/ subfolders then it can be called in examples with the system.file function (see other R functions that use this)

R/importTaxpasta.R
antagomir: I am ok with either solution but two points:

1) If this could be also used elsewhere then it might make sense to have the functions, just to make sure this is done in a standardized way every time.
2) if added straight in the code then there should be clear code comment explaining the purpose

Midnighter:
Hi @TuomasBorman,

Starting next week, I will have more time again and could try to complete this PR. I would need some advice and decisions from your side, though, please. If you can take another look, I will be grateful.

TuomasBorman:
> Hi @TuomasBorman,
> 
> Starting next week, I will have more time again and could try to complete this PR. I would need some advice and decisions from your side, though, please. If you can take another look, I will be grateful.

Alright, sounds good. I have also been busy, but I will check during this week.

R/importTaxpasta.R
```diff
@@ -0,0 +1,134 @@
+#' Import taxpasta-specific BIOM results to `TreeSummarizedExperiment`
+#'
+#' @param filename A character vector with one element that denotes the file
+#'   path to a BIOM file.
+#'
+#' @return A [TreeSummarizedExperiment::TreeSummarizedExperiment()] object.
```
TuomasBorman: Add link like this \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}

R/importTaxpasta.R
TuomasBorman: That would be even better solution.

Data goes to inst/extdata and python script goes to inst/scripts

R/importTaxpasta.R
TuomasBorman: Here is an example: https://github.com/EBI-Metagenomics/HoloFoodR/blob/devel/R/getMetaboLights.R

By using NULL and @export just above the function, you can specify which functions are exported and listed in the documentation.

R/importTaxpasta.R
```diff
@@ -0,0 +1,134 @@
...
+#'
+#' @examples
+#' \dontrun{
+#' tse <- importTaxpasta("profiles.biom")
+#' }
+#'
+#' @importFrom rhdf5 h5read
+#' @importFrom SummarizedExperiment rowData
+#' @importFrom SingleCellExperiment altExps
+importTaxpasta <- function(filename) {
```
TuomasBorman: The parameter name can be just "file" as here: https://github.com/microbiome/mia/blob/devel/R/importHumann.R

TuomasBorman: -> harmonize between functions

R/importTaxpasta.R
TuomasBorman: You are correct. Sorry for misinformation

tests/testthat/test-importTaxpasta.R
```diff
@@ -0,0 +1,60 @@
...
+
+test_that("the TreeSummarizedExperiment has one tree with four inner nodes", {
+  tree <- TreeSummarizedExperiment::rowTree(result)
+  expect_identical(ape::Nnode(tree), 4L)
+})
+
+test_that("the TreeSummarizedExperiment has one tree with five edges", {
+  tree <- TreeSummarizedExperiment::rowTree(result)
+  expect_identical(ape::Nedge(tree), 5L)
+})
```
TuomasBorman: SCE and TreeSE packages are already loaded so use them without specifying the package

tests/testthat/test-importTaxpasta.R
```diff
@@ -0,0 +1,60 @@
...
+result <- suppressWarnings({
+  importTaxpasta(test_path("data", "complete.biom"))
+})
+
+test_that("loading a taxpasta BIOM file returns a TreeSummarizedExperiment", {
+  expect_s4_class(result, "TreeSummarizedExperiment")
+})
+
+requireNamespace("ape", quietly = TRUE)
+requireNamespace("TreeSummarizedExperiment", quietly = TRUE)
```
TuomasBorman: These calls are not necessary as they should be already loaded with mia package

NAMESPACE
```diff
@@ -325,6 +326,7 @@ importFrom(ape,has.singles)
 importFrom(ape,is.rooted)
 importFrom(ape,read.tree)
 importFrom(ape,root)
+importFrom(biomformat,observation_metadata)
```
TuomasBorman: biomformat is not dependency of mia. This means that biomformat is not installed automatically. This leads to an error.

Instead, you could add .require_package("biomformat") in the beginning of function and specify biomformat functions like biomformat::observation_metadata()

TuomasBorman (changes requested):
Looks good already

tests/testthat/data/biom_test_data.py
TuomasBorman: inst/extdata and inst/scripts

For lazydata:https://contributions.bioconductor.org/description.html?q=lazydata#description-lazydata

man/mia-package.Rd
TuomasBorman: It seems now correct

man/importTaxpasta.Rd
TuomasBorman, line 1: Also the manual page seems to be correct now

tests/testthat/data/biom_test_data.py
Midnighter: That reasoning regarding lazydata is counter-intuitive, but fair enough.

R/importTaxpasta.R
Midnighter: It seems to me the same can be achieved by specifying on the helper functions.

```r
#' @noRd
```

R/importTaxpasta.R
Midnighter: That will shadow the `base::file` built-in function, but no problem.

Midnighter:
@TuomasBorman, I believe, I have addressed all of your comments. Please have another look and let me know if there's something else to be addressed.

TuomasBorman (approved):
Looks good!

PR #584, opened 2024-06-17
Deprecate z transform in transformAssay

0xMuluh:
This commit does the following:
1-  Deprecates the z method in transformAssay
2- Updates the documentation to  only reference transformation methods already present in vegan::decostand.

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -52,22 +52,8 @@
 #'
 #' \itemize{
 #' 
-#' \item 'alr': Additive log ratio (alr) transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#' 
-#' \item 'chi.square': Chi square transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#' 
-#' \item 'clr': Centered log ratio (clr) transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#'
-#' \item 'frequency': Frequency transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#' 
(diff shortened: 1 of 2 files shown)


R/transformCounts.R
```diff
@@ -310,6 +268,10 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
     MARGIN <- ifelse(MARGIN %in% c("samples", "columns", "col", 2), 2, 1)
     # Adjust method if mia-specific alias was used
     method <- ifelse(method == "relabundance", "total", method)
+    
+    if (method == "z") {
+        .Deprecated(old="z", new="standardize", "Now 'z' is deprecated. Use 'standardize' instead.")
```
TuomasBorman: This is enough .Deprecated(old="z", new="standardize")

R/transformCounts.R
```diff
@@ -52,22 +52,8 @@
...
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#'
-#' \item 'frequency': Frequency transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#' 
-#' \item 'hellinger':  Hellinger transformation, please refer to 
-#' \code{\link[vegan:decostand]{decostand}} for details.
-#' 
-#' \item 'log':  Logarithmic transformation, please refer to 
+#' \item \code{alr, chi.square, clr, frequency, hellinger, log, normalize, pa, rank, rclr
```
TuomasBorman: The code formats the values incorrectly (it shows that the list is one value). Instead, you have to format values one by one or just like 'alr', 'frequency', ...

TuomasBorman (approved):

antagomir:
This should be now updated to OMA as well, is this already on the way - just checking? @Daenarys8

Issue #577, opened 2024-06-12
prevalence functions and as_relative parameter


TuomasBorman:
Currently, for example agglomerateByPrevalence functions has as_relative parameter that applies relative transformation to assay. However, this might be confusing. --> Remove as_relative option and calculate the prevalence directly from the assay specified by user.

PR #589, opened 2024-06-19
remove as.relative param

0xMuluh:
Referenced from #577. This PR removes as.relative from codebase and updates examples and tests. Tests that tested as.relative functionality are also removed.


R/agglomerate.R
```diff
@@ -53,14 +53,10 @@
 #'        \item \code{make.unique}: A single boolean value for selecting
 #'        whether to make rownames unique. (By default: \code{make.unique = TRUE})
 #'        \item \code{detection}: Detection threshold for absence/presence.
-#'        Either an absolute value compared directly to the values of \code{x}
-#'        or a relative value between 0 and 1, if \code{as.relative = FALSE}.
+#'        It is an absolute value compared directly to the values of \code{x}
```
TuomasBorman: This description could be improved

R/getPrevalence.R
```diff
@@ -273,19 +260,12 @@ setMethod("getPrevalence", signature = c(x = "ANY"), function(
 #' @export
 setMethod("getPrevalence", signature = c(x = "SummarizedExperiment"),
     function(x, assay.type = assay_name, assay_name = "counts",
-            as.relative = as_relative, as_relative = FALSE, rank = NULL, ...){
-        # input check
-        if(!.is_a_bool(as.relative)){
-            stop("'as.relative' must be TRUE or FALSE.", call. = FALSE)
-        }
```
TuomasBorman, line 281: We probably cannot just remove the functionality. One option is to move the functionality to hidden functionality (move the relative cabundance calculation to internal function).

TuomasBorman (changes requested):

R/getPrevalence.R
0xMuluh: I am not sure I got you correctly because we have `.calc_rel_abund() `in internal functions for calculating relative abundance.

TuomasBorman: My point was that we cannot remove as.relative straightaway. Instead, we should deprecate it. The easiest solution probably would be to hide the option for as.relative (so that user cannot see that from manuals)

0xMuluh: Ahh, I hid the functionality in an internal function `.get_abundance()`

R/utils.R
```diff
@@ -153,6 +153,23 @@
...
+.get_abundance <- function(x, assay.type = assay_name, assay_name = "counts",
+                           rank = NULL, as.relative = FALSE, ...) {
+    # input check
+    if(!.is_a_bool(as.relative)){
+        stop("'as.relative' must be TRUE or FALSE.", call. = FALSE)
+    }
+    # check assay
+    .check_assay_present(assay.type, x)
+    x <- .agg_for_prevalence(x, rank = rank, ...)
+    mat <- assay(x, assay.type)
```
TuomasBorman: Move these lines back to original place. They are not directly related to this

R/estimateDominance.R
```diff
@@ -294,7 +294,7 @@ setMethod("estimateDominance", signature = c(x = "SummarizedExperiment"),
 }
 
 .calc_core_dominance <- function(mat, ...){
-    getPrevalentAbundance(mat, detection = 0, as.relative = TRUE)
```
TuomasBorman: We probably have to keep this as.relative=TRUE here

R/utils.R
```diff
@@ -153,6 +153,23 @@
     return(MARGIN)
 }
 
+# Get abundance. Determines if relative abundance is calculated or not.
+.get_abundance <- function(x, assay.type = assay_name, assay_name = "counts",
```
TuomasBorman: The name of the function could be .convert_to_relative_abundance or somethign

R/estimateDominance.R
TuomasBorman: Or  pass ... and check that it works

R/getPrevalence.R
```diff
@@ -273,19 +260,14 @@ setMethod("getPrevalence", signature = c(x = "ANY"), function(
...
-
+            rank = NULL, ...){
         # check assay
         .check_assay_present(assay.type, x)
         x <- .agg_for_prevalence(x, rank = rank, ...)
         mat <- assay(x, assay.type)
-        if (as.relative) {
-            mat <- .calc_rel_abund(mat)
-        }
+
```
TuomasBorman: Remove extra line break

tests/testthat/test-5prevalence.R
```diff
@@ -9,30 +9,16 @@ test_that("getPrevalence", {
...
-
-    # Same prevalences should be returned for as.relative T/F in certain cases.
-    pr1 <- getPrevalence(GlobalPatterns, detection=1, include.lowest=TRUE, as.relative=FALSE)
-    pr2 <- getPrevalence(GlobalPatterns, detection=0/100, include.lowest=FALSE, as.relative=TRUE)
-    expect_true(all(pr1 == pr2))
-
-    # Same prevalences should be returned for as.relative T/F in certain cases.
-    pr1 <- getPrevalence(GlobalPatterns, detection=1, include.lowest=TRUE, as.relative=FALSE)
-    pr2 <- getPrevalence(GlobalPatterns, detection=0, include.lowest=FALSE, as.relative=FALSE)
-    expect_true(all(pr1 == pr2))
```
TuomasBorman: You could leave these checks as the functionality is still there

R/getPrevalence.R
```diff
@@ -148,32 +139,28 @@
...
+#'                             prevalence = 50/100)
 #' head(prevalent)
 #'
 #' # Gets a subset of object that includes prevalent taxa
 #' altExp(tse, "prevalent") <- subsetByPrevalent(tse,
 #'                                              rank = "Family",
 #'                                              detection = 0.001,
-#'                                              prevalence = 0.55,
-#'                                              as.relative = TRUE)
+#'                                              prevalence = 0.55)
```
TuomasBorman: Now this detection threshold is wrong (the values are counts, but this is relative). You could add

assay.type = "relabundance"

tests/testthat/test-5prevalence.R
```diff
@@ -246,19 +212,19 @@ test_that("getRare", {
...
 
     # Test lias
-    alias <- getRare(gp_null, detection=0.0045, prevalence = 0.25, as.relative=TRUE)
+    alias <- getRare(gp_null, detection=0.0045, prevalence = 0.25)
     expect_equal(pr1, alias)
 
     pr1 <- getRare(GlobalPatterns, detection=0.004, prevalence = 0.1,
-                       as.relative=TRUE, rank = "Family")
+               rank = "Family")
     pr2 <- getRare(gp_null, detection=0.004, prevalence = 0.1,
```
TuomasBorman: Convert these to more sensible threshold. 0.0045 is not sensible detection threshold for counts

tests/testthat/test-5prevalence.R
```diff
@@ -385,31 +325,17 @@ test_that("agglomerateByPrevalence", {
...
-                                           detection = 0.0001,
-                                           prevalence = 50/100,
-                                           as.relative = TRUE,
-                                           other.label = "test"),
-                 agglomerateByPrevalence(GlobalPatterns,
-                                           rank = NULL,
-                                           detection = 0.0001,
-                                           prevalence = 50/100,
-                                           as.relative = TRUE,
-                                           other.label = "test"))
```
TuomasBorman: Does this work with

agglomerateByPrevalence(tse, assay.type = "relabundance", detection = 0.001, prevalence = 0.5)

Are the prevalences calculated based on relative abundance assay? If not, then we have a problem.

TuomasBorman (changes requested):
Check if the assay.type parameter is passed from agglomerateByPrevalence to getPrevalence.

After that, it might be best to just apply changes in getPrevalence.R file. We could prefer to not use `as.relative` but still keep supporting it (not remove the parameter from the documents even though that was the initial idea).

tests/testthat/test-5prevalence.R
0xMuluh: I have restored the tests since as.relative is still available. Should the threshold be modified still?

TuomasBorman: Is there tests that test the absolute values? If not, then there should be.

detection threshold should reflect the type of values. If there are counts (and as.relative=FALSE), then the values should be integer. If the assay.type="relabundance" or as.relative=TRUE, then detection threshold should be value 0-1

tests/testthat/test-5prevalence.R
0xMuluh: Hmm, no, unless after `tse <- transformAssay(tse, assay.type = "counts", method = "relabundance")`

TuomasBorman: Yes, that is something that must be done first

0xMuluh: `agglomerateByPrevalence(tse, assay.type = "relabundance", detection = 0.001, prevalence = 0.5)` works fine and `assay.type `is parsed in `getPrevalence`. is this what I needed to investigate or perhaps some comparison is needed?

TuomasBorman: That is good, thanks

TuomasBorman (approved):

PR #590, opened 2024-06-20
import functions and ranks

TuomasBorman:
importBIOM did not detect superkingdom rank. Also it was not possible to set taxonomy ranks automatically. These issues are now fixed.

Also, the supported ranks come from generic function to ensure that support can be extended easily in every function at once.

This is a draft, still checks to do.


TuomasBorman:
@Daenarys8 Can you check that everything works?

0xMuluh (commented):
LGTM. Imported some data files and it worked fine. Couldn't get dataset with d_ or sk_ for reliable checks. However;
- both `.parse_metaphlan` and `.parse_taxonomy` rely on `TAXONOMY_RANKS` which does not take into account `sk_` and `t_` . 
- can we also use `.get_lowest_taxonomic_level` in `.parse_taxonomy`?

R/utils.R
```diff
@@ -460,13 +457,11 @@
...
+    colnames(taxa_tab) <- names(all_ranks)
     
-    # Subset columns so that it includes TAXONOMY_RANKS columns by default.
-    # If strain column has values, it it also returned.
-    ind <- !(!tolower(colnames(taxa_tab)) %in% tolower(returned.ranks) &
-        colSums(is.na(taxa_tab)) == nrow(taxa_tab))
-    taxa_tab <- taxa_tab[ , ind, drop = FALSE]
+    # Subset columns so that include only those columns that have some value
+    non_empty <- colSums(is.na(taxa_tab)) != nrow(taxa_tab)
+    taxa_tab <- taxa_tab[ , non_empty, drop = FALSE]
```
TuomasBorman, line 464: The output had "all" the taxonomy ranks even though the data did not include them. Now only those ranks are returned that were detected from the data.

TuomasBorman:
Thanks for pointing out these problems. This requires little bit more thinking.

NAMESPACE
```diff
@@ -1,7 +1,6 @@
 # Generated by roxygen2: do not edit by hand
 
 export(IdTaxaToDataFrame)
-export(TAXONOMY_RANKS)
```
TuomasBorman, line 4: TAXONOMY_RANKS is not visible anymore for user. It is still used in the functions to determine taxonomy ranks. Moreover, user can still modify the value of it:

```
> getTaxonomyRanks()
 [1] "domain"       "superkingdom" "kingdom"      "phylum"       "class"        "order"       
 [7] "family"       "genus"        "species"      "strain"      
> setTaxonomyRanks(c("one", "two"))
> getTaxonomyRanks()
[1] "one" "two"
```

R/convertFromBIOM.R
```diff
@@ -230,18 +225,33 @@ convertFromBIOM <- function(
         assays = assays,
         colData = sample_data,
         rowData = feature_data)
+    
+    # Set ranks based on rowData columns if user has specified to do so
+    temp <- .set_ranks_based_on_rowdata(tse, ...)
```
TuomasBorman, line 228: Now importBIOM supports also setting new ranks based on data. In theory, the file can include more ranks that are supported by mia

R/convertFromBIOM.R
```diff
@@ -230,18 +225,33 @@ convertFromBIOM <- function(
...
+    
+    # Remove prefixes if specified and rowData includes info
+    if(prefix.rm && ncol(rowData(tse)) > 0){
+        rowData(tse) <- .remove_prefixes_from_taxa(rowData(tse), ...)
+    }
+
```
TuomasBorman, line 234: To set ranks correctly, this was moved to the end

R/convertFromBIOM.R
```diff
@@ -230,18 +225,33 @@ convertFromBIOM <- function(
...
     return(tse)
 }
 
 ################################ HELP FUNCTIONS ################################
 # This function removes prefixes from taxonomy names
 .remove_prefixes_from_taxa <- function(
-        feature_tab, prefixes = "sk__|([dkpcofgs]+)__",
+        feature_tab,
+        prefixes = paste0(
+            "(", paste0(.get_all_supported_ranks(), collapse = "|"), ")__"),
```
TuomasBorman: Prefixed are removed based on all supported ranks (shared in mia; not listed individually in every function).

R/convertFromBIOM.R
```diff
@@ -274,16 +286,9 @@ convertFromBIOM <- function(
...
-        "^p__",
-        "^c__",
-        "^o__",
-        "^f__",
-        "^g__",
-        "^s__"
-    )
+    all_ranks <- .get_all_supported_ranks()
+    prefixes <- paste0("^", all_ranks, "__")
+    names(prefixes) <- names(all_ranks)
```
TuomasBorman: Here also: taxonomy ranks are detected based on prefixes of all supported ranks in mia.

R/convertFromBIOM.R
```diff
@@ -292,7 +297,7 @@ convertFromBIOM <- function(
     # If only one prefix was found (like it should be), get the corresponding
     # rank name.
     if( sum(found_rank) == 1 ){
-        colname <- TAXONOMY_RANKS[found_rank]
+        colname <- names(prefixes)[found_rank]
```
TuomasBorman, line 298: There was a bug that is now fixed. Colnames were set based on index of listed prefix and TAXONOMY_RANKS but they did not match (there were domain prefix that was not in TAXONOMY_RANKS).

R/importMetaphlan.R
```diff
@@ -124,10 +124,8 @@ importMetaPhlAn <- function(
     tables <- .parse_metaphlan(data, ...)
 
     # Create multiple SE objects at different rank from the data
-    available_ranks <- names(tables)
     se_objects <- lapply(tables, function(x){
-        .create_se_from_metaphlan(
-            x, rowdata_col, returned.ranks = available_ranks, ...)
+        .create_se_from_metaphlan(x, rowdata_col, ...)
         })
```
TuomasBorman, line 130: `returned.ranks` controls which taxonomy ranks are included in rowData(tse). There is no really user-case for controlling it since we want all the data.

TuomasBorman: Also, I believe controlling it does not affect the results since the ranks should already match because they are got from the metaphlan file...

R/importMetaphlan.R
```diff
@@ -231,8 +229,8 @@ importMetaPhlAn <- function(
     # at specific rank
     tables <- split(table, levels)
     # Get the order
-    metaphlan_tax = c(TAXONOMY_RANKS, "strain")
-    indices <- match(metaphlan_tax, tolower(names(tables)))
+    metaphlan_tax = names(.get_all_supported_ranks())
+    indices <- match(tolower(metaphlan_tax), tolower(names(tables)))
```
TuomasBorman: Now all the supported ranks are detected from Metaphlan files.

R/importMetaphlan.R
```diff
@@ -243,17 +241,17 @@ importMetaPhlAn <- function(
 # Get the lowest level of the string that contains multiple taxonomic levels with prefixes
 # Output is single character that specifies the rank, e.g, "s" == "Species"
 .get_lowest_taxonomic_level <- function(string){
+    # List all ranks and what prefix they correspond
+    ranks <- .get_all_supported_ranks()
     # Get indices that specify location of rank prefixes 
-    levels <- gregexpr("([kpcofgst]+)__", string)[[1]]
+    ranks_pattern <- paste(ranks, collapse = "|")
+    ranks_pattern <- paste0("(", ranks_pattern, ")__")
+    levels <- gregexpr(ranks_pattern, string)[[1]]
```
TuomasBorman: Same here: the supported ranks are controlled from same place

R/taxonomy.R
```diff
@@ -137,12 +137,26 @@
...
+        phylum = "p",
+        class = "c",
+        order = "o",
+        family = "f",
+        genus = "g",
+        species = "s",
+        strain = "t"
+    )
+    return(ranks)
+}
```
TuomasBorman: This is the place where to control the ranks that are supported by default.

R/taxonomy.R
```diff
@@ -137,12 +137,26 @@
...
+#' @format a \code{character} vector of length containing all the taxonomy ranks
 #'   recognized. In functions this is used as case insensitive.
-#' @export
-TAXONOMY_RANKS <- c("domain","kingdom","phylum","class","order","family",
-                    "genus","species")
+TAXONOMY_RANKS <- names(.get_all_supported_ranks())
```
TuomasBorman: By default, TAXONOMY_RANKS equals to all supported ranks but user can control this. TAXONOMY_RANKS does affect for instance when hierarchy tree is calculated. However, it does not affect the importing functions. This is because, importing includes detection of ranks, and detection is made based on prefix which TAXONOMY_RANKS do not include. Moreover, I think it is best to keep some control in importing.

R/utils.R
```diff
@@ -410,8 +410,8 @@
 #' @noRd
 .parse_taxonomy <- function(
     taxa_tab, sep = "; |;", col.name = column_name, column_name = "Taxon",
-    remove.prefix = prefix.rm, prefix.rm = removeTaxaPrefixes, removeTaxaPrefixes = FALSE,
-    returned.ranks = TAXONOMY_RANKS, ...) {
```
TuomasBorman, line 412: returned.ranks control which taxonomy ranks are returned in taxonomy table. I think it is best to return all ranks that have data.

R/utils.R
```diff
@@ -428,16 +428,12 @@
...
-    }
     ############################## Input check end #############################
     
     #  work with any combination of taxonomic ranks available
-    all_ranks <- c(
-      "Kingdom","Phylum","Class","Order","Family","Genus","Species", "Strain")
-    all_prefixes <- c("k__", "p__", "c__", "o__", "f__", "g__", "s__", "t__")
+    all_ranks <- .get_all_supported_ranks()
+    all_prefixes <- paste0(all_ranks, "__")
+    names(all_prefixes) <- names(all_ranks)
```
TuomasBorman: Detect all supported ranks

TuomasBorman (commented):
@Daenarys8 Can you check that everything works as expected? Some checks are failing, they should be updated.

Check which checks fail and then check comments. Update the check only then when you are sure that the change is intended.

R/taxonomy.R
```diff
@@ -137,12 +137,43 @@
...
+    }
+    # Replace default value of mia:::.all_supported_ranks
+    assignInMyNamespace(".all_supported_ranks", prefixes)
+}
+
+# Function to get taxonomy ranks prefixes (not exported)
+getTaxonomyRankPrefixes <- function() {
+    return(.all_supported_ranks)
+}
+
```
TuomasBorman: Added functions to control supported ranks. They are not visible for user.

R/taxonomy.R
```diff
@@ -137,12 +137,43 @@
...
+    kingdom = "k",
+    phylum = "p",
+    class = "c",
+    order = "o",
+    family = "f",
+    genus = "g",
+    species = "s",
+    strain = "t"
+)
+
```
TuomasBorman: All supported ranks currently. TAXONOMY_NAMES == names(.all_supported_ranks)

tests/testthat/test-4IO.R
```diff
@@ -493,8 +495,8 @@ test_that("Import HUMAnN file", {
     file_path <- system.file("extdata", "humann_output.tsv", package = "mia")
     tse <- importHUMAnN(file_path)
     #
-    expect_true( length(taxonomyRanks(tse)) == 7 )
-    expect_true( ncol(rowData(tse)) == 9 )
+    expect_true( length(taxonomyRanks(tse)) == 2 )
+    expect_true( ncol(rowData(tse)) == 4 )
```
TuomasBorman, line 499: Now only levels that have information is returned. Previously the function returned always the whole taxonomy (Kingdom, Phylum,... Species). Now only Genus and Species are returned since those are only ranks found from data.

PR #591, opened 2024-06-24
Rename converters

theotimeee:
This PR makes the following naming modifications:

`makeTreeSEFromDADA2` --> `convertFromDADA2`
`makeTreeSEFromBIOM` --> `convertFromBIOM`
`makeTreeSEFromPhyloseq` --> `convertFromPhyloseq`
`makePhyloseqFromTreeSE` --> `convertToPhyloseq`

and deprecate the following functions:
`makeTreeSummarizedExperimentFromDADA2`
`makeTreeSummarizedExperimentFromPhyloseq`
`makeTreeSummarizedExperimentFromBIOM`
`makePhyloseqFromTreeSummarizedExperiment`


pkgdown/_pkgdown.yml
```diff
@@ -12,14 +12,11 @@ reference:
   - transformAssay
 - subtitle: Data loading
 - contents:
-  - makeTreeSummarizedExperimentFromPhyloseq
-  - makeTreeSummarizedExperimentFromDADA2
   - importBIOM
   - importQIIME2
   - importMothur
   - importMetaPhlAn
   - importHUMAnN
```
TuomasBorman, line 19: Add convert page

NAMESPACE
```diff
@@ -75,8 +79,6 @@ export(loadFromHumann)
 export(loadFromMetaphlan)
 export(loadFromMothur)
 export(loadFromQIIME2)
-export(makePhyloseqFromTreeSE)
-export(makePhyloseqFromTreeSummarizedExperiment)
```
TuomasBorman: These should be added to deprecate.R

man/convert.Rd
```diff
@@ -0,0 +1,208 @@
+% Generated by roxygen2: do not edit by hand
+% Please edit documentation in R/convertFromBIOM.R, R/convertFromDADA2.R,
+%   R/convertFromPhyloseq.R, R/convertToPhyloseq.R
+\name{convert}
+\alias{convert}
+\alias{importBIOM}
```
TuomasBorman: importBIOM is in wrong place.

Otherwise this PR looks good

theotimeee: Should I make a new man page for importBIOM only?

TuomasBorman: Do we have other importers in same man page? If there are general man page for importers, add it there. Otherwise, create new one for importBIOM

theotimeee: Each importer has its specific man page so I am going to create a new one for importBIOM

TuomasBorman: Ok! If that is the case, can you check what is the best approach (should functions have own man page or can they be merged into one bigger man page).

The initial idea was to highlight that these functions work similarly (for example agglomerate* functions). However, merging the man pages into one, have some disadvantages.

Also as we implemented the renaming, the synergy between functions should be evident

theotimeee: The main disadvantage I see is that many parameters under the current importer man pages have the same names (e.g. `file`) and I think we would need to describe the use of `file` parameter for all importers in one R file only.
It would then be more difficult to modify the documentation for one specific importer.

TuomasBorman: Alright, thanks. The documentation fixing was still a thing to do, so let's take this into account.

TuomasBorman (approved):
Great

TuomasBorman:
@thpralas Can you check other mia* packages and OMA that they reflect these changes?

Issue #482, opened 2024-01-12
More options from ggplot for plotAbundanceDensity


artur-sannikov:
`plotAbundanceDensity` function does not accepts arguments from ggplot, for example `jitter` layout does not have `width` option and uses `geom_point()` as its basis but it does not accept[ geom_point](https://ggplot2.tidyverse.org/reference/geom_point.html) options.

antagomir:
This issue should be moved to miaViz?

artur-sannikov:
You are write. It's possible to [transfer issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository) if you have write access.

PR #593, opened 2024-06-27
Fixes to mia (OMA issue #482) 

ake123:
Fixes to mia (OMA issue #482) 

- diversity functions - warning that tells that some indices where not calculated 
- documentation  fix on runRDA()


R/estimateDiversity.R
```diff
@@ -189,12 +189,17 @@
...
+#' valid_indices <- index[index %in% col_names]
+#' 
 #' # The colData contains the indices with their code names by default
-#' colData(tse)[, index]
+#' colData(tse)[, valid_indices]
 #'
 #' # Removing indices
-#' colData(tse)[, index] <- NULL
+#' colData(tse)[, valid_indices] <- NULL
+#' colData(tse)[, name] <- NULL
```
TuomasBorman: name was not yet used, so this line is extra (there cannot be index called name)

R/estimateDiversity.R
```diff
@@ -240,229 +248,273 @@ NULL
...
-                index = c("coverage", "fisher", "gini_simpson", 
-                        "inverse_simpson", "log_modulo_skewness", "shannon"),
+           function(x, assay.type = "counts", assay_name = NULL,
+                    index = c("coverage", "fisher", "gini_simpson", 
+                              "inverse_simpson", "log_modulo_skewness", "shannon"),
                     name = index, ...)
-                    standardGeneric("estimateDiversity"))
+               standardGeneric("estimateDiversity"))
 
 #' @rdname estimateDiversity
```
TuomasBorman: Does this make the indentation wrong?

R/estimateDiversity.R
```diff
@@ -240,229 +248,273 @@ NULL
...
+                       call. = FALSE)
+              }
+              .check_assay_present(assay.type, x)
+              .require_package("vegan")
+              
+              dvrsts <- BiocParallel::bplapply(index,
+                                               .get_diversity_values,
+                                               x = x,
+                                               mat = assay(x, assay.type),
+                                               BPPARAM = BPPARAM,
```
TuomasBorman: Only the indentation is changed. Also it seems that the indentation is not multipication of 4 spaces anymore

1. The indentation should be multiplication of 4
2. We do not want to change code like this. It looks like the whole code has changed. As people are already using the package, we must retain the code. It is extremely hard to see the history of code when there are this kind of changes that do not really affect the functionality. --> indentation is fixed only when the code/functionality is also changed

R/estimateDiversity.R
ake123: removed ...was made during some testing

R/estimateDiversity.R
ake123: i fixed it

R/estimateDiversity.R
ake123: Okay fixed !

ake123:
it only failed here https://github.com/microbiome/mia/actions/runs/9741095621/job/26879566844
It looks like some permission issue in the ubuntu

R/estimateDiversity.R
```diff
@@ -286,183 +293,226 @@ setMethod("estimateDiversity", signature = c(x="SummarizedExperiment"),
...
+                  mat <- mat[ !is.na(node.label), ]
+                  node.label <- node.label[ !is.na(node.label) ]
+                  # Rename
+                  rownames(mat) <- node.label
+              }
+              # Calculates Faith index
+              faith <- list(.calc_faith(mat, tree, ...))
+              # Adds calculated Faith index to colData
+              .add_values_to_colData(x, faith, name)
+          }
```
TuomasBorman: It seems that the indentation is still off. Also, there are many lines that are not really changed. Only the indentation is changed. That makes review harder, but also makes the history of functions less clear.

Sorry about this, but can you revert lines that do not have any "real" changes that are not related to the issue?

Also, the permission issue in GHA should resolve when pushed to devel

TuomasBorman (changes requested):

PR #594, opened 2024-06-28
Remove tree rooting in calculateUnifrac

0xMuluh:
This PR changes `calculateUnifrac` so it doesn't **change tree**(_root tree_) before calculating unifrac distance.


TuomasBorman:
calculateUnifrac works both for rows and columns. However, the implementation could be much simpler (now the same steps are done twice (once for rows and once for columns.

Can you modify the code so that you get first the functions and then run the steps only once. Similarly to this https://github.com/microbiome/mia/blob/c17fa9f70f556458247dda22112b14c9889efd3b/R/agglomerate.R#L452

R/calculateUnifrac.R
```diff
@@ -126,43 +126,41 @@ setMethod("calculateUnifrac",
...
             tree.name = tree_name, tree_name = "phylo", transposed = FALSE, ...){
+        # Get functions and parameters based on direction
+        tree_present_FUN <- switch(as.character(transposed), "TRUE" = .check_colTree_present, 
+            "FALSE" = .check_rowTree_present, stop("."))
+        tree_FUN <- switch(as.character(transposed), "TRUE" = colTree, "FALSE" = rowTree, stop("."))
+        links_FUN <- switch(as.character(transposed), "TRUE" = colLinks, "FALSE" = rowLinks, stop("."))
+        subsetting_FUN <- switch(as.character(transposed), "TRUE" = function(x, idx) x[, idx], 
+            "FALSE" = function(x, idx) x[idx, ], stop("."))
+        transpose_MAT <- switch(as.character(transposed), "FALSE" = function(mat) t(mat), 
+            "TRUE" = function(mat) mat, stop("."))
```
TuomasBorman: These are more clear to keep in the code as

```
if(transpose){
    mat <- t(mat)
}
```

The idea is to make the code easier to read

TuomasBorman: Also add input check for transpose

And this might be simpler

var <- FALSE
FUN <- if (var) print else cat

TuomasBorman: Also remove extra line breaks

TuomasBorman: I think this idea works; it makes the code more readable and decreases workload for maintenance

TuomasBorman: I applied these changes

TuomasBorman (approved):

PR #597, opened 2024-07-04
QMP example data #470

ShadmanIshraq:


R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
 #' \url{https://static-content.springer.com/esm/art%3A10.1186%2Fs40168-020-00816-x/MediaObjects/40168_2020_816_MOESM3_ESM.docx}
 #' 
 NULL
+
+#' CRC_QMP
```
antagomir: I suggest to use the name "Tito2024QMP"

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' extensive patient phenotyping from a group of 589 colorectal cancer (CRC) 
+#' patients, advanced adenoma (AA) patients, and healthy controls. 
+#' The researchers also reanalyzed data from 15 published studies encompassing 
+#' a total of 4,439 patients and controls. By implementing confounder control and 
+#' quantitative profiling methods, the study was able to reveal potential misleading 
+#' associations between microbial markers and colorectal cancer development that 
+#' were driven by other factors like intestinal inflammation, rather than the cancer 
+#' diagnosis itself.
+#'
+#' @format A Large Tree Summarized Experiment dataset with 676 features and 589 samples.
```
antagomir: TreeSummarizedExperiment is written together (it is the class name)

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
 #' 
 NULL
+
+#' CRC_QMP
+#'
+#' The study combined Quantitative Microbiome Profiling (QMP) with 
+#' extensive patient phenotyping from a group of 589 colorectal cancer (CRC) 
+#' patients, advanced adenoma (AA) patients, and healthy controls. 
+#' The researchers also reanalyzed data from 15 published studies encompassing 
+#' a total of 4,439 patients and controls. By implementing confounder control and
```
antagomir: I think this is unnecessary and potentially confusing to mention because the data set does not include these 4439 patients -> remove this sentence?

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' quantitative profiling methods, the study was able to reveal potential misleading 
+#' associations between microbial markers and colorectal cancer development that 
+#' were driven by other factors like intestinal inflammation, rather than the cancer 
+#' diagnosis itself.
+#'
+#' @format A Large Tree Summarized Experiment dataset with 676 features and 589 samples. 
+#' The rowData contains Species name. The colData includes:
+#' 
+#' \describe{
+#'   \item{sample_ID}{Sample ID from the corresponding study}
```
antagomir: Use "sample" or "sampleID"

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#'   \item{Diagnosis}{Diagnosis type, with possible values: ADE (advanced adenoma),
```
antagomir: use "diagnosis" lowercase (or is it "Diagnosis" in the originally downloaded data? - in such case perhaps keep as is )

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' were driven by other factors like intestinal inflammation, rather than the cancer 
+#' diagnosis itself.
+#'
+#' @format A Large Tree Summarized Experiment dataset with 676 features and 589 samples. 
+#' The rowData contains Species name. The colData includes:
+#' 
+#' \describe{
+#'   \item{sample_ID}{Sample ID from the corresponding study}
+#'   \item{Diagnosis}{Diagnosis type, with possible values: ADE (advanced adenoma), 
+#'   CRC (colorectal cancer), CTL (control)}
```
antagomir: I think the values should be in quotes? Like "ADE" etc

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#'   \item{Colonoscopy}{Colonoscopy result, with possible values: FIT_Positive,
```
antagomir: use lowercase for field names "colonoscopy"

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#'
+#' @format A Large Tree Summarized Experiment dataset with 676 features and 589 samples. 
+#' The rowData contains Species name. The colData includes:
+#' 
+#' \describe{
+#'   \item{sample_ID}{Sample ID from the corresponding study}
+#'   \item{Diagnosis}{Diagnosis type, with possible values: ADE (advanced adenoma), 
+#'   CRC (colorectal cancer), CTL (control)}
+#'   \item{Colonoscopy}{Colonoscopy result, with possible values: FIT_Positive, 
+#'   familial_risk_familial_CRC_FCC, familial_risk_no, abdomil_complaints}
```
antagomir: again, values should be in quotes here in the text

antagomir: Are these values with these same names in the original data?

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' }
+#'
+#' @name CRC_QMP
```
antagomir: See the suggeste data name above.

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' @docType data
+#' @keywords datasets
+#' @usage data(CRC_QMP)
```
antagomir: see above

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' @seealso \code{\link{mia-datasets}}
+#' @author 
+#' Tito, R. Y., et al.
+#' Raúl Y. Tito, Sara Verbandt, Marta Aguirre Vazquez, Leo Lahti, Chloe Verspecht, Verónica Lloréns-Rico, Sara Vieira-Silva,
+#' Janine Arts, Gwen Falony, Evelien Dekker, Joke Reumers, Sabine Tejpar & Jeroen Raes
```
antagomir: Here in author field you should write your own name (the author of this function)

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' @references
+#' Raúl Y. Tito, Sara Verbandt, Marta Aguirre Vazquez, Leo Lahti, Chloe Verspecht, Verónica Lloréns-Rico, Sara Vieira-Silva,
+#' Janine Arts, Gwen Falony, Evelien Dekker, Joke Reumers, Sabine Tejpar & Jeroen Raes (2024). 
+#' Microbiome confounders and quantitative profiling challenge predicted microbial targets in colorectal cancer development. 
+#' Nature Medicine,30, 1339-1348. 
+#' \url{https://www.nature.com/articles/s41591-024-02963-2#Sec8}
```
antagomir: Put here DOI url instead

R/mia.R
```diff
@@ -322,3 +322,54 @@ NULL
...
+#' 
+#' The supplementary tables contain colonoscopy referral reasons for the LCPM cohort,
+#' LCMP cohort variables and enterotypes, continuous variables' associations with cancer 
+#' progression, categorical variables' associations with cancer progression, microbiome 
+#' variation analysis in the LCMP cohort, included and excluded species list, species 
+#' abundance differences across diagnostic groups, species abundance associations, 
+#' species abundance differences in normal calprotectin levels subset, categorical 
+#' variables' associations with enterotype distribution, species detected in negative 
+#' controls, species abundance differences between CRC and controls in published cohorts, 
+#' and absolute taxonomic abundances at species level in the LCMP cohort.
```
antagomir: Are these included in the R data set? (associations not easy to include but if there is any additional sample specific info, such as DMM clusters, calprotectin levels, BMI etc etc.

antagomir (commented):
Some comments.

antagomir:
Can you also save code that was used to create the R data from original source files - that can be put in inst/scripts/

R/mia.R
antagomir: For all arguments, mention their type (factor, character, numeric..?). Factor is good for diagnosis and colonoscopy (as is already); character is good sample ID (which is usually unique to each row)

antagomir:
The colData(tse) has sampleID on the rows; it would be good to have this also as its own column, see e.g.

```
data(GlobalPatterns)
names(colData(GlobalPatterns))
```

```
[1] "X.SampleID"               "Primer"                  
[3] "Final_Barcode"            "Barcode_truncated_plus_T"
[5] "Barcode_full_length"      "SampleType"              
[7] "Description"             

```

antagomir:
I suggest to replace periods "." in the rowData species names with "_". The periods occasionally cause problems in some systems.

R/mia.R
ShadmanIshraq: No, this is the whole supplementary data information from the CRC paper, there are 14 supplementary tables. I needed to use only two from the above

ShadmanIshraq:
> Can you also save code that was used to create the R data from original source files - that can be put in inst/scripts/

antagomir:
Just curious, why this was closed? Another option would be to fix the suggested points and add new commits on the existing PR?

ShadmanIshraq:
Shall I put the coldata, rowdata and assaydata also with the R code in inst/scripts/  ?

R/mia.R
ShadmanIshraq: Shall I remove the table information which are unused?

R/mia.R
ShadmanIshraq: Yes, these value names are the same as original data

R/mia.R
ShadmanIshraq: Yes, sample_ID and Diagnosis was like this in the original data.

antagomir:
> Shall I put the coldata, rowdata and assaydata also with the R code in inst/scripts/ ?

That's usually not done but it would be helpful if had these source files available.

You could add here (as a PR):
https://github.com/microbiome/data

Remember to cite the original publication very clearly in the relevant README

R/mia.R
```diff
@@ -323,53 +323,39 @@ NULL
...
-#' were driven by other factors like intestinal inflammation, rather than the cancer 
-#' diagnosis itself.
+#' By implementing confounder control and quantitative profiling methods, the study 
+#' was able to reveal potential misleading associations between microbial markers 
+#' and colorectal cancer development that were driven by other factors like intestinal 
+#' inflammation, rather than the cancer diagnosis itself.
 #'
-#' @format A Large Tree Summarized Experiment dataset with 676 features and 589 samples. 
-#' The rowData contains Species name. The colData includes:
+#' @format A Large TreeSummarizedExperiment with 676 features and 589 samples.
```
antagomir: I don't this is particularly "Large" -> remove "Large"

antagomir (approved):
Great - did you add data files in the microbiome/data repository?

TuomasBorman (approved):

antagomir (approved):
good to go i think

antagomir:
@ShadmanIshraq could you resolve the comments that you have fixed already?

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

R/mia.R
ShadmanIshraq: I have fixed already

antagomir:
@TuomasBorman this is good to go for me, not sure why the GHA error?

TuomasBorman:
Every manual page should be listed in pkgdown/pkgdown.yml to be added to the manual. The new page about this new dataset was not added there which caused the error.

TuomasBorman:
Now the error is caused by permission issues "remote: Permission to microbiome/mia.git denied to github-actions[bot]". This should be fixed when pushed to devel branch.

antagomir:
Ok - could @ShadmanIshraq do that?

TuomasBorman:
Pushing to devel without checks passing requires bypassing branch rules (which can only be done by admins)

Issue #596, opened 2024-07-02
Compute and plot feature loadings in ordination methods


theotimeee:
With @ElySeraidarian, we have been working on computing and plotting feature loadings for several ordination methods. Currently, we can compute and plot the feature loadings for 3 ordination methods: PCA, LDA and NMF. However, we are uncertain about the optimal way to implement this feature in mia (and possibly miaViz for the visualization part).

First, we are uncertain about the optimal way to store feature loadings after calculation. The calculation of feature loadings is different for all methods but they all result in a matrix (features in lines and latent vectors in columns).
Ideally, we would like to store the feature loadings at the same place for all ordination methods (e.g. `attr` or `metadata`) but it might be difficult. The ordination functions we use to compute feature loadings are not storing them at the same place at all.
For instance, when running` runPCA`, the feature loadings matrix is directly stored in attribute "rotation". 

Also, we are wondering if it would be best to create a generic wrapper for plotting. @antagomir suggested that we build specific functions for each method and then a generic wrapper that would recognize the type of arguments and call the relevant method.

I think it is important to mention that to visualize the `rowTree` as well as the feature loadings, we need to restrict the number of features so that the ordination plot remains readable. For that, we use `agglomerateByPrevalence` (which also agglomerate the rowTree thanks to https://github.com/microbiome/mia/pull/575)

theotimeee:
@TuomasBorman Do you have any suggestions or preferences regarding this?

TuomasBorman:
1.
Can you list all the methods that you are implementing? I believe NMF and LDA are not currently available from TreeSE/SingleCellExperiment ecosystem?

PCA --> scater::runPCA()
RDA --> mia::runRDA()/getRDA()

I think first task is to implement LDA and NMF (if they are not yet available). There should be both get* and add* methods available.

get* should output ordination including feature loadings in attribute, that might be the best solution. add* should add the ordination result to reducedDim slot.

2.

Wrapper for feature loadings was already discussed briefly here https://github.com/microbiome/miaViz/issues/92. Certainly there is a need for wrapper. That could take either TreeSE or feature loadings matrix as input.

[...]

PR #598, opened 2024-07-05
LDA ordination with feature loadings computation

theotimeee:
This PR adds two new methods `getLDA` and `addLDA` that perform Latent Dirichlet Allocation.

`getLDA` returns the ordination matrix with the feature loadings matrix stored as attribute.
`addLDA` adds this ordination matrix to `reducedDims` of the TreeSummarizedExperiment.

This PR is related to issue https://github.com/microbiome/mia/issues/596.


antagomir:
Great! I will be first looking fwd to comments from @TuomasBorman and then let's see if I have anything to add.

antagomir:
Conflicts to resolve?

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+#'
+#' These functions perform Latent Dirichlet Allocation on data stored in a 
+#'  \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#' 
+#' @param x a \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#'   
+#' @param k Integer, number of latent vectors/topics 
+#'   (By default: \code{k = 2}).
```
TuomasBorman: (Default: \code{2})

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+#' 
+#' @param name The name to be used to store the result in the reducedDims of the
+#'  output (By default: \code{name = "LDA"}).
+#' 
+#' @param ... optional arguments.
+#' 
+#' @return 
+#' For \code{getLDA}, the ordination matrix with feature loadings matrix
+#'  as attribute \code{'feature_loadings'}.
```
TuomasBorman: Consider just "loadings"

R/runLDA.R
TuomasBorman, line 1: The name of the file should reflect the functions (there is no runLDA function. Other files might be named like run* because names are relics)

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+        scores <- t(as.data.frame(posteriors$topics))
+        feature_loadings <- t(as.data.frame(posteriors$terms)) 
+        attr(scores, "feature_loadings") <- feature_loadings
+        return(scores)
+    }
+)
+
+#' @export
+#' @rdname addLDA
+setMethod("addLDA", "SummarizedExperiment",
```
TuomasBorman: What if user calls this function with SE? --> This does not work. You should update the function to first TreeSE if there is no reducedDim slot (SE does not have one)

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+    function(x, k=2, name = "LDA", ...){
+        reducedDim(x, name) <- t(getLDA(x, k, ...))
```
TuomasBorman: Similarly to internal function ".add_values_to_coldata", we could have ".add_values_to_reduceddim". That could

1. Check if the object is at least SCE. If not, update the object and give warning.
2. Add the result.

That way we could harmonize these functions

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+#' @rdname addLDA
+#' @export
+setGeneric("addLDA", signature = c("x"),
+           function(x, ...)
+             standardGeneric("addLDA"))
+
+#' @export
+#' @rdname addLDA
+setMethod("getLDA", "SummarizedExperiment",
+    function(x, k=2, ...){
```
TuomasBorman: Add assay.type parameter and input checks

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+        lda_model <- topicmodels::LDA(t(assay(x, "counts")), k)
```
TuomasBorman: This is another dependency. Because this function is not used that often, we could

1. Add this dependency to suggests
2. Add .require_package("topicmodels") in the beginning of this function.

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+        df <- as.data.frame(t(assay(x, "counts")))
```
TuomasBorman: You could do this line first and then just input this variable to line 58 (simplify)

R/runLDA.R
```diff
@@ -0,0 +1,75 @@
...
+        feature_loadings <- t(as.data.frame(posteriors$terms)) 
+        attr(scores, "feature_loadings") <- feature_loadings
+        return(scores)
+    }
+)
+
+#' @export
+#' @rdname addLDA
+setMethod("addLDA", "SummarizedExperiment",
+    function(x, k=2, name = "LDA", ...){
```
TuomasBorman: k=2 and name = "LDA"

--> k = 2 --> use similar coding style

TuomasBorman (changes requested):
Looks good. Even though the code is rather short, structure it by adding tests and comments

TuomasBorman:
Couple things to discuss:

1. LDA has features from both ordination and clustering methods, however, it is more clustering method. That should be taken into account in documentation.
2. As this is clustering method, should we add clusters to rowData/colData? In theory, LDA can be applied also to rows, which makes reducedDim not suitable for storing the result. (Instead of ordination methods, addCluster() should be used as an example)
3. And finally, should this be implemented in bluster package (the Bioconductor clustering package)? We could make a PR as we did last year https://github.com/LTLA/bluster/blob/master/R/DmmParam.R

antagomir:
Intuitively, reducedDim is a logical place given that this would be usually done for samples. But it could be in principle applied on both rows and columns. The function generates both feature matrix and feature loading matrix, both should be stored. 

LDA is a matrix decomposition method, and I would not recommend to link it with clustering so directly.

But it could in principle go to the bluster package. Not sure if they would approve but it could be asked by opening an issue. That might take more time, however.. therefore I would meanwhile proceed with implementing these in mia, and later we can move them elsewhere if necessary? LDA and NMF have clear uses, it would be useful to have them easily accessible.

TuomasBorman:
With these explanations, I am fine with mia implementation and reducedDim()

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#'
+#' These functions perform Latent Dirichlet Allocation on data stored in a 
+#'  \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#' 
+#' @param x a \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#'   
+#' @param k Integer, number of latent vectors/topics 
+#'   (By default: \code{2}).
```
TuomasBorman: We are trying to harmonize these descriptions little bit. This could be:

Integer scalar. A number of latent vectors/topics. (Default: \code{2})

The type. Detailed description. Default value.

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#' 
+#' @param name The name to be used to store the result in the reducedDims of the
+#'  output (By default: \code{name = "LDA"}).
```
TuomasBorman: This could be:

Character scalar. The name to be used to store the result in the reducedDims of the output. (Default: \code{"LDA"})

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#'  
+#' @param assay.type a single \code{character} value for specifying which
+#'   assay to use for LDA ordination.
+#'
```
TuomasBorman: Character scalar. Specifies which assay to use for LDA ordination. (Default: \code{"counts"})

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#' @rdname addLDA
+#' @export
+setGeneric("addLDA", signature = c("x"),
+           function(x, ...)
+             standardGeneric("addLDA"))
+
+#' @export
+#' @rdname addLDA
+setMethod("getLDA", "SummarizedExperiment",
+    function(x, k=2, assay.type = "counts", ...){
```
TuomasBorman: k = 2 instead of k=2 --> use similar coding-style (you have assay.type = "counts" already)

R/utils.R
```diff
@@ -349,6 +349,38 @@
...
+    # If the object is SE, convert it to TreeSE
+    if( !is(x, "SingleCellExperiment") ){
+        x <- as(x, "TreeSummarizedExperiment")
+        warning(
+            "SummarizedExperiment does not have reducedDims slot. ",
+            "Therefore, it is converted to TreeSummarizedExperiment.",
+            call. = FALSE)
+    }
+    # Throw warning if values of reducedDim are overwritten
+    if ( c(name) %in% names(reducedDims(x)) ){
```
TuomasBorman: c(name) --> name

R/utils.R
```diff
@@ -349,6 +349,38 @@
...
+        warning(
+          "The following values are already present in `reducedDims` and will ",
+          "be overwritten: '",
+          paste(name),
```
TuomasBorman: No need to use paste(), it should work without it (moreover, Bioc guidelines gives say that they should not be used inside warning/stop)

TuomasBorman (changes requested):
Minor things

tests/testthat/test-5addLDA.R
```diff
@@ -0,0 +1,40 @@
...
+  )
+  expect_error(
+    addLDA(GlobalPatterns, k = 2, assay.type = TRUE, name = "LDA")
+  )
+  expect_error(
+    addLDA(GlobalPatterns, k = 2, assay.type = "counts", name = 1)
+  )
+  expect_error(
+    addLDA(GlobalPatterns, k = 2, assay.type = "counts", name = TRUE)
+  )
```
TuomasBorman, line 51: Can you still add test that the calculated result is same with getLDA, addLDA and topicmodels::LDA?

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#'  with feature loadings matrix as attribute \code{"loadings"}.
+#'  
+#' @name addLDA
+#' 
+#' @examples
+#' data(GlobalPatterns)
+#' 
+#' # Reduce the number of features 
+#' tse <- agglomerateByPrevalence(GlobalPatterns, rank="Phylum", prevalence=0.99, update.tree = TRUE)
+#'
```
TuomasBorman: tse <- GlobalPatterns

And then run aglomeratebyPrevalence

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+        if( !.is_an_integer(k) ){
+            stop("'k' must be an integer.", call. = FALSE)
+        }
+        if( !.is_a_string(assay.type) ){
+            stop("'assay.type' must be a non-empty single character value.",
+              call. = FALSE)
+        }
+        df <- as.data.frame(t(assay(x, assay.type)))
+        # Estimate LDA model using VEM algorithm
+        lda_model <- topicmodels::LDA(df, k)
```
TuomasBorman: Pass ... so that user can choose the method and control parameters of topicmodels::LDA

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+#' @param k Integer scalar. A number of latent vectors/topics. 
+#'  (Default: \code{2})
+#' 
+#' @param name Character scalar. The name to be used to store the result in the
+#'  reducedDims of the output. (Default: \code{"LDA"})
+#'  
+#' @param assay.type Character scalar. Specifies which assay to use for LDA 
+#'  ordination. (Default: \code{"counts"})
+#' 
+#' @param ... optional arguments.
```
TuomasBorman: In details section say clearly that topicsmodel::LDA is used

Here say that additional parameters are passed to that function

R/addLDA.R
```diff
@@ -0,0 +1,105 @@
...
+        }
+        df <- as.data.frame(t(assay(x, assay.type)))
+        # Estimate LDA model using VEM algorithm
+        lda_model <- topicmodels::LDA(df, k)
+        # Calculate scores and loadings
+        posteriors <- topicmodels::posterior(lda_model, df)
+        scores <- t(as.data.frame(posteriors$topics))
+        loadings <- t(as.data.frame(posteriors$terms)) 
+        # Add loadings as attribute of the scores matrix
+        attr(scores, "loadings") <- loadings
```
TuomasBorman, line 116: You could also add the model to attributes

TuomasBorman:
Usually, the number of topics "k" is determined by iterating over several numbers. The best fit is evaluated with coherence and perplexity metrics.

Should k be a vector of values so that the function iterates over them calculating LDA for all different number of topics? The function could evaluate the goodness of fit and then return the best model. The function could also return a table which tells the metrics for each k value (that could be then easily plotted). 

@antagomir

antagomir:
Ok to me to add that. 

However, if it feels a lot to add now, I would focus on finalizing the present one (with a single fixed k) first, and then adding that multi-k optimization as a separate PR.

R/addLDA.R
TuomasBorman: btw does

agglomerateByPrevalence(tse, rank="Phylum")

That would be more simple

TuomasBorman:
That would require something like this


```
addLDA <- function(eval.metric = "perplexity"){
    if(eval.metric == "coherence" && !require("topicdoc") ){
        stop("'topicmodel' package must be installed in order to calculate coherence.")
    }
    models <- lapply(k, function(x){
        LDA(k = x)
    }
    
    metric <- .calc_metrics(models)
    
    # Get the model with lowest eval.metric (dfault value perplexity).
    
    # Get scores
    # Add loadings, model, and metrics data.frame to attributes

}

.calc_metrics <- function(x){
    # Loop through models
    # Calculate perplexity with topicmodels
    # Calculate coherence (I believe you have to use topicdoc package). Calculate it if that package is available
    # Create df from both metrics
}
```

tests/testthat/test-5addLDA.R
theotimeee: I added tests but there is an average diff of 7*10**(-5) that comes from adding the result to reducedDims I think, so I put a tolerance level of 10**(-4)

R/utils.R
```diff
@@ -349,6 +349,38 @@
...
 }
+
+# This function can be used to add values to reducedDims
+.add_values_to_reducedDims <- function(x, values, name, ...){
+    # Check values
+    if( !((is(values, "matrix") && length(values) > 0)) ){
+        stop("'values' must be a matrix.", call. = FALSE)
+    }
+    # Check names
+    if( !is.character(name) && length(name) > 1L ){
```
TuomasBorman: Use .is_a_string()

R/addLDA.R
```diff
@@ -0,0 +1,113 @@
+#' Latent Dirichlet Allocation
+#'
+#' These functions perform Latent Dirichlet Allocation on data stored in a 
+#'  \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#' 
+#' @param x a \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'  object.
+#'   
+#' @param k Integer scalar. A number of latent vectors/topics.
```
TuomasBorman: Looks good. Can you still put the class inside  \code{] to make it nicer. --> \code{Integer scalar} (Check other also)

R/addLDA.R
```diff
@@ -0,0 +1,113 @@
...
+    function(x, k = 2, assay.type = "counts", ...){
+        .require_package("topicmodels")
+        # Input checks
+        if( !.is_an_integer(k) ){
+            stop("'k' must be an integer.", call. = FALSE)
+        }
+        if( !.is_a_string(assay.type) ){
+            stop("'assay.type' must be a non-empty single character value.",
+              call. = FALSE)
+        }
```
TuomasBorman: Intead, use .check_assay_present which also checks that the assay is available

TuomasBorman (changes requested):

TuomasBorman:
NMF would go with this same template I think. However, there k (rank) can be numeric vector and the function finds optimal k by itself

TuomasBorman:
For LDA we could have another PR for multiple k values, if you are not able to implement it now

The idea was this:


```
data("GlobalPatterns")
tse <- GlobalPatterns
tse <- agglomerateByPrevalence(tse, rank="Phylum")

eval.metric <- "perplexity"
k <- c(2, 3, 4, 5, 6, 7, 8, 9, 10)
k <- 5
assay.type <- "counts"

mat <- assay(tse, assay.type)

df <- as.data.frame(t(mat))




if( eval.metric == "coherence" ){
    .require_package("topicdoc")
}

models <- lapply(k, function(i){
    topicmodels::LDA(df, i)
})

metrics <- .calculate_lda_metrics(models, k)

lda_model <- models[[ which.max(metrics[[eval.metric]]) ]]

posteriors <- topicmodels::posterior(lda_model, df)
scores <- t(as.data.frame(posteriors$topics))
attr(scores, "loadings") <- loadings
attr(scores, "model") <- lda_model
attr(scores, "eval_metric") <- metrics 

.calculate_lda_metrics <- function(models, k){
    metrics <- lapply(models, function(model){
        res <- topicmodels::perplexity(model)
        names(res) <- "perplexity"
        if( require("topicdoc") ){
            coherence <- topicdoc::topic_coherence(model, df)
            names(coherence) <- paste0("coherence_", seq_len(length(coherence)))
            coherence <- c(coherence, coherence_mean = mean(coherence))
            res <- c(res, coherence)
        }
        return(res)
    })
    metrics <- as.data.frame( dplyr::bind_rows(metrics) )
    metrics[["k"]] <- k
    return(metrics)
}

#################################
df <- attr(scores, "eval_metric")

library(ggplot2)

ggplot(df) +
    geom_point(aes(x = k, y = perplexity), color = "red")
ggplot(df) +
    geom_point(aes(x = k, y = mean_coherence), color = "blue")


```

R/utils.R
```diff
@@ -349,6 +349,38 @@
...
+    }
+    # Throw warning if values of reducedDim are overwritten
+    if ( name %in% names(reducedDims(x)) ){
+        warning(
+          "The following values are already present in `reducedDims` and will ",
+          "be overwritten: '",
+          name,
+          "'. Consider using the 'name' argument to specify alternative ",
+          "names.", call. = FALSE)
+    }
```
TuomasBorman: Indentation is off

TuomasBorman (changes requested):
Is this PR ready? If the modification on multiple k values do not come to this PR, I think this is good to go

R/utils.R
theotimeee: I fixed the indentation and modified the input check to allow dist objects as well as matrices.

TuomasBorman (approved):

Issue #561, opened 2024-05-28
Use vegan in rarefy


TuomasBorman:
Should we switch using [vegan::rrafefy](https://rdrr.io/cran/vegan/man/rarefy.html) (it is fast since it is programmed with C)? One downside is that it does not support `replace`. Moreover, one thing to discuss (as I am not the expert of the terminology), could it make more sense to do iterations? As I have understood, the criticism towards rafefying is mostly related to bias and missing data caused by rarefaction single time. However, if enough iterations is done, that would solve the issue.

@antagomir

antagomir:
1. vegan::rrarefy : missing support for replace is not good (rarefaction is often done with replacement) but perhaps not so critical;  our case does seem prohibitively slow and subsampling is a simple operation so I am not sure if this is needed. Fast is good but it would help to know how much faster, if there is real advantage (in this or other regards).
2. iterations: this was validated by Pat Schloss in the context of alpha and beta diversity calculation over several iterations and those PRs are already open. I am not familiar with supporting iterations to just get rarefied data. The iterations would not solve the issue that the samples with more reads will have a higher resolution for that information even after iterations. I would refrain from implementing this until there is reference or experiments demonstrating benefits in a general case.

antagomir:
Action point: compare the standard rarefaction and rrarefy. If there are significant gains in speed with essentially similar outcomes then consider switching.

TuomasBorman:
Maybe @Daenarys8 when time allows?

0xMuluh:
Here is some update:
```
>library(mia)
>library(vegan)
>library(microbenchmark)
> data("GlobalPatterns")
> tse <- GlobalPatterns
> benchmark <- microbenchmark(
     vegan_run = rrarefy(assay(tse, "counts"), 1000),
     mia_run = rarefyAssay(tse, 1000, assay.type = "counts"),
     times = 100
 )
> benchmark
Unit: milliseconds
      expr      min       lq     mean   median       uq      max neval cld
 vegan_run 205.8761 232.9813 256.7301 247.6536 262.2308 1071.452   100  a 
   mia_run 699.3646 754.3515 844.8951 806.7249 841.4128 1639.997   100   b

> vegan_run_1 <- rrarefy(assay(tse, "counts"), 1000)
> dim(vegan_run_1)
[1] 19216    26
> mia_run_1 <- rarefyAssay(tse, 1000, assay.type = "counts")
> dim(assay(mia_run_1, "subsampled"))
[...]

PR #609, opened 2024-07-16
New dissimilarity functions

theotimeee:
This PR adds new methods `getDissimilarity` and `addDissimilarity`.
These methods combine overlap, JSD, and UniFrac dissimilarity calculations as implemented in mia, and additionally use vegan dissimilarity function when overlap, JSD, or UniFrac are not employed.

This PR also makes the following naming modifications:
`calculateOverlap` --> `getOverlap`
`runOverlap` --> `addOverlap`
`calculateUnifrac` --> `getUnifrac`
`calculateJSD` --> `getJSD`


NAMESPACE
```diff
@@ -1,16 +1,19 @@
 # Generated by roxygen2: do not edit by hand
 
+export(.getJSD)
```
TuomasBorman: Internal function exported

NAMESPACE
```diff
@@ -160,9 +163,6 @@ exportMethods(agglomerateByVariable)
 exportMethods(bestDMNFit)
 exportMethods(calculateDMN)
 exportMethods(calculateDMNgroup)
-exportMethods(calculateJSD)
-exportMethods(calculateOverlap)
-exportMethods(calculateUnifrac)
```
TuomasBorman: It seems that these are removed and not deprecated

R/calculateJSD.R
```diff
@@ -116,15 +116,15 @@ setMethod("calculateJSD", signature = c(x = "SummarizedExperiment"),
...
+#' @rdname getJSD
 #'
 #' @importFrom utils combn
 #' @importFrom stats as.dist
 #' @importFrom BiocParallel SerialParam register bplapply bpisup bpstart bpstop
 #' @importFrom DelayedArray getAutoBPPARAM setAutoBPPARAM
 #'
 #' @export
-runJSD <- function(x, BPPARAM = SerialParam(), chunkSize = nrow(x)){
+.getJSD <- function(x, BPPARAM = SerialParam(), chunkSize = nrow(x), ...){
```
TuomasBorman: Use naming scheme .get_jsd

R/calculateJSD.R
```diff
@@ -56,42 +56,42 @@
...
     function(x, assay.type = assay_name, assay_name = exprs_values, 
              exprs_values = "counts", transposed = FALSE, ...){
         mat <- assay(x, assay.type)
         if(!transposed){
             mat <- t(mat)
         }
-        calculateJSD(mat, ...)
+        getJSD(mat, ...)
     }
 )
```
TuomasBorman: This generic functions cause quite a lot overhead. Can we just remove them since they are called from getDissimilarity (the data should be ready for .getJSD)?

R/calculateOverlap.R
```diff
@@ -82,55 +82,72 @@ setMethod("calculateOverlap", signature = c(x = "SummarizedExperiment"),
...
+setMethod("getOverlap", signature = c(x = "ANY"),
+    function(x,detection = 0, ...){
+        ############################# INPUT CHECK ##############################
+        # Check detection
+        if (!.is_numeric_string(detection)) {
+          stop("'detection' must be a single numeric value or coercible to ",
+               "one.",
+               call. = FALSE)
+        }
+        detection <- as.numeric(detection)
```
TuomasBorman: Same here. Can we convert this to internal function and call it from getDissimilarity?

TuomasBorman: The problem is similar to this in addAlpha https://github.com/microbiome/mia/pull/612/files

When we are just adding new components, it makes the code harder to maintain.

We could just have getDissimilarity that calls

- .get_unifrac
- .get_overlap
- .get_jsd

Those internal functions should have as little as possible functionality (they should have only method-specific functionality.


In fact, I think getDissimilarity already have most of the functionality needed. It is only missing tree functionality. (Is the tree that we use from rowTree()/colTree() slot or is it directly provided by user). You could implement it here .calculate_dissimilarity (as it should be hidden because it is not used by every method).

Check addAlpha for idea. Let's discuss more if you did not get the idea

TuomasBorman (changes requested):

TuomasBorman:
@antagomir 

The idea was to add dissimilarity calculation method back (we had it couple years ago).

1. Dissimilarity calculation is common step and we could have wrapper for "all" the dissimilarity methods
    - We removed the previous wrapper because it just applied vegan::vegdist to assay
2. We have our "own" dissimilarity measures implemented in mia (JSD, overlap, unifrac) in any case
3. This makes applying dissimilarity calculation little bit easier and straightforward.

The idea is this:
```
data("GlobalPatterns")
tse <- GlobalPatterns

# Add to reducedDim
tse <- addDissimilarity(tse, method = "euclidean")
tse <- addDissimilarity(tse, method = "unifrac")
# Get only the dissimilarity matrix
res <- getDissimilarity(tse, method = "bray", diss.fun = vegan::vegdist)
# It works also for matrices
res <- getDissimilarity(assay(tse), method = "bray")
# We can use runMDS with getDissimilarity (makes things little bit easier when we can always use getDissimilarity)
tse <- runMDS(tse, method = "unifrac", tree = rowTree(tse), FUN = getDissimilarity)
tse <- runMDS(tse, method = "bray", FUN = getDissimilarity)
```

antagomir:
Yes, seems good.

In principle it would be great to be able to run methods like runMDS() for TreeSE objects without specifying the tree separately but that's a separate discussion.

The reducedDim slot works for this but it is also somewhat confusing because it is also used for ordination results are derived from dissimilarities.

TuomasBorman:
1. We could create addMDS wrapper for that. The code below just shows the idea.

```

data("GlobalPatterns")
tse <- GlobalPatterns

tse <- runMDS(
    tse,
    FUN = mia::calculateUnifrac,
    name = "Unifrac",
    tree = rowTree(tse),
    assay.type = "counts")

addMDS <- function(tse, tree = rowTree(tse), ...){
    runMDS(tse, tree = tree, ...)
}

tse <- addMDS(
    tse,
    FUN = mia::calculateUnifrac,
    name = "Unifrac",
    assay.type = "counts")
```
2. We could add it also to metadata(), but we should check what is the consensus. (however, as that is easy to change, we can check that thing after the rest of the function is ready). Adding to metadata() would solve the issue about dissimilarities of features as they cannot be stored in reducedDim().

R/calculateJSD.R
```diff
@@ -1,97 +1,3 @@
...
-#'
-#' jsd <- calculateJSD(enterotype)
-#' class(jsd)
-#' head(jsd)
-#'
-#' enterotype <- runMDS(enterotype, FUN = calculateJSD, name = "JSD",
-#'                      exprs_values = "counts")
-#' head(reducedDim(enterotype))
-#' head(attr(reducedDim(enterotype),"eig"))
-#' attr(reducedDim(enterotype),"GOF")
```
TuomasBorman, line 64: The documentation should be moved to getDissimilarity (good and important info)

R/calculateOverlap.R
```diff
@@ -1,133 +1,36 @@
...
-#' 
-#' @name calculateOverlap
-#' @export
-#'
-#' @author Leo Lahti and Tuomas Borman. Contact: \url{microbiome.github.io}
-#' 
-#' @examples
-#' data(esophagus)
-#' tse <- esophagus
-#' tse <- transformAssay(tse, method = "relabundance")
```
TuomasBorman, line 41: Same here. Move the documentation. (Also from unifrac)

R/calculateDistance.R
```diff
@@ -1,6 +1,201 @@
...
+    }
+    if( !identical(rownames(as.matrix(res)), colnames(assay(x, assay.type))) ){
+      warning("Samples of the dissimilarity matrix should be the same as the
+            samples in columns of the assay specified with 'assay.type'. The 
+            result is not added to reducedDim.")
+      return(res)
+    }
+    else{
+      .add_values_to_reducedDims(x, as.matrix(res), name)
+    }
```
TuomasBorman: This checking is related to all situations where values are arred to reduced dim. Can this be handled in the .add_values_to_reducedDims function?

TuomasBorman (changes requested):
I tried to simplify the structure as much as possible. Check if it is working as expected

This is the idea:
```
data("GlobalPatterns")
tse <- GlobalPatterns


getDissimilarity(tse, method = "unifrac")
getDissimilarity(tse, method = "unifrac", tree = rowTree(tse))

se <- as(tse, "SummarizedExperiment")
# BUG!!!!!!!!!!!!!!!!!
rownames(se) <- rownames(tse)
# Should not work
getDissimilarity(se, method = "unifrac")
# SHould work
getDissimilarity(tse, method = "unifrac", tree = rowTree(tse))
```

TuomasBorman:
Check also that getDissimilarity works correctly with runMDS

R/calculateDistance.R
theotimeee: Yes I am going to add it to `.add_values_to_reducedDims`, thank you

antagomir:
This seems to be in good hands but let me know if anything needs to be discussed.

R/calculateDistance.R
```diff
@@ -1,6 +1,315 @@
...
+    #
+    # Calculate dissimilarity
+    mat <- .calculate_dissimilarity(mat = x, method = method, ...)
+    return(mat)
+    }
+)
+
+.calculate_dissimilarity <- function(
+        mat, method, node.label = NULL, diss.fun = NULL, tree = NULL, ...){
+    # input check
```
TuomasBorman: diss.fun (user can define the dissimilarity function to be used), node.label (unifrac) and tree (unifrac)

R/calculateDistance.R
```diff
@@ -1,6 +1,315 @@
...
+#'   
+#' @param exprs_values Deprecated. Use \code{assay.type} instead.
+#'   
+#' @param transposed \code{Logical scalar}. Specifies if x is transposed with cells in
+#'   rows. (Default: \code{FALSE})
+#'
+#' @param tree.name \code{Character scalar}. Specific to unifrac dissimilarity. 
+#' Specifies the name of the tree used in calculation. (Default: \code{"phylo"})
+#' 
+#' @param tree_name Deprecated. Use \code{tree.name} instead.
```
TuomasBorman: As these are specific only for unifrac, these should be "hidden". CAn you put them to .get_tree_args()? Does it work?

R/utils.R
```diff
@@ -393,8 +394,16 @@
             "'. Consider using the 'name' argument to specify alternative ",
             "names.", call. = FALSE)
     }
-    reducedDim(x, name) <- values
-    return(x)
+    if( !identical(rownames(as.matrix(values)), 
+                  colnames(assay(x, assay.type))) ){
+        stop("Samples of the dissimilarity matrix should be the same as the",
+                " samples in columns of the assay specified with 'assay.type'.",
```
TuomasBorman: No need for assay here. The same colnames are in TreeSE: `colnames(x)`

R/utils.R
```diff
@@ -393,8 +394,16 @@
...
+                " The result is not added to reducedDim.")
```
TuomasBorman: This checking should be prior to the if sentence above. User should get only the error (warning is not related to this issue).

```
> addDissimilarity(tse, method = "overlap", detection = 0.25, transposed = TRUE)
Found more than one class "phylo" in cache; using the first, from namespace 'phyloseq'
Also defined by ‘tidytree’
Found more than one class "phylo" in cache; using the first, from namespace 'phyloseq'
Also defined by ‘tidytree’
Error in .add_values_to_reducedDims(x, values = as.matrix(res), name = name,  : 
  Samples of the dissimilarity matrix should be the same as the samples in columns of the assay specified with 'assay.type'. The result is not added to reducedDim.
In addition: Warning message:
The following values are already present in `reducedDims` and will be overwritten: 'overlap'. Consider using the 'name' argument to specify alternative names
```

R/calculateDistance.R
```diff
@@ -1,6 +1,315 @@
...
+  "addDissimilarity", signature = c("x"), function(x, method, ...)
+    standardGeneric("addDissimilarity"))
+
+#' @rdname getDissimilarity
+#' @export
+setMethod(
+  "addDissimilarity", signature = c(x = "SummarizedExperiment"),
+  function(
+    x, method, assay_name = "counts", assay.type = assay_name, name = method,
+    transposed = FALSE, tree_name = "phylo", tree.name = tree_name, ...){
```
TuomasBorman: transposed means that the matrix is transposed. This means ultimately that the matrix do not fit anymore to the original TreeSE (dissimilarity matrix should have samples in rows). --> We should remove and disable transposed from addDissimilarity. Catch it so that it is not fed to getDissimilarity()

TuomasBorman (changes requested):
Looks very good. The functionality is there, I think. Now the final polishing. Also can you add unit tests for getDissimilarity vs vegan::vegdist. And that getDissimilarity works with external tree: getDissimilarity(tse, tree = rowTree(tse), method = "unifrac")

R/calculateDistance.R
TuomasBorman: These are not inclueded in parameter description. SHould be added

TuomasBorman: Check from old unifrac function

R/calculateDistance.R
antagomir: Btw what is the motivation for having "transposed" argument in the first place?

R/utils.R
antagomir: I think that `colnames(assay(x, assay.type))` is equal to `colnames(x)`? Better to use the shorter form?

R/addDissimilarity.R
```diff
@@ -0,0 +1,312 @@
...
+    }
+    #
+    # Calculate dissimilarity
+    mat <- .calculate_dissimilarity(mat = x, method = method, ...)
+    return(mat)
+    }
+)
+
+.calculate_dissimilarity <- function(
+        mat, method, node.label = NULL, diss.fun = NULL, tree = NULL, ...){
```
antagomir: Should we use argument name `dis.fun` instead? The fun fact is that this would match both "distance" and "dissimilarity", and some of our dissimilarity measures are distances.

R/addDissimilarity.R
```diff
@@ -0,0 +1,312 @@
...
+#'   
+#'   \item \code{detection}: (overlap) \code{Numeric scalar}.
+#'   Defines detection threshold for absence/presence of features. Feature that 
+#'   has abundance under threshold in either of samples, will be discarded when 
+#'   evaluating overlap between samples. (Default: \code{0}) 
+#' }
+#'
+#' @return 
+#' \code{getDissimilarity} returns a sample-by-sample distance matrix, suitable 
+#'   for NMDS, etc.
```
antagomir: Why only NMDS mentioned? Either mention no method, or multiple methods (all available ones?)

R/addDissimilarity.R
```diff
@@ -0,0 +1,312 @@
...
+#' 
+#' \code{addDissimilarity} returns \code{x} that includes distance matrix in its
```
antagomir: distance matrix -> dissimilarity matrix?

R/addDissimilarity.R
```diff
@@ -0,0 +1,312 @@
...
+#'   Specifies whether the calculation should be parallelized.
+#'   
+#'   \item \code{detection}: (overlap) \code{Numeric scalar}.
+#'   Defines detection threshold for absence/presence of features. Feature that 
+#'   has abundance under threshold in either of samples, will be discarded when 
+#'   evaluating overlap between samples. (Default: \code{0}) 
+#' }
+#'
+#' @return 
+#' \code{getDissimilarity} returns a sample-by-sample distance matrix, suitable
```
antagomir: distance matrix -> dissimilarity matrix? (also check elsewhere, let's be consistent)

R/calculateDistance.R
TuomasBorman: It is to allow calculation row-wise (dissimilarity between features). Also, one option is to store dissimilarity matrix to metadata() for instance (however, reducedDim is optimal place for sample-wise dissimilarities).

antagomir:
Does the new dissimilarity function support rarified values? If not, let us open a new issue on that one and solve later.

R/calculateDistance.R
antagomir: Yes right. 

It is true that reducedDim is quite helpful solution for that. I am somewhat concerned, however, that it might be confusing as a) it can only handle one type of output from dissimilarity function, and b) it is not an actual reduced dimensional presentation (nor altExp). 

How about using a slot with recommended standard name in metadata()?

TuomasBorman:
> Does the new dissimilarity function support rarified values? If not, let us open a new issue on that one and solve later.

No. I am wondering if we should add rarefaction also for overlap, jsd and unifrac (for vegan dissimilarities there is already vegan::avgdist)

antagomir:
> > Does the new dissimilarity function support rarified values? If not, let us open a new issue on that one and solve later.
> 
> No. I am wondering if we should add rarefaction also for overlap, jsd and unifrac (for vegan dissimilarities there is already vegan::avgdist)

IMO it would be helpful if getDissimilarity could support `vegan::avgdist` (for rarefaction purposes). 

In addition, we could have rarefaction supported also for overlap, jsd and unifrac but I do not consider this urgent. The unifrac seems to me the most common one out of these so I would start with that (after adding support for vegan::avgdist).

R/calculateDistance.R
theotimeee: If I understood correctly, I should remove the `transposed` argument from the `getDissimilarity` and `addDissimilarity` functions but should I also make modifications about the storage of dissimilarity matrix or will it be addressed in another PR?

NAMESPACE
```diff
@@ -117,7 +119,6 @@ export(runJSD)
 export(runNMDS)
 export(runOverlap)
 export(runRDA)
-export(runUnifrac)
```
TuomasBorman: It seems that this is removed?

R/calculateDistance.R
TuomasBorman: I would I would keep `transposed`. Instead of storing the result to reducedDim(), we could store it to `metadata()`.

However, one thing to keep in mind is that metadata() is just a list. So this does not work `metadata(tse, "diss")`, but this works `metadata(tse)[["diss"]]`

R/utils.R
```diff
@@ -367,7 +367,8 @@
 }
 
 # This function can be used to add values to reducedDims
-.add_values_to_reducedDims <- function(x, values, name, ...){
+.add_values_to_reducedDims <- function(x, values, name, 
+                                        assay.type = "counts", ...){
     # Check values
```
TuomasBorman: assay.type is not needed. Can you remove it

R/utils.R
```diff
@@ -385,6 +386,11 @@
             "Therefore, it is converted to TreeSummarizedExperiment.",
             call. = FALSE)
     }
+    if( !identical(rownames(as.matrix(values)), colnames(x)) ){
+        stop("Samples of the dissimilarity matrix should be the same as the",
+             " samples in columns of the assay specified with 'assay.type'.",
+             " The result is not added to reducedDim.")
+    }
```
TuomasBorman: This message is not true, since the matrix is not usually dissimilarity matrix (it is ordination matrix usually) -->

"Rownames of the matrix should match with colnames(x)."

R/addDissimilarity.R
TuomasBorman: I agree

TuomasBorman (changes requested):

TuomasBorman:
> > > Does the new dissimilarity function support rarified values? If not, let us open a new issue on that one and solve later.
> > 
> > 
> > No. I am wondering if we should add rarefaction also for overlap, jsd and unifrac (for vegan dissimilarities there is already vegan::avgdist)
> 
> IMO it would be helpful if getDissimilarity could support `vegan::avgdist` (for rarefaction purposes).
> 
> In addition, we could have rarefaction supported also for overlap, jsd and unifrac but I do not consider this urgent. The unifrac seems to me the most common one out of these so I would start with that (after adding support for vegan::avgdist).

Now rarefaction is supported for all methods through vegan::avgdist

TuomasBorman:
@thpralas can you check that everything works. Can you do some checks for rarefaction?

I will still fix documentation. assay_name was additional to getDissimilarity but removing it messed up other doc pages...

E: the documentation should be fine now

TuomasBorman:
vegan is in imports. However, there are multiple` .require_package("vegan")`. They all can be removed

theotimeee:
I tried adding checks on rarefaction but I get different results using the wrapper and `vegan::avgdist`. I guess it is caused by the random sampling in rarefaction but even when setting seed to obtain the same result, it does not work as intended.

antagomir:
> I tried adding checks on rarefaction but I get different results using the wrapper and `vegan::avgdist`. I guess it is caused by the random sampling in rarefaction but even when setting seed to obtain the same result, it does not work as intended.

Setting the seed probably doesn't help because these functions have different command sequences anyway, then the results from even same random seed do not necessarily match exactly.

More essential is that the results are sufficiently close. Could you paste here example comparison code and resulting comparison plot? There is related active discussion in #561 - we have similar issue with alpha diversity.

If we can conclude that both methods work correctly (there can be some technical choices that are acceptable and explain the differences), and if they yield comparable results then it would be very helpful to just get this merged soon, and open a separate issue that should investigate why results from vegan::avgdist, vegan::rrarefy, and mia wrappers are different, and what should be done about that.

theotimeee:
Here is the check and its ouput:

```
tse <- GlobalPatterns
mat <- assay(tse, "counts")
ntop <- 5
tse_sub <- tse[head(rev(order(rowSds(assay(tse, "counts")))), ntop), ]
mat <- assay(tse_sub, "counts")
clr <- function (x) {
    vegan::decostand(x, method="clr", pseudocount=1)
}
set.seed(123)
res1 <- vegan::avgdist(t(mat), distfun = vegdist, dmethod = "euclidean",
                                     sample = min(colSums2(mat)), iterations = 10, 
                                     transf = clr)
set.seed(123)
res2 <- getDissimilarity(tse_sub, method = "euclidean", niter = 10, 
                                      transf = clr)
expect_equal(as.matrix(res1), as.matrix(res2))
```
>Error: as.matrix(res1) not equal to as.matrix(res2).
650/676 mismatches (average diff: 10.3)
[2]  1.43 -  5.01 ==  -3.58
[3]  2.68 -  7.61 ==  -4.93
[4]  3.84 - 14.92 == -11.09
[5]  3.81 - 14.84 == -11.03
[6]  7.17 - 24.08 == -16.91
[7]  3.82 - 12.16 ==  -8.33
[8]  5.43 - 17.06 == -11.62
[9]  8.03 - 26.99 == -18.97
[10] 6.16 - 19.03 == -12.87

The results of `getDissimilarity` and `vegan::avgdist` are of the same order of magnitude but differ significantly.

TuomasBorman:
There was a problem: dist() function was used incorrectly. Intention was to convert "normal" matrix to distance matrix. dist() was used instead of as.dist(). I removed conversion completely. Without rarefaction, getDissimilarity outputs distance matrix; with rarefaction it outputs normal matrix. (Not sure if this is a big deal)

Does it work now?

theotimeee:
> There was a problem: dist() function was used incorrectly. Intention was to convert "normal" matrix to distance matrix. dist() was used instead of as.dist(). I removed conversion completely. Without rarefaction, getDissimilarity outputs distance matrix; with rarefaction it outputs normal matrix. (Not sure if this is a big deal)
> 
> Does it work now?

Yes it works now, thanks!

TuomasBorman (approved):

TuomasBorman:
Nice work! I think this is now good to go after checks pass

PR #614, opened 2024-07-22
adding css method

SHillman836:


R/transformCounts.R
```diff
@@ -317,6 +325,25 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+  # If mat is a DelayedArray, convert it to a regular matrix for computation
+  if (inherits(mat, "DelayedArray")) {
+    mat <- as.matrix(mat)
+  }
+  
+  # Calculate the cumulative sum for each sample
+  cum_sum <- apply(mat, 2, cumsum)
+
+  # Find the scaling factor for each sample, typically the 75th percentile of the cumulative sum
+  scaling_factors <- apply(cum_sum, 2, function(x) quantile(x, 0.75))
```
antagomir: The 0.75 value should be something that the user can change if they wish. Put this to the function arguments, and 0.75 can be the default when it is not specified. See other functions for similar stuff on how it is best implemented.

tests/testthat/test-5transformCounts.R
```diff
@@ -70,6 +70,20 @@ test_that("transformAssay", {
...
         
+        ############################ CSS ######################################
+        # Apply CSS normalization using transformAssay
+        tmp <- mia::transformAssay(tse, method = "css")
+        ass <- assays(tmp)$css
+        
+        # Calculates CSS normalization. Should be equal.
+        mat <- as.matrix(assay(tse, "counts"))
+        
+        # Manually compute CSS normalization
```
antagomir: Should we also include comparison with an external package implementation? Or at least you could test that manually and include the code here even if it is commented out.

R/transformCounts.R
```diff
@@ -56,6 +56,9 @@
 #' 'relabundance', 'rrank', 'standardize', 'total': please refer to 
 #' \code{\link[vegan:decostand]{decostand}} for details.
 #' 
+#' \item 'css': Cumulative Sum Scaling (CSS) can be used to normalize count data by accounting for differences 
+#' in library sizes.
```
antagomir: Any chance to include a literature reference where this method is explained? 

Also explain that this requires the cutoff parameter (default 0.75)

R/transformCounts.R
TuomasBorman: Perhaps pass it with `...` as it affect only this transformation

R/transformCounts.R
```diff
@@ -228,6 +235,7 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
     FUN <- switch(method,
                   log10 = .calc_log,
                    log2 = .calc_log,
+                  css = .calc_css,
```
TuomasBorman: Can you make the indentation to e multiplication of 4 spaces (also the lines 236-237)

R/transformCounts.R
TuomasBorman: indentation is off

R/transformCounts.R
```diff
@@ -317,6 +325,25 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+.calc_css <- function(mat, ...){
+  # If mat is a DelayedArray, convert it to a regular matrix for computation
+  if (inherits(mat, "DelayedArray")) {
+    mat <- as.matrix(mat)
+  }
+  
+  # Calculate the cumulative sum for each sample
+  cum_sum <- apply(mat, 2, cumsum)
+
+  # Find the scaling factor for each sample, typically the 75th percentile of the cumulative sum
```
TuomasBorman: The maximum characters per line should be <80

R/transformCounts.R
```diff
@@ -317,6 +325,25 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+####################################.apply_css###################################
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, ...){
+  # If mat is a DelayedArray, convert it to a regular matrix for computation
+  if (inherits(mat, "DelayedArray")) {
+    mat <- as.matrix(mat)
+  }
+  
+  # Calculate the cumulative sum for each sample
+  cum_sum <- apply(mat, 2, cumsum)
```
TuomasBorman: How about `colCumsums(mat)` so we can remove the step that converts DelayedMatrix to matrix?

TuomasBorman: Add importFrom tag above .calc_css

SHillman836: sorry as in how do you mean above .calc_css - I thought that kind of thing went in the namespace?

TuomasBorman: Yes, they go, but namespace is generated automatically based on tags in R files. See other functions. IN your case, this should be something like this (syntax might be wrong)
```

#'  @importFrom MatrixGenerics colCumsum
.calc_css <- function(x, ...){

}
```

SHillman836: ok thanks will do

tests/testthat/test-5transformCounts.R
SHillman836: so yeah my method and metagenomeSeqs give quite different results. This is their code - https://github.com/HCBravoLab/metagenomeSeq/blob/master/R/cumNorm.R - which obviously has quite a few more steps than my current implementation. Should I just add those steps into my implementation in a way that fits with mia?

antagomir: Yep this is why we usually want to verify by comparing with external implementations. 

They use the same (compatible) open source license than we do, and @HCBravo is one of the original authors in the mia/OMA project, so I think you could even copypaste that (open source) code to mia, add SE support, and remember to cite the original source and authors in the references field.

antagomir: They seem to have fast variants, could you check if that's useful?

https://github.com/HCBravoLab/metagenomeSeq/blob/master/R/cumNormStatFast.R

SHillman836: yeah I was gonna ask you - I'm not too sure - I know that for alot of datasets the end value usually ends up being 0.5 anyway, so not sure how needed it is?

antagomir: Would it harm to allow users choose it if most of the time they do not have to worry about it when the default is good?

SHillman836: No it wouldn't harm, it just makes the method abit more complex, but I can add it - should I add it?

R/transformCounts.R
```diff
@@ -56,6 +56,36 @@
 #' 'relabundance', 'rrank', 'standardize', 'total': please refer to 
 #' \code{\link[vegan:decostand]{decostand}} for details.
 #' 
+#' \item 'css': Cumulative Sum Scaling (CSS) can be used to normalize count data 
+#' by accounting for differences in library sizes. [Paulson et al. (2013) Differential abundance analysis for microbial marker-gene surveys]
+#' (https://doi.org/10.1038/nmeth.2658) outlines this in more detail.
```
antagomir: Could we just cite here and put the more exact literature reference in the @reference section?

R/transformCounts.R
```diff
@@ -97,6 +127,14 @@
 #'                       
 #' head(assay(tse, "clr"))
 #' 
+#' # Perform CSS normalization.
+#' tse <- transformAssay(tse, method = "css")
+#' head(assay(tse, "css"))
+#' 
+#' # Perform CSS fast variant normalization.
+#' tse <- transformAssay(tse, method = "css_fast")
```
antagomir: should be css.fast according to our naming conventions.

antagomir (commented):
Seems good. Just minor comments

tests/testthat/test-5transformCounts.R
antagomir: Yes, this would be the standard way to do. We like to support user freedom even if the defaults are good.

R/transformCounts.R
SHillman836: ok thanks sorry haven't seen these changes till today, was travelling back to London yesterday, will make these changes and any other comments from the past few days this evening/tmrw morning if that's ok?

antagomir: yes, certainly - great if you can help to finalize!

man/transformAssay.Rd
TuomasBorman, line 1: When trying to open this file for preview, I get this error

Error: transformAssay.Rd:103: Bad \link text

DESCRIPTION
```diff
@@ -85,6 +85,7 @@ Imports:
     bluster,
     MatrixGenerics,
     mediation,
+    metagenomeSeq,
```
TuomasBorman: Is this package used oly in tests? If yes, then this should be in suggests

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
 
+####################################.calc_css###################################
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, css_percentile = NULL, ...) {
+    # Check for the presence of the scaling factor percentile and scaling 
+    # constant in additional args
+    if (is.null(css_percentile)) {
+      css_percentile <- ifelse("css_percentile" %in% names(list(...)), 
+                               list(...)$css_percentile, 0.5)
+    }
```
TuomasBorman, line 357: You are fetching the

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
     return(mat)
 }
 
+####################################.calc_css###################################
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, css_percentile = NULL, ...) {
```
TuomasBorman: This parameter name should be css.percentile or just percentile

Also add scaling_constant here. It should be scaling.constant or preferrably just scaling

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    # Check for the presence of the scaling factor percentile and scaling 
+    # constant in additional args
+    if (is.null(css_percentile)) {
+      css_percentile <- ifelse("css_percentile" %in% names(list(...)), 
+                               list(...)$css_percentile, 0.5)
+    }
+    scaling_constant <- ifelse("scaling_constant" %in% names(list(...)), 
+                               list(...)$scaling_constant, 1000)
+
```
TuomasBorman, line 360: What is the

R/transformCounts.R
TuomasBorman: What is the thing with this? css_percentile should already catch the value in the function definition?

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, css_percentile = NULL, ...) {
+    # Check for the presence of the scaling factor percentile and scaling 
+    # constant in additional args
+    if (is.null(css_percentile)) {
+      css_percentile <- ifelse("css_percentile" %in% names(list(...)), 
+                               list(...)$css_percentile, 0.5)
+    }
+    scaling_constant <- ifelse("scaling_constant" %in% names(list(...)), 
+                               list(...)$scaling_constant, 1000)
```
TuomasBorman: Move this to function definition

TuomasBorman: Add input check for scaling_constant and css_percentile

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    normalized_data <- sweep(mat, 2, scaling_factors / scaling_constant, "/")
+    return(normalized_data)
+}
+
+####################################.calc_css_fast##############################
+# This function applies cumulative sum scaling (CSS) to the abundance table 
+# using the fast variant method.
+.calc_css_fast <- function(mat, ...) {
+    # Check for the presence of the rel additional arg and set default if not present
+    rel <- ifelse("rel" %in% names(list(...)), list(...)$rel, 0.1)
```
TuomasBorman: Add this to function definition.

TuomasBorman: Add input check

R/transformCounts.R
TuomasBorman: And if the value is 0.5 by default (as it seems to be), it should be defined here -->

css_percentile = 0.5

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    
+    # Calculate the percentile using .calc_css_percentile
+    css_percentile <- .calc_css_percentile(mat, rel, ...)
+    
+    # Call .calc_css with the calculated percentile
+    normalized_data <- .calc_css(mat, css_percentile, ...)
```
TuomasBorman: What if user specifies the css_percentile? Then css_percentile is overwritten. Is that what you expect?

See:

`transformAssay(tse, method = "css_fast", css_percentile = 0.6)`

Value 0.6 will be used in calculation, not the value calculated here in this internal function

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+}
+
+#################################.calc_css_percentile######################
+# Calculates the cumulative sum scaling (css) scaling percentiles from the given 
+# data
+#'  @importFrom matrixStats rowMedians
+.calc_css_percentile <- function(mat, rel, ...) {
+    # Sort the non-zero values in each column of the matrix in decreasing order
+    smat <- lapply(1:ncol(mat), function(i) {
+      sort(mat[which(mat[, i] > 0), i], decreasing = TRUE)
```
TuomasBorman, line 391: Fix indentation, multiplication of 4 spaces

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    })
+    
+    # Find the maximum length of the sorted columns
+    leng <- max(sapply(smat, length))
+    
+    # Check if any column has only one or zero features and stop if true
+    if (any(sapply(smat, length) == 1)) stop("Warning: sample with one or zero features")
```
TuomasBorman, line 398: stop and warning calls should have `call. = FALSE`. Also check that this line does not exceed 80 characters

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    
+    # Compute the row medians of absolute differences
+    diffr1 <- matrixStats::rowMedians(abs(diffr))
+    
+    # Find the first point where the relative change exceeds the threshold (rel)
+    x <- which(abs(diff(diffr1)) / diffr1[-1] > rel)[1] / length(diffr1)
+    
+    # If the calculated percentile is less than or equal to 0.50, use 0.50 as the default value
+    if (is.na(x) || x <= 0.50) {
+      message("Default value being used.")
```
TuomasBorman, line 435: You could add what is the value

R/transformCounts.R
```diff
@@ -307,6 +346,120 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    smat2[is.na(smat2)] <- 0
+    
+    # Compute the row means of smat2 to create a reference profile
+    ref1 <- rowMeans(smat2)
+    
+    # Number of columns in rmat2
+    ncols <- ncol(rmat2)
+    
+    # Calculate the differences between the reference profile and each column's quantiles
+    diffr <- sapply(1:ncols, function(i) {
```
TuomasBorman, line 423: Instead of 1:10, use seq_len(10)

TuomasBorman: Also, based on Bioconductor recommendations, avoid sapply(). Use vapply or lapply

R/transformCounts.R
TuomasBorman: Also, as this is used only in .calc_css_percentile, this parameter should be defined there

R/transformCounts.R
```diff
@@ -307,6 +346,109 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    res <- which(res > rel)
+    res <- res[[1]] / max(found_features)
+    # If the calculated percentile is less than or equal to 0.50, use 0.50 as
+    # the default value
+    if( res < 0.50 ){
+        res <- 0.50
+        message("Default value being used (", res, ").")
+    }
+    return(res)
+}
```
TuomasBorman: I simplified the code and removed loops. The functionality should be the same. Check that it is still working as you expect

TuomasBorman:
Couple points from .calc_css_percentile function (I think that I fixed most of them already). The functions looked fine and it worked well, but there was room for improvement

1. Some of the indentations were incorrect. Should be multiplication of 4 spaces
2. Some lines were too long. Should be max 80 characters
3. Avoid loops. Vectorize if possible (faster than loops, and also good practice)
4. Instead of sapply, use vapply. (Comes from Bioconductor recommendations)
5. Instead of 1:10, use seq_len(10). (Comes also from Bioc)

antagomir:
@Daenarys8 could you add the suggested updates and make a PR against the SHillman836::adding-css-method branch? Or can @SHillman836 help to finalize this one?

0xMuluh:
metagenomeSeq is not on cran and that is messing up the checks. Experienced it before?

0xMuluh:
> metagenomeSeq is not on cran and that is messing up the checks. Experienced it before?

Actually it is there but since we ran workflow with devel where `metagenomeSeq` is currently failing build, gha could not add it. I have put temporal solution to run gha with release. We can restore devel later. It was either this or passing the env `_R_CHECK_FORCE_SUGGESTS_` as `false`  which perhaps is not the best ...

antagomir:
we can ask them?

R/transformCounts.R
```diff
@@ -24,12 +24,22 @@
...
+#' @param ... additional arguments.
 #' \itemize{
-#'   \item \code{reference}: \code{Character scalar}. use to
+#'   \item \code{reference}: \code{Character scalar}. Used to
 #'   to fill reference sample's column in returned assay when calculating alr. 
 #'   (Default: \code{NA})
 #'   \item \code{ref_vals} Deprecated. Use \code{reference} instead.
+#'   \item \code{css.percentile}: \code{Numeric scalar}.  Used to set the 
+#'   percentile value that calculates the scaling factors in the css normalization.
+#'   (Default: \code{"0.5"})
```
TuomasBorman: This should be numeric. "0.5" is character

R/transformCounts.R
```diff
@@ -24,12 +24,22 @@
...
+#'   \item \code{scaling}: \code{Integer scalar}. Adjusts the normalization scale 
+#'   by dividing the calculated scaling factors, effectively changing 
+#'   the magnitude of the normalized counts. (Default: \code{"1000"}).
```
TuomasBorman: Here also "1000"

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    # Input check
+    if( !.is_integer(scaling) ){
+        stop("'scaling' must be an integer value.", call. = FALSE)
+    }
+    if( !is.numeric(css.percentile) ){
+        stop("'css.percentile' must be a numeric value.", call. = FALSE)
+    }
+    # Call .calc_scaling_factors method to calculate scaling factors
+    scaling_factors <- .calc_scaling_factors(mat, css.percentile)
+
```
TuomasBorman: Remove extra line break

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
 
+####################################.calc_css###################################
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, css.percentile = 0.5, scaling = 1000, ...) {
+    # Input check
+    if( !.is_integer(scaling) ){
+        stop("'scaling' must be an integer value.", call. = FALSE)
+    }
+    if( !is.numeric(css.percentile) ){
+        stop("'css.percentile' must be a numeric value.", call. = FALSE)
```
TuomasBorman: This probably pass through vectors?

--> .is_an_integer and is.numeric() && length() == 1

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    
+    # Normalize the count data by dividing by the scaling factor
+    normalized_data <- sweep(mat, 2, scaling_factors / scaling, "/")
+    return(normalized_data)
+}
+
+####################################.calc_css_fast##############################
+# This function applies cumulative sum scaling (CSS) to the abundance table 
+# using the fast variant method.
+.calc_css_fast <- function(mat, css.percentile = NULL, ...) {
```
TuomasBorman, line 375: It seems that css.percentile is not checked?

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    # Calculate the percentile using .calc_css_percentile
+    if(is.null(css.percentile)) {
+        css.percentile <- .calc_css_percentile(mat, ...)
+    }
+    message("css.percentile value used (", css.percentile, ").")
```
TuomasBorman: Put this message inside above if sentence

TuomasBorman: If user has specified the css.percentile, they already know what it is

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    res <- res[[1]] / max(found_features)
+    # If the calculated percentile is less than or equal to 0.50, use 0.50 as
+    # the default value
+    if( res < 0.50 ){
+        res <- 0.50
+        message("Default rel value used (", res, ").")
+    } else {
+        message("rel value used (", res, ").")
+    }
+    return(res)
```
TuomasBorman: From where this 0.5 comes from?

R/transformCounts.R
```diff
@@ -131,20 +170,21 @@ setGeneric("transformAssay", signature = c("x"),
...
-             method = c("alr", "chi.square", "clr", "frequency", "hellinger",
-                        "log", "log10", "log2", "max", "normalize", "pa",
-                        "range", "rank", "rclr", "relabundance", "rrank",
-                        "standardize", "total", "z"),
-             MARGIN = "samples",
-             name = method,
-             pseudocount = FALSE,
-             ...){
+        assay.type = "counts", assay_name = NULL,
+        method = c("alr", "chi.square", "clr", "css", "css.fast", "frequency",
```
TuomasBorman: "css.fast" should probably be "css_fast". For instance, we have in alpha diversity indices "gini_dominance"

tests/testthat/test-5transformCounts.R
```diff
@@ -70,6 +70,112 @@ test_that("transformAssay", {
                          log2(x+5)
                      }), check.attributes = FALSE)
         
+        ############################ CSS ######################################
+        skip_if_not(require("metagenomeSeq", quietly = TRUE))
+        # Define counts matrix for the css and css.fast testing
+        counts_matrix <- as.matrix(assay(tse, "counts"))
```
TuomasBorman: If this package is not available, it skips all tests below. These tests should be at the end.

R/transformCounts.R
```diff
@@ -308,6 +351,115 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
     return(mat)
 }
 
+####################################.calc_css###################################
+# This function applies cumulative sum scaling (CSS) to the abundance table.
+.calc_css <- function(mat, css.percentile = 0.5, scaling = 1000, ...) {
+    # Input check
```
TuomasBorman: argument names could be just "percentile", "scaling" and "threshold"

TuomasBorman:
It seems that metagenomeSeq is only used in testing? We can probably just hard-code those values to testing file. That way we do not need this new dependency.

0xMuluh:
> we can ask them?

Sure. I found an existing issue from March this year and updated it. It seems the package had been inactive for a while.

R/transformCounts.R
0xMuluh: it is referenced from the percentile

R/transformCounts.R
0xMuluh: ACK

R/transformCounts.R
0xMuluh: ACK

R/transformCounts.R
0xMuluh: ACK.

R/transformCounts.R
0xMuluh: ACK.

R/transformCounts.R
0xMuluh: ACK.

R/transformCounts.R
0xMuluh: should we restrict percentile to integer?

TuomasBorman: No, I meant that you should use .is_an_integer(scaling) and is.numeric(css.percentile) && length(css.percentile) == 1

R/transformCounts.R
```diff
@@ -24,12 +24,22 @@
...
 #'
-#' @param ... additional arguments passed on to \code{vegan:decostand}:
+#' @param ... additional arguments.
 #' \itemize{
-#'   \item \code{reference}: \code{Character scalar}. use to
+#'   \item \code{reference}: \code{Character scalar}. Used to
 #'   to fill reference sample's column in returned assay when calculating alr. 
 #'   (Default: \code{NA})
 #'   \item \code{ref_vals} Deprecated. Use \code{reference} instead.
+#'   \item \code{percentile}: \code{Numeric scalar}.  Used to set the
```
TuomasBorman: Consider \code{percentile} \code{Numeric scalaer} (css)

--> It might be easier for user to see where this applies

R/transformCounts.R
TuomasBorman: The documentation says that the default value of percentile is 0.5. However, that is not true as it is calculated (it is >= 0.5)

I was also wondering what makes the 0,5 special number. Why it cannot be lower by default. However, I think that is ok since that is calculated only if user does not specify the value

R/transformCounts.R
0xMuluh: ACK.

R/transformCounts.R
0xMuluh: thanks.

R/transformCounts.R
```diff
@@ -308,6 +347,145 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+    res <- res[[1]]
+    # Convert the count value to a relative proportion by dividing by the
+    # maximum number of features. This scales the count to a proportion of the
+    # total feature count, normalizing it against the library size.
+    res <- res/max(found_features)
+    # # If the calculated percentile is less than or equal to 0.50, use 0.50 as
+    # # the default value
+    # if( res < 0.50 ){
+    #     res <- 0.50
+    # }
```
TuomasBorman: I did not find any information on why is this 0.5. @antagomir DO you have information on this? I think this is not transparent to hardcode the lower limit especially as this does not seem to be part of the algorithm

R/transformCounts.R
```diff
@@ -45,10 +57,18 @@
...
 #' 'relabundance', 'rrank', 'standardize', 'total': please refer to 
 #' \code{\link[vegan:decostand]{decostand}} for details.
 #' 
+#' \item 'css': Cumulative Sum Scaling (CSS) can be used to normalize count data 
+#' by accounting for differences in library sizes. By default, the function
+#' determines the normalization percentile for summing and scaling 
+#' counts. If you want to specify the percentile value, good default value
+#' might be \code{0.5}.The method is inspired by
+#' \code{\link[https://www.bioconductor.org/packages/metagenomeSeq/]{metagenomeSeq}}.
+#'
```
TuomasBorman: I removed "fast css" method as the only difference to this standard method is that the value of percentile is estimated from the data. Also "fast" did not describe the algorithm and 2 similar methods were more confusing.

Now percentile is estimated from the data by default, but user can choose manually the percentile also.

Moreover, our code is different from metagenomeSeq (it does essentially the same thing, though). That is why the method is only "inspired" by metagenomeSeq

0xMuluh: The hard-coded test fail now and the results are different. Should we check or go ahead with the new changes?

0xMuluh: it's all good. Updated the percentile to get the right matrix. everything works now.

0xMuluh:
It seems css and css_fast is already in mia?
@TuomasBorman

TuomasBorman (approved):

TuomasBorman:
> It seems css and css_fast is already in mia? @TuomasBorman

It did not include the functionality; it included only "css" as one of the options.  I am not sure how that ended up there, probably I copy-pasted the method-vector there incorrectly

Issue #489, opened 2024-02-13
Use `pseudocount=TRUE` in `transformAssay` even with missing or zero values


artur-sannikov:
**Is your feature request related to a problem? Please describe.**

At the moment, `transformAssay` [function](https://github.com/microbiome/mia/blob/master/R/transformCounts.R#L556) requires specifying a numerical value manually if there are negative or missing values in the assay. 

**Describe the solution you'd like**
Even if we have missing values or zero values, it's possible to calculate the minimum non-NA value when `pseudocount=TRUE`. That is required anyway if `The assay contains missing or negative values. "'pseudocount' must be specified manually`.

My solution is simple:

```R
# Logical vector to for all NAs and 0 values of relative abundance assay
relabundance_no_na_and_zero <- !is.na(assay(tse, "relabundance")) & assay(tse, "relabundance") != 0

# Minimum value of relative abundance assay divided by 2 to be added
# as pseudocount for calculation of centered log-ratio
# Only specifying pseudocount=TRUE for an automatic calculation does not work
# due to the presence of NAs
relabundance_pseudocount <- min(assay(tse, "relabundance")[
  relabundance_no_na_and_zero
]) / 2
```
The same solution can work also if there are negative values. We just need to add them in the logical vector.

antagomir:
Do you mean that `pseudocount=TRUE` does not work if the data contains zero or missing values? I am not sure if I understand whether the proposed solution here is to  replace `pseudocount=TRUE` automated calculation with more lengthy manual calculation?

artur-sannikov:
> Do you mean that pseudocount=TRUE does not work if the data contains zero or missing values? 

Sorry, I just realized I wrote 0 or missing values. I meant negative or missing values!

It will tell you that the data contains negative or missing values:

> Error: The assay contains missing or negative values. 'pseudocount' must be specified manually.

I then manually calculate the required pseudocount value (which is the minimum value of relabundance / 2). We can skip this step if `pseudocount=TRUE` calculates the minimum value or sets it to 0 even if we have missing or negative values like it's [already done](https://github.com/microbiome/mia/blob/master/R/transformCounts.R#L564) if there are not missing or negative values. That is what I do anyway in the code above.

> I understand whether the proposed solution here is to replace pseudocount=TRUE automated calculation with more lengthy manual calculation?

Yes, sort of. The automatic calculation at the moment assumes that there are no negative or missing values and sets the pseudocount to the minimum value:

```R
[...]

PR #615, opened 2024-07-24
ignore NAs when pseudocount is TRUE

0xMuluh:
Ignore NAs and perform calculation automatically.

ping: #489

### DIFF
--- a/R/transformCounts.R
+++ b/R/transformCounts.R
@@ -13,7 +13,8 @@
 #'   (Default: \code{"samples"})
 #' 
 #' @param pseudocount \code{Logical scalar} or \code{numeric scalar}. When TRUE,
-#'   automatically adds half of the minimum positive value of \code{assay.type}.
+#'   missing values(NAs) are ignored and, it automatically adds half of the 
+#'   minimum positive value of \code{assay.type}.
 #'   When FALSE, does not add any pseudocount (pseudocount = 0).
 #'   Alternatively, a user-specified numeric value can be added as pseudocount.
 #'   (Default: \code{FALSE}).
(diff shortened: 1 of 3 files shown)


R/transformCounts.R
```diff
@@ -361,15 +362,20 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
+        # If pseudocount TRUE but some negative values, numerical pseudocount needed
+        if ( pseudocount && any(mat < 0, na.rm = TRUE) ){
+            stop("The assay contains negative values. ",
                  "'pseudocount' must be specified manually.", call. = FALSE)
         }
         # If pseudocount TRUE, set it to  half of non-zero minimum value
         # else set it to zero.
         # Get min value
-        value <- min(mat[mat>0])
+        value <- min(mat[mat > 0], na.rm = TRUE)
```
antagomir: I think we could ultimately leave the decision to user so na.rm could be a parameter that is passed with ...

antagomir: but it could have default TRUE in the user function

antagomir (approved):
Seems good to me.

antagomir:
The unit test should test the actual calculation and compare it to cases that are calculated manually, checking that the real use scenarios work properly. Can you add that?

R/transformCounts.R
```diff
@@ -13,7 +13,8 @@
 #'   (Default: \code{"samples"})
 #' 
 #' @param pseudocount \code{Logical scalar} or \code{numeric scalar}. When TRUE,
-#'   automatically adds half of the minimum positive value of \code{assay.type}.
+#'   missing values(NAs) are ignored and, it automatically adds half of the 
+#'   minimum positive value of \code{assay.type}.
```
TuomasBorman: Missing values is just a side note. Better description might be

When \code{TRUE}, automatically adds half of the minimum positive value of \code{assay.type} (missing values ignored by default: \code{na.rm = TRUE}).

TuomasBorman (changes requested):
vegan::decostand has also na.rm parameter. Check that this is not causing problems. (I think this not causing problems even though same value handles both pseudocount and vegan::decostand)

R/transformCounts.R
0xMuluh: Thanks

R/transformCounts.R
```diff
@@ -359,17 +361,22 @@ setMethod("transformAssay", signature = c(x = "SummarizedExperiment"),
...
-        if ( pseudocount && (any(is.na(mat)) || any(mat < 0, na.rm = TRUE)) ){
-            stop("The assay contains missing or negative values. ",
+        # If pseudocount TRUE and some NAs, a warning is issued
+        if ( pseudocount && any(is.na(mat)) ){
+            warning("The assay contains missing values (NAs). These will be 
+                ignored in pseudocount calculation.", call. = FALSE)
+        }    
+        # If pseudocount TRUE but some negative values, numerical pseudocount needed
+        if ( pseudocount && any(mat < 0, na.rm = TRUE) ){
+            stop("The assay contains missing values (NAs) or negative values. ",
```
TuomasBorman: This message is now incorrect. It should be just "The assay contains negative values.."

After this fix, good to go

TuomasBorman (approved):

PR #616, opened 2024-07-25
NMF ordination with feature loadings computation

theotimeee:
This PR adds two new methods `getNMF` and `addNMF` that perform Non-negative Matrix Factorization.

`getNMF` returns the ordination matrix with the feature loadings matrix and NMF model stored as attributes.
`addNMF` adds this ordination matrix to `reducedDims` of the TreeSummarizedExperiment.

This PR is related to issue https://github.com/microbiome/mia/issues/596.


theotimeee:
I am facing an issue with `NMF` package. Apparently, `NMF` uses a `seed()` function but `SummarizedExperiment` package uses `DelayedArray` which also has a `seed()` function. Therefore when loading `SummarizedExperiment` package, the `seed()` function from `NMF` is overwrited.

The `getNMF `and `addNMF` functions work well if I load the `NMF` package after `mia`, but otherwise I get this error:
>Error in (new("standardGeneric", .Data = function (x)  : 
  unused arguments (model = list("NMFstd", 2, 0), method = "random")
Called from: do.call(getGeneric("seed"), c(list(x = x, model = init, method = seed.method), 
    parameters.seed))

This error is discussed in the following bioconductor post: https://support.bioconductor.org/p/110844/
and in this github issue: https://github.com/renozao/NMF/issues/85

antagomir:
They also seem to propose solutions. Are those not successful?

theotimeee:
> They also seem to propose solutions. Are those not successful?

Detaching and reloading the NMF package works well but the user would have to do it themselves, so it is not viable. 

The issue cited above was closed but I get the error even though I have the last version of the NMF package.

antagomir:
Is it possible to create a wrapper function with another name that calls it?

I expect that the notation package::function() does not suffice here to specify which package we want to take the function from?

R/addNMF.R
```diff
@@ -0,0 +1,104 @@
...
+              stop("'name' must be a non-empty single character value.",
+                   call. = FALSE)
+            }
+            # Fit the model
+            nmf <- getNMF(x, k = k, assay.type = assay.type, ...)
+            # Add scores matrix with loadings as attribute to reducedDims
+            x <- .add_values_to_reducedDims(x, name = name, values = nmf)
+            return(x)
+          }
+)
```
TuomasBorman, line 152: End files with empty line

R/addNMF.R
```diff
@@ -0,0 +1,104 @@
...
+            # store scores
+            scores <- nmf_model@fit@W
+            # Add loadings as attribute of the scores matrix
+            attr(scores, "loadings") <- nmf_model@fit@H
+            # Add NMF model as attribute of the scores matrix
+            attr(scores, "model") <- nmf_model
+            # Return scores with loadings, metrics and model as attribute
+            return(scores)
+          }
+)
```
TuomasBorman: INdentation is not correct (multiplication of 4 spaces)

TuomasBorman (changes requested):
Does the requested modification fix the problem?

One thing that you have not taken into account is multiple k values. NMF returns a list when k is vector of multiple values

- measures: goodness of fit for each k value
- consensus: how well the models with different k correspond each other
- models for each k

I think you could select the best fit based on certain metric and return that. Add all the other values (metrics, consensus, other models) to attributes. I think you could just add it as one attribute "NMF_output" or something.

Update documentation to reflect these changes so that user knows where to find the data.

theotimeee:
> Does the requested modification fix the problem?
> 
> One thing that you have not taken into account is multiple k values. NMF returns a list when k is vector of multiple values
> 
> * measures: goodness of fit for each k value
> * consensus: how well the models with different k correspond each other
> * models for each k
> 
> I think you could select the best fit based on certain metric and return that. Add all the other values (metrics, consensus, other models) to attributes. I think you could just add it as one attribute "NMF_output" or something.
> 
> Update documentation to reflect these changes so that user knows where to find the data.

I tried making a wrapper but it was still throwing the same error. With the commit you made, this issue seems to be fixed, thank you.

I am going to work on the multiple k values returned.

theotimeee:
In `getNMF`, I calculate NMF rank object for ranks from 2 to 10 and select the best fit based on explained variance, then calculate the NMF model for rank selected and return scores with NMF model and features loadings as attributes. This is not exactly what you asked but I think it could be easier to manipulate for the user, as it would be similar to `getLDA` returned value.

TuomasBorman:
I am still on holiday, I will check this early next week in detail. There is clear advantage in your implementation, however, this kind of waste resources. This can be a real problem if the datasets are big. Moreover, there are other more philosophical issues on not using resources as efficiently as possible.

Can you still try to manipulate multiple rank values to look like one coming from the method when only one rank value was specified?

You are correct that the output should look same for user despite the number of ranks  specified.

TuomasBorman:
After taking closer look to your solution, I found out possible problem with your suggested solution. nmf() can return 3 types of classes: NMFfit (when length(k)==1, and nrun==1), NMFfitX (when length(k)==1, and nrun>1), NMFList (when length(k)>1).
```

data("GlobalPatterns")
tse <- GlobalPatterns
mat <- t(assay(x))

# Calculate with multiple k values --> NMFList (with NMFfitX objects)
res <- NMF::nmf(mat, c(2, 3))
# Calculate with single k value --> NMFfit
res2 <- NMF::nmf(mat, 2)
# Calculate with multiple nrun --> NMFfitX
res3 <- NMF::nmf(mat, 2, nrun = 2)
# Calculate with multiple k-values but single nrun --> NMFList (with NMFfit objects)
res4 <- NMF::nmf(mat, c(2, 3), nrun = 1)
```


This means that the format of output can differ in any case (it is either NMFfit or NMFfitX). However, the only things that we need to find from the output are scores and weights.

Does this work:

```
eval.metric <- "evar"
eval_metrics <- c(
    "sparseness.basis",
    "sparseness.coef",
    "rss",
    "evar",
    "silhouette.coef",
    "silhouette.basis",
    "cophenetic",
    "dispersion")

mat <- t(assay(x))
k <- c(2,3)
res <- NMF::nmf(mat, k)

if( is(res, "NMF.rank") ){
    best_fit <- .get_best_nmf_fit(res, eval.metric, ...)
} else{
    best_fit <- res
}

scores <- best_fit@fit@H
weigths <- best_fit@fit@W

attr(scores, "weights") <- weigths
attr(scores, "NMF_output") <- res
if( is(res, "NMF.rank") ){
    attr(scores, "best_fit") <- best_fit
}



.get_best_nmf_fit <- function(res, eval.metric){

    maximize <- c(
        "sparseness.basis" = TRUE,
        "sparseness.coef" = TRUE,
        "rss" = FALSE,
        "evar" = TRUE,
        "silhouette.coef" = TRUE,
        "silhouette.basis" = TRUE,
        "cophenetic" = TRUE,
        "dispersion" = FALSE)
    maximize <- maximize[ eval.metric ]
    measures <- res[["measures"]]
    values <- measures[[eval.metric]]
    if( maximize ){
        FUN <- which.max
    } else{
        FUN <- which.min
    }
    ind <- FUN(values)
    #
    model <- res[["fit"]][[ind]]
    return(model)
}

```

theotimeee:
I implemented your solution for eval metrics in the code, I just modified the name of attribute "weights" to "loadings" so that it fits the LDA attribute. I also modified the documentation to take into account new possibilities.

TuomasBorman (approved):
Can you check that tests pass

TuomasBorman (approved):
Thanks, ready to go after checks are finished

TuomasBorman:
Are you interested in updating OMA? There is an example on NMF but it is using NMF::nmf directly

theotimeee:
> Are you interested in updating OMA? There is an example on NMF but it is using NMF::nmf directly

Yes I am going to open a PR

PR #619, opened 2024-07-31
Fix importTaxpasta

d-callan:
using taxpasta via nf-core/taxprofiler v1.1.8 and tried to use importTaxpasta to get a tse. I got an error like `type field has unsupported value` (see [here](https://github.com/joey711/biomformat/blob/0de05b1e78ea6ed42e079e07ccb13cecfd9eba6a/R/validity-methods.R#L66)). after some digging around i see it looks like the taxpasta biom leaves this field empty.

this pr represents a proposal to have this type default to 'OTU table'. it also proposes making `addHierarchyTree` optional when importing taxpasta data. my particular use case doesnt require it, and its a bit time consuming. 

i also notice addHierarchyTree seems to not handle fully unclassified entities well. Im considering improving that situation out of scope for this PR unless someone objects, and am happy to file an issue if i dont find one once i have more details. but its also influenced my desire to make it optional in this context, so thought id mention it here.


antagomir:
These suggestions sound good to me but it would be good to see what @TuomasBorman has to say.

R/importTaxpasta.R
```diff
@@ -46,6 +46,9 @@ importTaxpasta <- function(file) {
     if( !file.exists(file) ){
         stop("'", file, "' not found.", call. = FALSE)
     }
+    if (!addHierarchyTree %in% c(TRUE,FALSE)) {
+        stop("'addHierarchyTree' must be TRUE or FALSE.", call. = FALSE)
+    }
```
TuomasBorman: Use

if( .is_a_bool(add.tree) )

R/importTaxpasta.R
```diff
@@ -34,7 +34,7 @@ NULL
 #' @export
 #'
 #' @importFrom SummarizedExperiment rowData
-importTaxpasta <- function(file) {
+importTaxpasta <- function(file, addHierarchyTree = TRUE) {
```
TuomasBorman: We are using certain naming convention (I believe we have forgot to add it in guidelines) but:

Class: SummarizedExperiment
Exported function: exportedFunction
Internal function: .internal_function
Parameter: parameter.name

--> So this would be add.hierearchy.tree. However, I would prefer shorter name "add.tree"

R/importTaxpasta.R
```diff
@@ -104,6 +107,7 @@ importTaxpasta <- function(file) {
     format <- sprintf("Biological Observation Matrix %s.%s", vs[1], vs[2])
     format_url <- attr(h5array, "format-url")
     type <- attr(h5array, "type")
+    type <- ifelse(type == "", 'OTU table', type)
```
TuomasBorman: Can you add comment what it does here? Why are you overwriting the type from the file? It makes the debugging easier if something happens in the future

TuomasBorman: I mean you already explained it in this PR, but it is also good to add in the code

TuomasBorman: Also you are using "" and ''. To harmonize coding style, you could modify 'OTU table' to this "OTU table"

R/importTaxpasta.R
```diff
@@ -61,7 +64,7 @@ importTaxpasta <- function(file) {
         # Create rowData and rowTree
         rowData(tse) <- .create_row_data(biom, ranks)
         .set_ranks_based_on_rowdata(tse, set.ranks = TRUE)
-        tse <- addHierarchyTree(tse)
+	if (addHierarchyTree) tse <- addHierarchyTree(tse)
```
TuomasBorman: The indentation is off. This is inside if statement so the indentation should be +4 spaces

TuomasBorman (changes requested):
Looks good!

1. Can you add description for the new parameter

\code{Logical scalar}. Specifies whether to calculate and add hierarchy tree. (Default: \code{TRUE})

2. To DESCRIPTION file, add your name

TuomasBorman:
I think set.ranks should also be optional (not related to the issue, but might be good to fix here also). You are free to do that also if you want, but I can also do that after you have finished finalizing your code.

Your plan about addHiearchyTree: Sounds very good! We can continue this in separate issue/PR to discuss about the solution.

d-callan:
hi @TuomasBorman and thanks for the detailed feedback! I wasnt sure how people would feel about these changes and so didnt put any real effort into styling/ linting/ etc yet, but i will do that now. I like your suggestions, but wonder about `addHierarchyTree` -> `add.tree`.. funnily enough i usually prefer longer names than shorter for things. i like the explicitness of it, the clarity and readability, as i figure anyone who knows the method for addHierarchyTree will immediately guess what the similarly named parameter does and vice versa. ill make it `add.tree` if you like, but wonder if youd be alright w `add.hierarchy.tree`

R/importTaxpasta.R
d-callan: ive maintained the existing indentation here, just added the conditional before. are single-line if statements against mia convention?

d-callan: ^ gen q. i do appreciate convention

TuomasBorman: single-line if statements are ok if they are easy to read (as here). To be honest, I am not 100% sure what my comment is referring; this line is ok; sorry about confusion.

R/importTaxpasta.R
```diff
@@ -61,7 +68,7 @@ importTaxpasta <- function(file) {
         # Create rowData and rowTree
         rowData(tse) <- .create_row_data(biom, ranks)
         .set_ranks_based_on_rowdata(tse, set.ranks = TRUE)
-        tse <- addHierarchyTree(tse)
+	if (add.hierarchy.tree) tse <- addHierarchyTree(tse)
         # Agglomerate to all existing ranks
         tse <- agglomerateByRanks(tse, agglomerate.tree = TRUE)
```
TuomasBorman: agglomerate.tree --> update.tree (Can you change this line)

We updated parameter names this summer, and I believe this imporTaxpasta was merged during the naming process and we did not notice this earlier

TuomasBorman:
> hi @TuomasBorman and thanks for the detailed feedback! I wasnt sure how people would feel about these changes and so didnt put any real effort into styling/ linting/ etc yet, but i will do that now. I like your suggestions, but wonder about `addHierarchyTree` -> `add.tree`.. funnily enough i usually prefer longer names than shorter for things. i like the explicitness of it, the clarity and readability, as i figure anyone who knows the method for addHierarchyTree will immediately guess what the similarly named parameter does and vice versa. ill make it `add.tree` if you like, but wonder if youd be alright w `add.hierarchy.tree`

I understand your point about the importance of explicitness and transparency in naming. However, using add.tree aligns well with other existing parameters, such as update.tree. Additionally, the risk of misunderstanding is minimal because the documentation is intended to clearly explain the functionality. However, you pointed a problem in the documentation as it is not currently clearly explaining where the tree comes from. (Would you be able to fix this?) The documentation should describe that tree is calculated with addHierarchyTree. I guess the names are balance; they should be clear enough but not too long since they might be annoying for some users.

d-callan:
ok. i believe all comments addressed now, but let me know.

TuomasBorman (approved):
Looks very nice. Will be merged when checks pass. Thanks!

TuomasBorman:
@d-callan I was pushing to Bioconductor and noticed that the maximum size of file can be 5 MB. However, the biom file is over 10 MB. Can you subset it? Unfortunately, I did not remember this earlier.

d-callan:
hey. sry. this week has been awful at work and this got away from me. fixing now.

Issue #415, opened 2023-08-09
perSampleDominantFeatures works only with taxonomyRanks 


antagomir:
The 
perSampleDominantFeatures and addPerSampleDominantFeatures 

This works:

```
library(mia)
data(GlobalPatterns, package="mia")
tse <- GlobalPatterns
rowData(tse)$group <- as.character(sample.int(5, nrow(tse), replace = TRUE))
perSampleDominantFeatures(tse, rank="Class")
```

This gives error:
```
perSampleDominantFeatures(tse, rank="group")
```

> Error: 'rank' must be a value from 'taxonomyRanks()'

We recently implemented the `mergeFeatures` functions so that they also allow other fields than ranks.  Could we enable it here as well?

antagomir:
A related issue for `addPerSampleDominantFeatures` is that it adds the new field in colData with the sample names. The sample names are unnecessary (since it is clear from the context in which sample each of them belongs to) and may disturb downstream analyses. 

However, this can be justified when there are samples with multiple features that are all dominant. Then this is a list, where each (sample) element may have multiple dominant features.

Consider adding an argument that picks a single dominant feature at random in such cases. Not necessarily a good idea but to discuss.

antagomir:
@Daenarys8 could you check if this is readY?

Daenarys8:
I think for the first action point we already accept names other than ranks;

```
library(mia)
data(GlobalPatterns, package="mia")
tse <- GlobalPatterns
tse <- addDominant(tse, name = "some_name")

```
As for the the second point, it's still unclear to me or is it open for discussion?

PR #623, opened 2024-08-07
accept other than ranks in addDominant

0xMuluh:
ping: #415


antagomir:
When the functionalitys is modified/extended, it is good to do the following by default:

- include relevant new unit tests 
- update the manpage documentation

Is there something we could have on that?

0xMuluh:
good to check

R/getDominant.R
```diff
@@ -73,7 +74,7 @@ setGeneric("getDominant",signature = c("x"),
 #' @export
 setMethod("getDominant", signature = c(x = "SummarizedExperiment"),
     function(x, assay.type = assay_name, assay_name = "counts", 
-             rank = NULL, other.name = "Other", n = NULL, complete = TRUE, ...){
+             rank = NULL, other.name = "Other", n = NULL, complete = TRUE, by = NULL, f = NULL, ...){
```
TuomasBorman: There is no use for "by"

R/getDominant.R
```diff
@@ -90,7 +91,13 @@ setMethod("getDominant", signature = c(x = "SummarizedExperiment"),
         if(!is.null(rank)){
             x <- agglomerateByRank(x, rank, ...)
             mat <- assay(x, assay.type)
-        } # Otherwise, if "rank" is NULL, abundances are stored without ranking
+        # or factor that is specified by user
+        } else if (!is.null(f)) {
```
TuomasBorman: How about just using "rank" without "by" and "f"

-->


if( !is.null(rank) %% rank %in% taxonomyRanks(tse)){
aggglomerateByRank
} else if( !is.null(rank) ){<
# Catch f and by so that they are not fed to agglomerateByVariable
agglomerateBYVariable
}

TuomasBorman (changes requested):

R/getDominant.R
0xMuluh: > # Catch f and by so that they are not fed to agglomerateByVariable

I didn't get this part

0xMuluh: did you mean?

```
if( !is.null(rank) && rank %in% taxonomyRanks(tse)){
            x <- agglomerateByRank(x, rank, ...)
        } else if (!is.null(rank)) {
            x <- agglomerateByVariable(x, "rows", rank , ...)
        }
        else {
       ...
        }
```

antagomir:
Confirm when this is ready.

0xMuluh:
> Confirm when this is ready.

It is ready. Waiting on @TuomasBorman review.

TuomasBorman (approved):

antagomir:
The documentations explains now: "With \code{rank} parameter, it is possible to agglomerate taxa based on taxonomic
ranks. E.g. if 'Genus' rank is used, all abundances of same Genus are added
together, and those families are returned. See \code{agglomerateByRank()} for
additional arguments to deal with missing values or special characters."

It might be helpful to also mention that in addition to "rank", other grouping variables would be accepted.

In fact, "rank" is somewhat misleading argument name if any grouping variable is accepted. Should we consider renaming this argument (as it is not only about ranks)?

TuomasBorman:
> The documentations explains now: "With \code{rank} parameter, it is possible to agglomerate taxa based on taxonomic ranks. E.g. if 'Genus' rank is used, all abundances of same Genus are added together, and those families are returned. See \code{agglomerateByRank()} for additional arguments to deal with missing values or special characters."
> 
> It might be helpful to also mention that in addition to "rank", other grouping variables would be accepted.
> 
> In fact, "rank" is somewhat misleading argument name if any grouping variable is accepted. Should we consider renaming this argument (as it is not only about ranks)?

At least the documentation should clearly mention what the function does. `rank` might still be good parameter name if the main purpose of the function is to agglomerate the data to taxonomy level and other groupings are rarely used

antagomir:
> At least the documentation should clearly mention what the function does. `rank` might still be good parameter name if the main purpose of the function is to agglomerate the data to taxonomy level and other groupings are rarely used

- `agglomerateByVariable()` is using argument name "f"; 
- `agglomerateByRank()` is using argument name "rank"

Seems to me that here the argument name "f" would be closer to the intended purpose, although it is not very clear as an argument name.

TuomasBorman:
I believe the `f` comes from base functions where f is used to define grouping factors https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/split

I do not have strict opinion on this. However, I think we should either use `f` or `rank` or then we should modify all `f` arguments in mia (maybe `group`). There are advantages when we use same argument names wherever possible.

0xMuluh:
Hmm, so are we going with `group` as the argument in addDominant?

antagomir:
`group` would be more clear than `f` but the latter (f) has advantage that it could be in many cases a passed variable and thus add less (?) maintenance. I agree that we should then aim to harmonize this across all functions.

0xMuluh:
should i amend changes in new PR or current PR? @TuomasBorman

TuomasBorman:
I think we can merge this and implement the other changes in other PR. scuttle package uses `f` as we currently. Of course, we can make our own decisions.

Also, because we have kind of changed the meaning of `f`, we should change the argument name (usually `f` is vector, but in mia it can be also a name of grouping variable).

However, I think it is little bit problematic if getDominant() does different things even thought the parameters are the "same". I am referring here that the agglomeration is done differently if the grouping variable is taxonomy rank or not. The problem is that this is not transparent for user.

```
# Uses agglomerateByRank
getDominant(tse, group = "Genus")
# Uses agglomerateByVariable
getDominant(tse, group = "bacteria_group") 
```

At least, we should notice this problem and document this as clearly as possible.

1. Change the argument name `f` -> `group` everywhere in mia
2. Change `rank` parameter in getDominant to `group`.

0xMuluh:
I think this good. I have opened new issue for suggested changes.

TuomasBorman (approved):

PR #625, opened 2024-08-09
spell checker

0xMuluh:
- minor spelling errors in comments and some docs


R/getPrevalence.R
```diff
@@ -109,7 +109,7 @@
 #'                                       sort = TRUE)
 #' head(prevalence.frequency)
 #'
-#' # Get prevalence estimates for phylums
+#' # Get prevalence estimates for phylum
```
antagomir: The plural is "phyla"

R/getPrevalence.R
```diff
@@ -228,7 +228,7 @@ setMethod("getPrevalence", signature = c(x = "ANY"), function(
 .agg_for_prevalence <- function(
         x, rank, relabel = FALSE, make.unique = TRUE, na.rm = FALSE,
         agg.na.rm = TRUE, ...){
-    # Check na.rm. It is not used in this function, it is only catched so that
+    # Check na.rm. It is not used in this function, it is only cached so that
```
antagomir: are you sure that "cached" is what is meant here - can you double check?

antagomir (commented):
some small comments

R/getPrevalence.R
0xMuluh: ACK

R/getPrevalence.R
0xMuluh: Right, the word should be caught. Thanks

PR #626, opened 2024-08-10
Give informative error when values are missing in addRDA

### LINKED ISSUE #432: Suggestion to remove `na.action` argument from `runRDA`
Hi!

What do you think about removing the `na.action` argument from `runRDA` and instead providing instructions in OMA how users can filter missing values themselves?

This could simplify the code of `runRDA` and prevent bugs due to differing dimensionality between tse and rda object.

RiboRings:
Fix for issue https://github.com/microbiome/mia/issues/432.

Usage examples:

```
# Import TreeSE
library(mia)
data("enterotype", package = "mia")
tse <- enterotype

# Throw error when na.action is na.fail and some values are missing

tse <- addCCA(tse, formula = assay ~ ClinicalStatus + Gender + Age)
# Error: Variables contain missing values. Set na.action to na.exclude to remove
# samples with missing values.

tse <- addRDA(tse, formula = assay ~ ClinicalStatus + Gender + Age,
              FUN = vegan::vegdist, method = "bray")
# Error: Variables contain missing values. Set na.action to na.exclude to remove
# samples with missing values.

# Work as usual when na.action is na.omit or na.exclude

tse <- addCCA(tse, formula = assay ~ ClinicalStatus + Gender + Age,
              na.action = na.exclude)

tse <- addRDA(tse, formula = assay ~ ClinicalStatus + Gender + Age,
              FUN = vegan::vegdist, method = "bray",
              na.action = na.exclude)
```


antagomir:
Thanks!

Original data dimension
> dim(enterotype)
[1] 553 280

Dimension after RDA (or CCA):
> dim(reducedDim(tse, "RDA"))
[1] 280   6

Ok the samples match and they are not dropped. This solution seems OK to me.

We could consider providing an optional (hidden?) argument to drop out samples with missing values but perhaps this is not critical for now.

Confirm that we have sufficient unit tests in place, and documentation is clear about the key issues related to this one.

I am wondering where the 6 component solution is defined, why the default is 6 dimensions and is it possible to change that somehow? (not really related to this issue but it would be useful to find out - if time not available now, perhaps open another issue on that?).

RiboRings:
> We could consider providing an optional (hidden?) argument to drop out samples with missing values but perhaps this is not critical for now.

I think this can be simply done by using `na.omit` instead of `na.exclude`, so in a sense the argument is already there.

antagomir:
> > We could consider providing an optional (hidden?) argument to drop out samples with missing values but perhaps this is not critical for now.
> 
> I think this can be simply done by using `na.omit` instead of `na.exclude`, so in a sense the argument is already there.

Right, sounds good.

Could/should we mention this explicitly in the documentation (@details section perhaps)?

Good to check that sufficient unit tests are in place.

antagomir:
Resolve conflicts.

TuomasBorman (approved):
Seems good!

R/runCCA.R
```diff
@@ -180,19 +181,27 @@ setGeneric("addRDA", signature = c("x"),
     #
     x <- as.matrix(t(x))
     variables <- as.data.frame(variables)
+    
+    if( any(is.na(variables)) && isTRUE(all.equal(na.action, na.fail)) ){
+      stop("Variables contain missing values. Set na.action to na.exclude",
+           " to remove samples with missing values.", call. = FALSE)
+    }
```
TuomasBorman, line 188: SHould this message be something else? That is because the option can be also na.omit. We do not want to make the decision for user.

TuomasBorman:
> > We could consider providing an optional (hidden?) argument to drop out samples with missing values but perhaps this is not critical for now.
> 
> I think this can be simply done by using `na.omit` instead of `na.exclude`, so in a sense the argument is already there.

That works only for get* functions. When add* function is used, the results is added with [.add_object_to_reduceddim](https://github.com/microbiome/mia/blob/89bad8e1fa1f714f8d140c5757b84865407c227e/R/runCCA.R#L378). That function checks if the result is missing samples compared to the TreeSE. If the result is missing samples and `subset_result = FALSE`, this functions adds samples to result to match with TreeSE.

However, that might not be the behavior that user wants. (If the result is missing samples, user specified na.omit and that is what user wants.)

So I suggest that we

1. replace default value of `subset_result` to TRUE
2. Rename `subset_result` --> `subset.result`


`subset_result` is not documented anywhere and it is intended just for us to control the behavior. Otherwise, the function should work (check).

antagomir:
ok

RiboRings:
Hi! I got back to this PR.

I noticed that the example with GlobalPatterns is very slow. Is it ok if I change the dataset to enterotype?

RiboRings:
This PR is ready to merge from my side.

TuomasBorman (approved):

PR #631, opened 2024-08-17
smaller test file for type-less taxpasta biom

d-callan:
my bad. i remember thinking i needed to do this while working https://github.com/microbiome/mia/pull/619 and then forgot.


antagomir:
Hmm there seems to be some problem in checks:

Error: The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌
Notice: Deployment failed! ❌

TuomasBorman (approved):
The checks are fine. The deployment error occurs because this branch cannot deploy (only devel can)

Issue #634, opened 2024-08-28
dbRDA and loadings


TuomasBorman:
It seems that the results of `getRDA()` function are somewhat incorrect. First of all, it does not return "loadings" as the "rotations" are empty. Moreover, the function seems to try to add "species scores" to place of "rotations" which is incorrect as they are different things.

Loadings = how much environmental variables such as diagnosis or sample type associates with each axes
Site scores = the coordinates of sample in ordination space
Species scores = the coordinates of taxa in ordination space (can be interpreted as association between axes and taxa, we can plot them with plotLoadings)

```
library(mia)
data("enterotype", package = "mia")
tse2 <- enterotype
tse2 <- transformAssay(tse2, method = "relabundance")
tse2 <- addRDA(
    tse2,
    assay.type = "relabundance",
    formula = assay ~ ClinicalStatus + Gender + Age,
    na.action = na.exclude)

# Empty
attr(reducedDim(tse2, "RDA"), "rotation")

# Get rda object
rda <- attr(reducedDim(tse2, "RDA"), "rda")

scores <- vegan::scores(rda)
# Loadings
scores$biplot
# Site scores
scores$sites
# Species scores
scores$species

```

@antagomir

0xMuluh:
hmm, `vegan::dbrada` initially does not give species scores and we have to use an accessor(`sppscores`) to add it.
```
....
 attr(ans, "rotation") <- rda$CCA$v
....
```

```
loadings_matrix <- attr(reducedDim(tse2, "RDA"), "rotation")
plotLoadings(loadings_matrix)
```

![Rplot](https://github.com/user-attachments/assets/01ad7da9-ebe3-4a4f-9514-7dfc9d132209)

@TuomasBorman

PR #637, opened 2024-09-10
get variable scores form sppscores in dbRDA

0xMuluh:
ping #634


TuomasBorman:
This is already taken into account in PR https://github.com/microbiome/mia/pull/639

PR #639, opened 2024-09-12
Update RDA/CCA functions

TuomasBorman:
Hello,

this PR is partly overlapping with this one: https://github.com/microbiome/mia/pull/637

I completely rewrote *RDA and *CCA functions. That is for easier maintenance and to make the code simpler. I have been looking this file a while and thought that it is far more complex than it could be.

Keypoints:

1.

RDA and CCA functions use same code. The interface to vegan is similar in both methods, so it is easier for us to use same code.

2.

It was little but unclear what was the output as we use different naming scheme than vegan. Now we use directly the output from vegan. Moreover, scores parameter was kind of useless, so it is now removed (not deprecated since I think there was no clear use)

Thing to do:

1. Update documentation

a)

Clearly state that these a wrappers for vegan::cca and vegan::dbrda.

b)

Clearly state that the output matrix is site scores. The attributes include the output from vegan::scores, eigenvalues, the object and possible significance testing results.

2.Check that scores parameter is removed from test file.

3. Fix plotRDA and plotLoadings

They are expecting that attr(mat, "rda") includes the object. However, now it is in attr(mat, "obj") (Check if the object is found from "obj", "rda" or "cca" --> no visible change to user)

User probably might want to plot either biplot values or species scores. We could have a parameter for this. Take into account that in PCA the "loadings" are in "rotation" attribute.

3 Check if scater package have a method for plotting eigenvalues.

4 Run BiocCheck and fix everything related to this file

@Daenarys8 Do you have time to implement these?


R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
-    # Create a formula based on variables
-    formula <- as.formula(paste0("x ~ ", 
-                                 paste(variables, collapse = " + ")))
-    formula
-}
+    #
+    res <- .calculate_rda(
+        x, formula = formula, data = data, ord.method = "CCA", ...)
+    return(res)
+    })
```
TuomasBorman, line 212: Method for matrix. It was not too usable previously

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
 #' @rdname runCCA
 setMethod("getCCA", "SummarizedExperiment",
-    function(x, formula, variables, test.signif = TRUE,
-             assay.type = assay_name, assay_name = exprs_values, exprs_values = "counts",
-             scores = "wa", ...)
-    {
-        # Check assay.type and get assay
+    function(
+        x, formula = NULL, col.var = variables, variables = NULL,
+        test.signif = TRUE, assay.type = assay_name, assay_name = exprs_values,
```
TuomasBorman, line 219: vairables are renamed to col.var. That is because they are column names from colData

TuomasBorman: We also use col.var in other functions

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
-        
+        ########################### Input check end ############################
+        # Get assay
+        mat <- assay(x, assay.type)
+        # Get formula and variables
+        temp <- .get_formula_and_covariates(x, formula, col.var)
+        formula <- temp[["formula"]]
+        covariates <- temp[["variables"]]
+        # Calculate CCA
+        cca <- getCCA(mat, formula = formula, data = covariates, ...)
```
TuomasBorman, line 246: We use matrix method to avoid overlapping code

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
+        formula <- temp[["formula"]]
+        covariates <- temp[["variables"]]
+        # Calculate CCA
+        rda <- getRDA(mat, formula = formula, data = covariates, ...)
+        # Test significance if specified
+        if( test.signif ){
+            res <- .test_rda(mat, attr(rda, "obj"), covariates, ...)
+            attr(rda, "significance") <- res
+        }
+        return(rda)
```
TuomasBorman, line 361: The structure is same as in getCCA

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
+
+# This function performs dbRDA or CCA. It returns side scores with other
+# information scores in attributes.
+#' @importFrom stats as.formula na.fail
+#' @importFrom vegan cca dbrda sppscores<- eigenvals scores
+.calculate_rda <- function(
+    x, formula, data, scores, scale = TRUE, na.action = na.fail,
+     method = distance, distance = "euclidean", ord.method = "CCA", ...){
+    # input check
+    if(!.is_a_bool(scale)){
```
TuomasBorman: We use same function for calling vegan::cca and vegan::dbrda

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
+    # Ensure that the left hand side of the formula points to abundance matrix
+    formula <- as.character(formula)
+    formula[[2]] <- "x"
+    # Create a formula from string
+    formula <- as.formula(
+        paste(as.character(formula)[c(2,1,3)], collapse = " "))
+    
+    # Initialize an argument list with common arguments
+    args <- c(
+        list(formula = formula, data = data, na.action = na.action), list(...))
```
TuomasBorman, line 499: We use always formula. If user do not want to incorporate variables, the formula is x ~ 1

R/runCCA.R
```diff
@@ -157,311 +155,394 @@
...
+    }
+    
+    # Get eigenvalues from the object
+    eig <- eigenvals(res_obj)
+    # Get total number of coordinates. There might be imaginary axes, exclude
+    # them because otherwise error occurs in scores() call.
+    tot_num_coord <- sum(eig >= 0)
+    # Now we know how many coordinates there are. We use that info to get
+    # scores for all coordinates.
+    res_values <- vegan::scores(res_obj, seq_len(tot_num_coord))
```
TuomasBorman, line 542: The output is from vegan::scores

tests/testthat/test-6cca.R
```diff
@@ -16,42 +16,38 @@ test_that("CCA", {
...
     actual <- reducedDim(sce,"CCA")
-    expect_equal(as.vector(actual), as.vector(mcca$CCA$wa))
+    ref <- mcca$CCA$wa
+    actual <- actual[, seq_len(ncol(ref)), drop = FALSE]
+    expect_equal(actual, ref)
     #
     mcca <- vegan::cca(form, dune.env, scale = TRUE)
     mrda <- vegan::rda(form, dune.env, scale = FALSE)
     
-    sce <- addCCA(sce, form, scores = "u")
```
TuomasBorman, line 46: scores is now removed as the output is now always from vegan::scores

tests/testthat/test-6cca.R
```diff
@@ -87,48 +83,54 @@ test_that("CCA", {
...
+    ref <- vegan::scores(rda_bray, display = "sites")
+    actual <- actual[, seq_len(ncol(ref)), drop = FALSE]
+    expect_equal(as.vector(actual), as.vector(ref))
     #
-    sce <- addRDA(sce, scores = "u")
+    sce <- addRDA(sce)
     test <- reducedDim(sce,"RDA")
     # Test that eigenvalues match
-    test <- attr(test, "rda")$CA$eig
+    test <- attr(test, "obj")$CA$eig
```
TuomasBorman, line 112: The object can be now found from attr(mat, "obj")

0xMuluh:
GTG

0xMuluh:
> User probably might want to plot either biplot values or species scores. We could have a parameter for this. Take into account that in PCA the "loadings" are in "rotation" attribute.
> 
new param should be added to plotLoadings?
> 3 Check if scater package have a method for plotting eigenvalues.
> 
Currently, scater does not support a method for plotting eigenvalues
> 4 Run BiocCheck and fix everything related to this file
> 
This file is now updated and works fine.

TuomasBorman:
> > User probably might want to plot either biplot values or species scores. We could have a parameter for this. Take into account that in PCA the "loadings" are in "rotation" attribute.
> 
> new param should be added to plotLoadings?
> 
> > 3 Check if scater package have a method for plotting eigenvalues.
> 
> Currently, scater does not support a method for plotting eigenvalues
> 
> > 4 Run BiocCheck and fix everything related to this file
> 
> This file is now updated and works fine.

new parameter should be added to both plotLoadings and plotRDA (maybe hidden as it might be hard for user to understand from where the values are fetched). The function could check if there are "rotation" or "loadings" or "species" avilable if the parameter is NULL.

We can consider if we should add plotEigenvalues()

0xMuluh:
> new parameter should be added to both plotLoadings and plotRDA (maybe hidden as it might be hard for user to understand from where the values are fetched). The function could check if there are "rotation" or "loadings" or "species" avilable if the parameter is NULL.
> 
right. plotLoadings has some basic layouts(barplot, heatmap). 
1. Should we consider adding scatter plots as in `plotCCA`  for the loadings plots? 
2. Alternatively we could have `plotPCA() ` to seperate the function. (there's wrapper for this)
> We can consider if we should add plotEigenvalues()

is the target scree plots?

TuomasBorman:
1. I think no. The plot type is different. Also, I am not sure how often features are plotted as scatter plot. I have seen those in RDA context, only.

2. Yes, scree plot.  However, do not yet implement it, it might be good idea to ask Leo if he has opinion. That might be relatively straight-forward.

It could have options show.bar (show bar plot), show.line (show line plot), and show.label (show percentages as label). Also, it might be good to be able to determine how many components to plot.

Found from internet one example
image

TuomasBorman:
This is good to go, I think. @antagomir Do you have any comments?

1. We removed scores parameter since it did not really have use case. (Not deprecated)
2. Returned values follow the same terminology as in vegan. 
3. Previously, the cca/rda object was stored to attributes "rda"/"cca". Now it is stored in attributes "obj". (This might be problematic if user's workflow relies that the object is in old place)

I think these aforementioned things are not that big; we can change these without deprecation/warnings.

TuomasBorman:
@antagomir

antagomir:
Hmm seems good to go and more will be learnt when testing these in real use?

OK to implement the screeplot, there could be some use for it.

PR #640, opened 2024-09-17
Alphabetize packages

artur-sannikov:
Alphabetize packages in DESCRIPTION.


artur-sannikov:
The workflow fails ` The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌`

Can be a problem with a token?

TuomasBorman (approved):
Thanks!

The problem is that this branch cannot deploy the newly created documentation to mia reference page.

PR #650, opened 2024-10-17
Improve support for spaces in variable names

TuomasBorman:
I noticed that formula creation in dbRDA can lead to an error when variable names have spaces. Now it is ensured that, the formula works with spaces.

Moreover, `meltSE` replaced spaces with dots. Now the data is melted so that variable names are kept untouched.


TuomasBorman:
In theory, the formula shoudl work. However, in here https://github.com/microbiome/miaViz/pull/156, there are problems in facetting and formula formatting. Check that everything works.

R/meltAssay.R
```diff
@@ -14,263 +15,293 @@
...
+    # Get assay, ensure that it is a matrix. Ensure that it has row and colnames
+    # from the TreeSE
     mat <- assay(x, assay.type) %>%
         as.matrix() 
     rownames(mat) <- rownames(x)
     colnames(mat) <- colnames(x)
-    mat %>%
-        data.frame(check.names = check.names) %>%
+    # Convert it to long format
+    mat <- mat %>%
```
TuomasBorman: Replace %>% with |>

TuomasBorman:
> In theory, the formula shoudl work. However, in here [microbiome/miaViz#156](https://github.com/microbiome/miaViz/pull/156), there are problems in facetting and formula formatting. Check that everything works.


It seems that the problem in miaViz was related to different thing. I will still keep this as a draft since checks are failing.

TuomasBorman:
I confirm that everything works now

PR #657, opened 2024-11-05
Add PERMANOVA functions

TuomasBorman:
Add wrappers for `vegan::adonis2()`


antagomir:
PERMANOVA (adonis2) and dbRDA are closely related. 

It might be useful to cross-reference them and mention (perhaps also demonstrate) situation where they can be used for essentially similar purpose (e.g. two-group comparison). And perhaps then also situations where they differ.  

PERMANOVA is non-parametric and makes less assumptions, (db)RDA has a bit wider scope overall.


```
library(vegan)
data(varespec, varechem)
g <-  sample(1:2, 24, replace=TRUE)

## dbrda
res.rda <- dbrda(varespec ~ g, dist="bray")
anova(res.rda)

res.ado <- adonis2(varespec ~ g, method = "bray")
res.ado

```

antagomir (approved):

Issue #654, opened 2024-11-01
agglomerateByVariable should have `na.rm` parameter?


artur-sannikov:
`agglomerateByRank` has `na.rm` parameter, but `agglomerateByVariable` does not. Is there a reason for this?

```R
library(mia)
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns

tse_sub <- agglomerateByVariable(tse, by = "cols", group = "SampleType",
                                 na.rm = TRUE) # Won't work
tse_sub
```

```
Error in .local(x, ...) : unused argument (na.rm = TRUE)
```

What if there are some NA values in microbial abundances?

I'm using mia_1.13.45.

antagomir:
I agree it would be logical to have similar behavior for both. The na.rm should be dealth with argument passing.

That being said, should be investigated more closely.

TuomasBorman:
`agglomerateByVariable` uses `scuttle::summarizeAssayByGroup` (for cols) and `scuttle:sumCountsAcrossFeatures` (for rows) internally. These functions do not have `na.rm` parameter.

For columns, this can be easily fixed: just add `na.rm`: If `na.rm=TRUE`, all `NA` values will be converted to 0.

For rows, we cannot do that directly, because `sumCountsAcrossFeatures` can also calculate average, and converting NA to 0 would lead to wrong answer.

--> requires some thinking what to do

TuomasBorman:
Thing to note is that `na.rm` in `agglomerateByRank` is controlling whether we remove those rows that do not have taxonomy information in specified taxonomy level. Summing NAs still lead to NA:
```

> agglomerateByRank(tse, rank = "Kingdom", na.rm = TRUE) |> assay()
            CL3     CC1    SV1 M31Fcsw M11Fcsw M31Plmr M11Plmr F21Plmr M31Tong M11Tong LMEpi24M SLEpi20M  AQC1cm  AQC4cm  AQC7cm    NP2     NP3     NP5 TRRsed1 TRRsed2
Archaea      NA    1248  28811      33      57      42     112     140     303      30      131      145    4459   24692   28051   1826   43197   33996     843    8418
Bacteria 862815 1134209 668698 1543418 2076419  718901  433782  186157 2000099  100157  2117461  1217167 1163289 2332489 1671242 521808 1435768 1618758   57845  484708
         TRRsed3   TS28    TS29   Even1  Even2   Even3
Archaea    14250   1598    1690     150     23      91
[...]

PR #658, opened 2024-11-11
agglomeration functions and na.rm

TuomasBorman:
In this issue https://github.com/microbiome/mia/issues/654, we noticed that the use of `na.rm` is not intuitive in agglomeration functions. 

This PR adds option for excluding NAs in agglomeration functions (`agglomerateByRank`, `agglomerateByVariable` and `agglomerateByPrevalence`). `na.rm` now works similarly to `sum(x, na.rm = TRUE)` in these aforementioned functions.

In `agglomerateByVariable` `na.rm` controlled whether we remove those rows that did not have any group information, i.e., their group was `NA`. Now this parameter is named `group.rm`.

In `agglomerateByRank` `na.rm` controlled whether we remove those rows that did not have any taxonomy information. Now this parameter is called `empty.rows.rm`.

These parameter names can be still modified. Maybe we could unify so that both would have parameter `empty.row.rm`.

`na.rm` is now handled in mia, but I opened issue in `scuttle` if this can be handled directly in function that mia utilizes: https://github.com/LTLA/scuttle/issues/33

Let's wait until the issue in scuttle is solved.


antagomir:
How about `empty.group.rm` or (`group.na.rm` and `row.na.rm`) to enhance consistency?

TuomasBorman:
row.rm might also be one option to keep it short. I think we could use same name for both

antagomir:
ok to me as long as it is consistent.

TuomasBorman:
The parameter name for removing NA rows is `empty.rm`

TuomasBorman:
After discussing in scuttle package, it was decided that na.rm will not be implemented in scuttle. This is why it is implemented in mia. mia uses scuttle function by default. If `na.rm=TRUE` and there are NAs, mia uses own implementation

PR #662, opened 2024-12-07
Incorporate Bioconductor recommendations

TuomasBorman:
This PR incorporates Bioconductor recommendations for coding practice. It does not modify the functionality.


TuomasBorman:
The checks are failing due to installation problems: "package ‘phyloseq’ is not available for this version of R". This same error is preventing OMA to build. I thought that this will go away automatically at some point (quite odd error...), but it seems not. Have to check in more detail how to fix this.

TuomasBorman:
> The checks are failing due to installation problems: "package ‘phyloseq’ is not available for this version of R". This same error is preventing OMA to build. I thought that this will go away automatically at some point (quite odd error...), but it seems not. Have to check in more detail how to fix this.

It seems that there are fail in phyloseq build (some error in their vignette). Hopefully, this will be fixed soon: https://bioconductor.org/checkResults/devel/bioc-LATEST/phyloseq/

TuomasBorman:
https://github.com/joey711/phyloseq/pull/1777

TuomasBorman:
I went through this PR, and everything seems fine. Functionality is not changed, only the coding style is changed to follow the Bioc recommendations.

TuomasBorman:
phyloseq is now installed from GH, we have to remember to remove that when Bioc is fixed

PR #663, opened 2024-12-09
 Default: update.tree = TRUE, improve tree merging

TuomasBorman:
This implements the renaming of the tree. Still some things to check and perhaps some optimization.


R/mergeSEs.R
```diff
@@ -491,30 +491,10 @@ setMethod("mergeSEs", signature = c(x = "list"),
...
-        # Order the link data frame, take largest trees first
-        links$whichTree <- factor(links$whichTree, levels = max_trees)
-        links <- links[order(links$whichTree), ]
-        # Remove factorization
-        links$whichTree <- unfactor(links$whichTree)
-        # Remove duplicated links
-        links <- links[!duplicated(links$names), ]
-        # Subset trees
-        trees <- trees[unique(links$whichTree)]
-    }
```
TuomasBorman, line 513: I noticed that this might lead to problems if the trees are named equally. This can happen in `unsplitOn()` function. This chunk was not necessarily need, it was just to remove duplicated trees. After removing this, the memory issue is fixed later.

R/mergeSEs.R
```diff
@@ -538,41 +518,41 @@ setMethod("mergeSEs", signature = c(x = "list"),
...
+            tree[["node"]] <- seq_len(nrow(tree))
+            # Reorder the old tree to match new trees parent node order
+            old_tree <- old_tree[ match(tree[["parent"]], old_tree[["node"]]), ]
+            # Reindex parent nodes of new tree
+            parent <- tree[ match(old_tree[["label"]], tree[["label"]]), ]
+            parent <- parent[["node"]]
+            tree[["parent"]] <- parent
+            # Convert back to phylo object
+            tree <- as.phylo(tree)
+        }
```
TuomasBorman, line 555: Instead of combining all trees at once, the tree is built incrementally. Combining all trees at once might lead to large trees that exceed memory. This current version combine trees one by one, and after each binding it prunes the tree so that there are only unique tips present. This leads to same result but uses less memory.

PR #667, opened 2025-01-07
Add addPrevalence function

TuomasBorman:
This PR adds function that adds the calculated prevalences to rowData.


TuomasBorman:
Add unit test, that the results of add* correspond to get*

PR #669, opened 2025-01-07
Add getAlpha

TuomasBorman:
Currently, we have only addAlpha function that adds values to colData. getAlpha function returns data.frame with the indices without adding them to colData


TuomasBorman:
This PR also simplifies the alpha diversity functions. Previously we combined different diversity functions to be used through addAlpha. We did minimal changes to get the functions working, however, that led to overhead and suboptimal solutions (internal function was calling other internal function that was calling other...). Moreover, ace index calculation was not even working correctly because of these suboptimal choices that were made previously.

antagomir:
Seems good, just one note that the original functions were splitted into informative categories (richness, evenness, diversity, dominance, rarity). It would be good to keep that information (what index belongs to what category) in the documentation at least.

TuomasBorman:
Yes, the documentation has this information

TuomasBorman:
TODO and things to answer:

- Remove code that is not used anymore.
- Investigate why the rarefaction fails (might be also that the old code was incorrect)

R/addAlpha.R
```diff
@@ -574,56 +599,96 @@ setMethod("addAlpha", signature = c(x = "SummarizedExperiment"),
...
+#' @importFrom DelayedMatrixStats rowMeans2
+#' @importFrom BiocParallel bplapply
+.alpha_rarefaction <- function(x, assay.type, index, name, niter, BPPARAM, ...){
+    # Subsample the data and then calculate index based on the rarified data
+    res <- bplapply(seq(niter), function(i){
         x_sub <- rarefyAssay(x, assay.type = assay.type, verbose = FALSE, ...)
-        # Calculating the diversity indices on the subsampled object
-        x_sub <- do.call(FUN, args = list(
-            x_sub, assay.type = assay.type, index = index,
-            name = "rarefaction_temp_result", list(...)))
```
TuomasBorman, line 603: This is causing the discrepancy in results. These old lines included 2 bugs. First of all, the alpha diversity index is calculated based on original table (assay.type), not subsampled ones (by default, it is called "subsampled"). Moreover, the parameters are not passed correctly with "...".

TuomasBorman:
> TODO and things to answer:
> 
>     * Remove code that is not used anymore.
> 
>     * Investigate why the rarefaction fails (might be also that the old code was incorrect)

- Unused code removed.
- The old code was incorrect. Now rarefaction works as expected.

PR #697, opened 2025-02-21
C++ implementation of the Stacked Faith algorithm

### LINKED ISSUE #522: Faith index
We co-authored a fast Faith index calculation, potentially this could be called from R: https://genome.cshlp.org/content/31/11/2131

Ideally, we could utilize the same fast solution by calling it from mia. This might be doable, instead of using slower implementations in external R pkgs.

Leo

jepasan:
This PR implements a fast C++ algorithm (via Rcpp) for calculating Faith's index, based on the C++ code from the sources in #522. For simplicity, I ended up extracting the Faith-related code and condensing it into just a few source files, and doing some amount of rewriting to simplify the code and integrate it more smoothly with R.

The algorithm can be toggled with the parameter `fast.faith` to `addAlpha()` (`TRUE` by default). Based on my own testing, even large datasets have their indices calculated pretty much instantly.

```
> tse <- microbiomeDataSets::GrieneisenTSData()
> system.time(x <- addAlpha(tse, index="faith", fast.faith=TRUE))
   user  system elapsed 
   0.47    0.03    0.50 
> system.time(x2 <- addAlpha(tse, index="faith", fast.faith=FALSE))
   user  system elapsed 
  35.93    0.47   36.92 
> sum(abs(colData(x)$faith - colData(x2)$faith) > 0.00000001)
[1] 7
```

As you can see in the example, the output of the new algorithm may sometimes differ somewhat from the old one. This is because the new algorithm always includes lengths all the way down to the root in its calculations, while the old algorithm seems to drop them sometimes during the process of pruning the tree. In concrete terms, the new algoritm matches the output of `picante::pd()` with `include.root=TRUE`, while the old algorithm matches `include.root=FALSE`. For most datasets I checked this did not lead to significant differences, but for e.g. `SprockettTHData` from `microbiomeDataSets` there is a consistent difference of ~1.17 for basically every sample. I don't know how significant this is going to be in practice, however.

Another thing to note is that the original C++ code is licensed under the 3-clause BSD license - this is compatible with mia's licensing, but I'm not sure what, if anything, needs to be done to properly indicate this in the package.


TuomasBorman (changes requested):
Looks very good!

Details section of `addAlpha` should mention clearly the origin of the implementation. 

You can put the info here:

```
#'  \item 'faith': Faith's phylogenetic alpha diversity index measures how
#' long the taxonomic distance is between taxa that are present in the sample.
#' Larger values represent higher diversity. Using this index requires
#' rowTree. (Faith 1992)
```

antagomir:
Awesome.

BSD3 is the same as BSD2, but additionally prohibiting the names of the authors from being used to endorse or promote the software. If you could mention the source (url) where you took parts of the code from this should be sufficient. The function roxygen documentation @details section would be a suitable place I think.

Imo the differences are acceptable as long as the @details section in the roxygen documentation would mention them clearly. The function works correctly and the differences are justified and transparent.
 
Some more comments in the code PR.

R/estimateDiversity.R
```diff
@@ -115,19 +115,34 @@
...
 }
 
-.calc_faith <- function(mat, tree, only.tips = FALSE, ...){
+# These tags are required to enable the use of Rcpp in the package
+#' @useDynLib mia
+#' @importFrom Rcpp sourceCpp
+NULL
+
+#' @importFrom ape reorder.phylo
+.calc_faith <- function(mat, tree, only.tips = FALSE, fast.faith = TRUE, ...){
```
antagomir: I am not sure if we want to keep the old Faith index at all - could we just remove it?

antagomir: @TuomasBorman

src/faith_R.cpp
```diff
@@ -0,0 +1,70 @@
...
+/* Access the C++ implementation of the fast Faith's PD algorithm from R
+ *  
+ * treeSE <const Rcpp::S4 &> an R TreeSummarizedExperiment object.
+ * faith <Rcpp::NumericVector> the resulting vector of computed Faith PD values
+ *
+ * This functions makes several assumptions about treeSE:
+ * 
+ *  - It must contain a non-empty counts assay and a non-empty RowTree
+ *  - The RowTree must be sorted in cladewise order
+ *  - The RowTree must be rooted - Unrooted trees can be passed without error, but don't produce correct results
```
antagomir: RowTree -> rowTree

src/faith_R.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "tree.h"
+#include "propmap.h"
+
+
+/* Access the C++ implementation of the fast Faith's PD algorithm from R
+ *  
+ * treeSE <const Rcpp::S4 &> an R TreeSummarizedExperiment object.
+ * faith <Rcpp::NumericVector> the resulting vector of computed Faith PD values
+ *
+ * This functions makes several assumptions about treeSE:
```
antagomir: treeSE -> TreeSE usually

src/faith_R.cpp
antagomir: shall we throw a warning with an unrooted tree?

antagomir (changes requested):

antagomir:
I suggest to 

- remove the old and slow Faith implementation entirely (or let's think why we should keep it)
- update the documentation accodingly
- add unit tests that show that this yields comparable results with some external implementation

R/estimateDiversity.R
TuomasBorman: I think we can remove it, there is no need for it anymore, I guess.

Removing the old implementation also simplifies things

antagomir:
Hi @jepasan - could you implement the (hopefully small) additional suggestions, then we could merge?

jepasan:
I've removed the old implementation and updated the documentation. After doing some checking, it seems like the code works fine on both rooted and unrooted trees (it still matches `picante::pd()` in the latter case, just with `include.root=FALSE`). I think the differences were caused by another bug that I fixed, so it shouldn't be a problem anymore.
Regarding unit tests, the existing tests already checks against results from `picante`, which as far as I can tell are still the same, so I'm not sure there's any reason to change them.

R/RcppExports.R
```diff
@@ -0,0 +1,39 @@
...
+#' 
+#' @details
+#' This function makes several assumptions about the contents of
+#' \code{assay} and \code{rowTree}, namely that:
+#' \itemize{
+#'  \item \code{assay} and \code{rowTree} are both non-empty.
+#'  \item \code{assay} has row and column names.
+#'  \item \code{rowTree}'s nodes are arranged in cladewise order.
+#' }
+#' These checks should all be handled in the surrounding R code.
```
antagomir: Does the function throw an informative warning or error if these are not met? This could be mentioned in the documentation.

R/RcppExports.R
```diff
@@ -0,0 +1,39 @@
...
+#' 
+#' The values returned by this function are equivalent to the values returned
+#' by \code{picante::pd()} with the parameter \code{include.root=TRUE}.
+#' 
+#' The C++ code was adapted from an implementation by the Unifrac team (see
+#' \url{https://genome.cshlp.org/content/31/11/2131} or
+#' \url{https://github.com/biocore/unifrac}), which is licensed under the BSD
```
antagomir: I suggest to add instead a bibliographic information in the @references section and then cite here with (Lastname, year) instead. That give better recognition for the original authors.

Or even better, set up references with this
https://cran.r-project.org/web/packages/Rdpack/vignettes/Inserting_bibtex_references.pdf

antagomir: The references already are in @references section so just update the citing style here.

antagomir (approved):
Just some small remaining comments attached. Very good.

R/RcppExports.R
TuomasBorman: 1. This is already checked earlier
2. Rownames are added earlier. It should be checked that the function works even when colnames() are NULL. It might be good to check with all the indices.
3. Cladewise sorting does not affect the tree or results, it just sorts the edges

-->  warning might not be needed as these are just internal functionality that do not affect the results

R/estimateDiversity.R
```diff
@@ -115,73 +115,31 @@
...
-        # Sum up all the lengths of edges
-        temp <- sum(temp$edge.length)
-        return(temp)
-    })
-    faiths_for_taxa_present <- unlist(faiths_for_taxa_present)
-    faiths[ind] <- faiths_for_taxa_present
-    return(faiths)
+    
+    # The tree must be in cladewise order for the algorithm to work correctly
+    temp <- reorder.phylo(tree, "cladewise")
```
TuomasBorman: Store back to tree to save memory

R/RcppExports.R
```diff
@@ -0,0 +1,39 @@
...
+#' object.
+#' @param rowTree An \code{ape::phylo} object containing the rowTree of a
+#' \code{TreeSE} object.
+#' @return A vector containing Faith's PD values.
+#' 
+#' @keywords internal
+#' @export
+.faith_cpp <- function(assay, rowTree) {
+    .Call('_mia_faith_cpp', PACKAGE = 'mia', assay, rowTree)
+}
```
TuomasBorman: This function should not be exported. It should not be visible for user

src/faith_R.cpp
```diff
@@ -0,0 +1,95 @@
...
+        node = tree_sheared.postorderselect(k);
+        
+        // get branch length
+        length = tree_sheared.lengths[node];
+        
+        // get node proportions and set intermediate scores
+        node_proportions = set_proportions(tree_sheared, node, table, propmap, false);
+        
+        for (unsigned int sample = 0; sample < table.n_samples; sample++){
+            // calculate contribution of node to score
```
TuomasBorman: Can you still polish the code formatting:

1. All indentations should be multiplication of 4 spaces
2. The line should not exceed 80 characters
3. Check that ifs, for loops, and while loops use same styling (currently this seems not to be harmonized)

antagomir: I think those issues should come up in the automated checks. Have you also run these? like

~/bin/R-x.y.z/bin/R CMD build ../../ #--resave-data #--no-examples  --no-build-vignettes 
~/bin/R-x.y.z/bin/R CMD check mia_a.b.c.tar.gz #--no-build-vignettes --no-examples
~/bin/R-x.y.z/bin/R CMD BiocCheck mia_a.b.c.tar.gz
~/bin/R-x.y.z/bin/R CMD INSTALL mia_a.b.c.tar.gz

jepasan:
The code should now work whether or not the assay has colnames. I did also try running the automated checks, but as far as I can tell they don't say anything about the C++ code. Still, I've tried to clean it up as best I could.

antagomir:
Seems good to me! @TuomasBorman could confirm?

TuomasBorman (approved):
You are correct that BiocCheck does check only formatiing of R files, but good to keep the code format harmonized anyways

I removed roxygen documentation of the c++ function as we do not want that it is visible for user.

Looks good! Can you put your name to DESCRIPTION file? After that, this is ready to be merged

PR #708, opened 2025-03-13
Support precalculated dissimilarity matrix in dbRDA

### LINKED ISSUE #633: RDA speedup
Running mia::runRDA can be very slow for large data sets. This is problem in particular when we want to calculate alternative RDA models with different formula (e.g. assay ~ BMI + AGE vs.  assay ~ BMI vs. assay ~ AGE etc), as in:

```
mia::runRDA(tse, 
                assay.type = "relabundance",
                formula = assay ~ BL_AGE + MEN,
                distance = "bray",
                na.action = na.exclude)
```

One problem is that the beta diversity is here re-calculated for every combination. 

Speedups could be obtained by using pre-calculated beta diversity matrix, stored in TreeSE object and then supporting the use of that instead, e.g. something like:

```
mia::runRDA(meta(tse)$betadiv, 
                assay.type = "relabundance",
                formula = assay ~ BL_AGE + MEN,
                distance = "bray",
                na.action = na.exclude)
```

Implementation details can be discussed but this would be a substantial improvement.

TuomasBorman:
Related to this issue: https://github.com/microbiome/mia/issues/633

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.15.29
+Version: 1.15.30
 Authors@R:
     c(person(given = "Tuomas", family = "Borman", role = c("aut", "cre"),
              email = "tuomas.v.borman@utu.fi",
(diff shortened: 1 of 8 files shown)


TuomasBorman:
Should `diss.name` be `diss.name` or `dis.name`? --> dissimilarity vs distance

antagomir:
Yes, whynot `dis.name`.. it seems like a good fit

PR #718, opened 2025-03-27
Update mia.R

AliHajj20:
crohn_survival data description


TuomasBorman:
Hi,

this should be added to miaTime package instead.

-Tuomas

TuomasBorman:
To miaTime.R file, to same PR that you have already opened.

PR #739, opened 2025-06-12
Logratio

eotamm:
This pull request adds support for log-ratio transformation of relative abundances in the transformAssay() function. The method computes all pairwise log(x/y) ratios between taxa (rows) for each sample, based on log-transformed relative abundances and an optional pseudocount. This transformation can be useful for downstream modeling and comparative analyses. The method name "log-ratio" was chosen for now, as it explicitly and clearly describes the operation being performed.


antagomir:
Thanks - two more fundamental comments to get started:

1. For some reason the PR shows the entire file contents changed. Usually the PR shows only the specific changes on individual lines. This makes it possible to see and comment the exact individual changes in the code. Without this it is very difficult to assess the new updates. Imo it would be necessary to open a new PR that only shows the specific changes. If not clear how this is done, we can first test with a single minimal change and when that works the rest of the updates can be added in the PR.

2. The issue #739 comments discuss the relation between x/y ("division"?) vs. x-y ("difference"?) transformations. That is a valid point, both might be useful in different situations and LRT could be obtained by chaining log and difference. So my suggestions is to include these instead. Then we can discuss whether a wrapper for combinations (e.g. log+diff) is additionally needed or if it is better to keep them separate.

antagomir:
To concretize the problem with (1), see

This PR changes (entire file changes): https://github.com/microbiome/mia/pull/739/files

Another PR changes (more specific lines change only): https://github.com/microbiome/miaViz/pull/193/files

eotamm:
Yeah I noticed that too now, I will close this PR and open a new one

PR #740, opened 2025-06-12
Adding difference and division

eotamm:
This pull request adds now two new transformations to the transformAssay() function: difference (computes all pairwise differences (x - y) between features (rows) across samples) and division (computes all pairwise ratios (x / y) between features)


R/transformCounts.R
```diff
@@ -98,6 +98,14 @@
 #' log2 = log2(x)}
 #' where \eqn{x} is a single value of data.
 #'
+#' \item 'difference': Pairwise differences between features.
```
antagomir, line 104: Just thinking if we should have shorter name forms (or aliases) like "diff" and "div"? Or perhaps they are too unclear(?). Comment from @TuomasBorman could be welcome.

R/transformCounts.R
```diff
@@ -98,6 +98,14 @@
...
+#' Calculates \eqn{x - y} for all unique feature pairs across samples,
+#' where \eqn{x} and \eqn{y} are relative abundances or transformed values.
```
antagomir: x and y could be anything, relative abundances do not have particular precedence here - so this could be rather something like "where x and y are entries of the specified assay.type"

R/transformCounts.R
```diff
@@ -98,6 +98,14 @@
...
+#'
+#' \item 'division': Pairwise ratios between features.
+#' Calculates \eqn{x / y} for all unique feature pairs across samples,
+#' where \eqn{x} and \eqn{y} are relative abundances or transformed values.
```
antagomir: similar comments as for difference

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
     return(mat)
 }
 
+########################### .apply_transformation_difference ##################
+# Computes all pairwise differences (x - y) between features across samples.
+# Returns a matrix with one row per feature pair.
+.apply_transformation_difference <- function(mat, ...) {
+  # Get the feature (taxa) names
+  taxa <- rownames(mat)
```
antagomir: I am not sure if rownames are generally available here (?). 

At least a check could be included for this, so that an informative error or warning can be thrown.

@TuomasBorman or @Daenarys8 might have pointers to solutions on the same in other functions.

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+  
+  # Generate all unique pairwise combinations of taxa
+  combs <- utils::combn(taxa, 2, simplify = FALSE)
+  
+  # Compute pairwise differences: x - y for each taxa pair across samples
+  diffs <- lapply(combs, function(pair) {
```
antagomir: Just wondering if this can be sped up with vapply/sapply?

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+    t1 <- pair[1]; t2 <- pair[2]
+    name <- paste0("diff_", t1, "_", t2)
+    values <- mat[t1, ] - mat[t2, ]
+    matrix(values, nrow = 1, dimnames = list(name, colnames(mat)))
```
antagomir: It might speed execution if dimnames are added only after the lapply() rather than on every step (not sure?)

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+.apply_transformation_division <- function(mat, ...) {
+  taxa <- rownames(mat)
+  combs <- utils::combn(taxa, 2, simplify = FALSE)
+  
+  ratios <- lapply(combs, function(pair) {
+    t1 <- pair[1]; t2 <- pair[2]
+    name <- paste0("div_", t1, "_over_", t2)
+    values <- mat[t1, ] / mat[t2, ]
+    matrix(values, nrow = 1, dimnames = list(name, colnames(mat)))
+  })
```
antagomir: similar comments as above

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+# Computes all pairwise ratios (x / y) between features across samples.
+# Returns a matrix with one row per feature pair.
+.apply_transformation_division <- function(mat, ...) {
+  taxa <- rownames(mat)
+  combs <- utils::combn(taxa, 2, simplify = FALSE)
+  
+  ratios <- lapply(combs, function(pair) {
+    t1 <- pair[1]; t2 <- pair[2]
+    name <- paste0("div_", t1, "_over_", t2)
+    values <- mat[t1, ] / mat[t2, ]
```
antagomir: What happens if there are zeroes in mat[t2,] - is this tested and is pseudocount supported? (perhaps it has support already in the upstream before we get in this point?). At least an informative warning could be thrown here.

R/transformCounts.R
```diff
@@ -779,6 +793,52 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+    matrix(values, nrow = 1, dimnames = list(name, colnames(mat)))
+  })
+  
+  result <- do.call(rbind, ratios)
+  attr(result, "mia") <- "division"
+  return(result)
+}
+
 # This function is used to add transformed table back to TreeSE. With most of
 # the methods it is simple: it is added to assay. However, with philr, the
```
antagomir: not only with philr but also with difference and division - you could change the text as "However, with transformations that change the dimensionality (e.g. philr, difference, division), the transformed table is added to altExp instead." (or if it is just returned as a matrix rather than added as altExp then say that, I didnt check yet which one this does)

antagomir (commented):
See the suggestions.

Unit tests should be added as well.

Examples section could also show how to do logx-logy by using the combination of log and difference transformations.

R/transformCounts.R
eotamm: Would this be good:
  if (is.null(rownames(mat))) {
            warning("No rownames found in the matrix, generated names will be used.")
            rownames(mat) <- paste0("Feature", seq_len(nrow(mat)))
  }

eotamm:
Added rowname check, zero value warning, and switched to vapply in difference and division functions.

R/transformCounts.R
antagomir: yes but remember the indentations (4 spaces)

eotamm:
I updated transformCounts unit tests and documentation for division and difference.

R/transformCounts.R
TuomasBorman: I would keep full names to avoid unclarity. We could add these aliases later if needed

R/transformCounts.R
```diff
@@ -779,11 +793,96 @@ setMethod("transformAssay", signature = c(x = "SingleCellExperiment"),
...
+    if( is.null(rownames(mat)) ) {
+        warning("No rownames found in the matrix — generated names will be used.")
+        rownames(mat) <- paste0("Feature", seq_len(nrow(mat)))
+    }
+  
+    # Get the feature (taxa) names
+    taxa <- rownames(mat)
+  
+    # Generate all unique pairwise combinations
+    combs <- utils::combn(taxa, 2, simplify = FALSE)
```
TuomasBorman: Add

#' @importFrom stats combn

and use comb() without defining the package

TuomasBorman (changes requested):
I started to wonder if these current methods are meaningful. As far as I know, usually, some reference is used instead of calculating all pairwise differences/ratios (reference for instance mean of all features).

This also causes technical problem as number of pairs increases quickly and the resulting matrix is huge.

R/transformCounts.R
TuomasBorman: More robust solution would be to use indices

R/transformCounts.R
antagomir: Could it be just "-" ... and for division "/" ... then how about supporting the same for sums (+) and products (*).. (no immediate need but would be easy to implement on the same go.. but perhaps just - and / are enough,

R/transformCounts.R
antagomir: It is better indeed if rownames are not critically needed.

antagomir:
There are uses for full log-ratios, for instance this one: https://academic.oup.com/nargab/article/6/2/lqae038/7658053 and we would like to add comparisons to it. And one might have similar needs in settings where number of features is more limited.

TuomasBorman:
> There are uses for full log-ratios, for instance this one: https://academic.oup.com/nargab/article/6/2/lqae038/7658053 and we would like to add comparisons to it. And one might have similar needs in settings where number of features is more limited.

OK. I think we should set limit for number of features, and give error if it exceeds it. Otherwise, users just keep on crashing their sessions.

eotamm:
What do you think would be a reasonable maximum number of features?

antagomir:
This depends on the computing capacity so it is tricky to set a definite hard limit.

How about throwing an informative warning instead?

Also: make sure not to double calculate the ratios (x/y vs. y/x),

eotamm:
I added warning if the matrix has too many rows, supported "-" and "/" aliases for method selection, and switched to using feature indices.

TuomasBorman:
Can you still explore with SparseAssays?

Problem also is usage of `combn()` as it creates a huge data.frame which is loaded into memory. Can it create SparseAssay?

If not, you could first initialize SparseArray and then create a nested loop that goes through the feature pairs. It is not the nicest solution, but should preserve memory which is the problem here. Could you check if that works?

Another option is C++...

eotamm:
Thanks for the suggestions! I will try to check if using a SparseAssay is possible, either directly or with a nested loop.

eotamm:
I replaced combn() with a nested loop implementation. As far as I understand, SparseAssay doesn’t support direct use with combn(). Updated unit tests too.

TuomasBorman:
Can you make comparisons: this new approach vs the old one. Test these methods with different number of samples, let's say 100, 200, 500, 1000, 2000, 5000, 10000. Record the time it takes to run foe each. (looping is also suboptimal solution, but might be safer and more practical than combn)

eotamm:
I compared the combn()-based and nested loop implementations using 100 features and varying the number of samples. Below are the runtimes (in seconds):

100 samples: combn = 0.03 s, loop = 10.63 s
200 samples: combn = 0.05 s, loop = 21.61 s
500 samples: combn = 0.06 s, loop = 66.34 s
1000 samples: combn = 0.14 s, loop = 170.16 s
2000 samples: combn = 0.23 s, loop = 827.42 s
5000 samples: combn = 1.58 s, loop = 1787.61 s
10000 samples: combn = 1.70 s, loop = 3397.47 s 

Would it make sense to add a sparse = TRUE argument or something like that (defaulting to FALSE, i.e., the faster combn-based version), so that users can explicitly choose the sparse loop-based implementation when needed? Or would it be better to always use just one of the methods?

TuomasBorman:
> I compared the combn()-based and nested loop implementations using 100 features and varying the number of samples. Below are the runtimes (in seconds):
> 
> 100 samples: combn = 0.03 s, loop = 10.63 s 200 samples: combn = 0.05 s, loop = 21.61 s 500 samples: combn = 0.06 s, loop = 66.34 s 1000 samples: combn = 0.14 s, loop = 170.16 s 2000 samples: combn = 0.23 s, loop = 827.42 s 5000 samples: combn = 1.58 s, loop = 1787.61 s 10000 samples: combn = 1.70 s, loop = 3397.47 s
> 
> Would it make sense to add a sparse = TRUE argument or something like that (defaulting to FALSE, i.e., the faster combn-based version), so that users can explicitly choose the sparse loop-based implementation when needed? Or would it be better to always use just one of the methods?

The loop seems to scale very badly, however, N features (instead of samples) might be more interesting as that is the bottleneck in the current implementation (too big data.frame).

Anyways, loop seems not to be feasible. Also, the current combn() is problematic. Can you still check options? For instance, DelayedArray, C++, or datatable (https://stackoverflow.com/questions/26828301/faster-version-of-combn)

eotamm:
The data.table-based approach seems slightly faster than combn(), but since it returns a dense matrix, I'm not entirely sure about its memory efficiency in practice. The C++ implementation, which directly constructs a sparse matrix, appears to be similarly fast and is likely the most memory-efficient option of those tested. I translated the sparse matrix logic from R to C++, but I don’t have previous experience with C++, so I’m not fully confident in the implementation details.

antagomir:
If the tests produce correct results then it should be fine?

TuomasBorman:
> The data.table-based approach seems slightly faster than combn(), but since it returns a dense matrix, I'm not entirely sure about its memory efficiency in practice. The C++ implementation, which directly constructs a sparse matrix, appears to be similarly fast and is likely the most memory-efficient option of those tested. I translated the sparse matrix logic from R to C++, but I don’t have previous experience with C++, so I’m not fully confident in the implementation details.

Could you check faith index implementation (along with its PR)? @jepasan  implemented that some time ago, if you propose initial draft I could help you from that

https://github.com/microbiome/mia/pull/697

eotamm:
I just pushed my own changes there as well at the same time. I'll check that everything follows the same style and let you know once it's all good. At least all the tests passed successfully.

eotamm:
I added the corresponding entries to RcppExports, following the same approach as in the Faith index implementation.  I'm not entirely sure whether these generated entries in RcppExports.R and RcppExports.cpp should be committed to Git, since they are automatically created (at least locally), but I included them just in case. It should now work.

antagomir (approved):

src/transformCounts.cpp
```diff
@@ -0,0 +1,164 @@
...
+    
+    // Generate default rownames if not present
+    CharacterVector original_rownames = rownames(mat);
+    bool has_valid_rownames = !Rf_isNull(original_rownames) && 
+        original_rownames.size() == n_features;
+    
+        if( !has_valid_rownames ){
+            Rcpp::warning("No rownames found in the matrix. "
+                          "Generated labels like Feature1 will be used.");
+        }
```
TuomasBorman: For the sake of maintenance, all these checks could be moved to R function instead

src/transformCounts.cpp
```diff
@@ -0,0 +1,164 @@
...
+                                  "Pseudocount will be added.");
+                    i = n_features;
+                    break;
+                }
+            }
+        }
+        for( int i = 0; i < n_features; ++i )
+            for( int j = 0; j < n_samples; ++j )
+                mat(i, j) += pseudocount;
+    }
```
TuomasBorman: You could utilize the pseudocount functionality that already exists. The only thing that is missing is the error that the data includes zeroes and one should apply pseudocount. This could be added to R function instead.

Moreover, you add pseudocount to only zeroes, however, usually it is applied to all values.

TuomasBorman (commented):
Awesome!!!

For the sake of maintenance, can you move functionality that can be done in R to R functions?

C++ functions should only contain the calculation and they should output sparse matrix. R can do the rest.

TuomasBorman:
Also add your name to DESCRIPTION file

eotamm:
I moved the warnings and rowname generation logic to R, only the calculations are now handled in C++.

TuomasBorman (approved):
Great, thanks!!

PR #744, opened 2025-06-27
Initialise agglomerateByModule

RiboRings:
Basic agglomerateByModule function leveraging cross-product was initialised for SummarizedExperiment class.

Consider whether we should:
- [x] Add separate method for TreeSE
- [x] Add support for sample modules in colData
- [x] Integrate more with existing agglomerate and merge functions


RiboRings:
Example usage:

```
devtools::load_all()

data("GlobalPatterns")
tse <- GlobalPatterns

# Add some random modules
N_module <- 3L
modules <- sample(c(TRUE, FALSE),
                  size = nrow(tse) * N_module,
                  prob = c(0.2, 0.8),
                  replace = TRUE)

modules <- modules |> matrix(nrow = nrow(tse))
colnames(modules) <- paste0("module_", seq_len(ncol(modules)))
rowData(tse) <- cbind(rowData(tse), modules)

# Extract module columns
module_columns <- grep("module_", colnames(rowData(tse)), value = TRUE)

# Agglomerate based on modules
tse_module <- agglomerateByModule(tse, by = 1, group = module_columns)
```

antagomir:
Seems good. Some notes:

**Add separate method for TreeSE: ** this is feasible for agglomerateByRanks where the grouping can follow phylogenetic tree. For modules the tree aggomeration might be ill-defined in many cases? Is it meaningful to do that for modules in the general case (or are there relevant special cases to support?). This can be added later if this requires more thinking.

**Add support for sample modules in colData: ** ideally this method could be symmetric and could be applied for rows and cols alike. We also have other such functions and that should be the default.
 
**Integrate more with existing agglomerate and merge functions:** Is there a fundamental difference compared to agglomerateByVariable? If not, we might like to combine these two. It might currently operate only with one column at the time but that is minor difference and just a special case of multi-column thing.

RiboRings:
**Add separate method for TreeSE** I agree with you, so it seems preferable to downgrade TreeSE to SE and remove trees when agglomerating by module.

**Add support for sample modules in colData**: Now it works for both features and samples.

**Integrate more with existing agglomerate and merge functions**: I came to notice a subtle difference between the two. While agglomerateByVariable takes factor variables as groups, agglomerateByModule requires one or more binary (0/1, F/T) variables. Therefore, the two would be analogous only if factor variables were first one-hot encoded and used to agglomerate, but this does not sound like a convenient solution for factors with many levels. So in general, I would keep them as they are (partially integrated) and recommend agglomerateByVariable for single factor groups and agglomerateByModule for one or multiple logical groups.

RiboRings:
This PR is ready for review

R/AllGenerics.R
```diff
@@ -71,6 +71,11 @@ setGeneric("agglomerateByRank", signature = "x", function(x, ...)
 setGeneric("agglomerateByVariable", signature = "x", function(x, ...)
     standardGeneric("agglomerateByVariable"))
 
+#' @rdname agglomerate-methods
+#' @export
+setGeneric("agglomerateByModule", signature = "x", function(x, ...)
+  standardGeneric("agglomerateByModule"))
```
TuomasBorman: Add 4 spaces

DESCRIPTION
```diff
@@ -71,6 +71,7 @@ Imports:
     dplyr,
     IRanges,
     MASS,
+    Matrix,
```
TuomasBorman: Can we avoid this dependency? You could use `DelayedArray` and `crossprod()`

R/agglomerate.R
```diff
@@ -212,6 +221,26 @@
...
 #' merged
+#' 
+#' ## Agglomerate by multiple modules
+#' 
+#' # Generate 30 random modules
+#' N_module <- 30L
+#' modules <- sample(c(TRUE, FALSE),
+#'                   size = nrow(tse) * N_module,
+#'                   prob = c(0.2, 0.8),
+#'                   replace = TRUE)
```
TuomasBorman: Use the following formatting

```
modules <- sample(
    c(TRUE, FALSE),
    ...
)
```

R/agglomerate.R
```diff
@@ -389,6 +418,67 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
...
+#' @importFrom Matrix Matrix
+setMethod("agglomerateByModule", signature = c(x = "SummarizedExperiment"),
+    function(x, by, group, na.rm = FALSE){
+        # Check margin
+        by <- .check_MARGIN(by)
+        # Select side information based on margin
+        FUN <- switch(by, rowData, colData)
+        # Check group
+        if( !all(group %in% names(FUN(x))) ){
+            stop("'group' contained elements that did not match with any",
```
TuomasBorman: Add space after "any"

R/agglomerate.R
```diff
@@ -389,6 +418,67 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
...
+                "column of ", as.character(substitute(rowData)), call. = FALSE)
```
TuomasBorman: should rowData be FUN?

R/agglomerate.R
```diff
@@ -389,6 +418,67 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
...
+        assays <- assays |> SimpleList()
+        # Construct module-wise experiment
+        x <- SummarizedExperiment(
+            assays = assays,
+            colData = if (by == 1L) colData(x) else DataFrame(row.names = group),
+            rowData = if (by == 1L) DataFrame(row.names = group) else rowData(x),
+            metadata = metadata(x)
+        )
+        # Add new names to agglomerated dimension
+        if (by == 1L) {
```
TuomasBorman: Styling of this if differs from others

R/merge.R
```diff
@@ -147,39 +124,72 @@
...
+.agglomerate_assay <- function(
+    assay.type, assay, by, ids, na.rm, average, BPPARAM
+    ){
+    # Check assay
+    .check_assay_for_merge(assay.type, assay)
+    # Transpose if we are merging columns
+    if( by == 2L ){
+        assay <- t(assay)
     }
-    return(x)
```
TuomasBorman: Is there a reason why the functionality from this function is moved? It is better to modify only necessary lines to improve clarity and maintainability while minimizing the risk of introducing unintended changes

R/agglomerate.R
```diff
@@ -32,17 +32,24 @@
...
 #'
 #' @return
 #' \code{agglomerateByRank} returns a taxonomically-agglomerated,
 #' optionally-pruned object of the same class as \code{x}.
 #' \code{agglomerateByVariable} returns an object of the same class as \code{x}
 #' with the specified entries merged into one entry in all relevant components.
-#' \code{agglomerateByRank} returns a taxonomically-agglomerated,
-#' optionally-pruned object of the same class as \code{x}.
+#' \code{agglomerateByModule} returns a
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment},
```
TuomasBorman: I think the function should return an object with a same class than input.

R/agglomerate.R
```diff
@@ -389,6 +418,67 @@ setMethod("agglomerateByVariable", signature = c(x = "SummarizedExperiment"),
...
+        if( any(is_na) ){
+            warning("NAs were found in 'groups' variables and were removed",
+                "before agglomerating the experiment.", call. = FALSE)
+            # Zero out NA modules
+            modules[is.na(modules)] <- FALSE
+        }
+        # Merge assays by module
+        assays <- mapply(.agglomerate_module_assay, assayNames(x), assays(x),
+            MoreArgs = list(by = by, modules = modules, na.rm = na.rm),
+            SIMPLIFY = FALSE)
```
TuomasBorman: You could consider moving these to internal functions

```
assays <- .agglomerate_modules(x, modules)
x <- .constuct_tse_from_modules(x, assays)
```

I guess the general recommendation for length of function is 50 lines

TuomasBorman (changes requested):
Looks good!

I guess this is more efficient than implementation in https://github.com/microbiome/mia/tree/agglomerateByModule (i.e., the branch can be removed)?

DESCRIPTION
RiboRings: So I looked a bit into this. Basically Matrix provides an optimised method for crossprod and sparse array for large adjacency matrices. This functionality is absent in DelayedArray, which focuses on delaying (instead of optimising) matrix operations.

In a nutshell, we have two options:
1. downgrade to base matrix and crossprod if we prioritise reducing package dependencies
2. use Matrix crossprod if we prioritise operations with optimised efficiency

R/merge.R
RiboRings: Sorry for the extra change. I was trying to make `.merge_rows_or_columns` more efficient, and give it a similar structure of `.agglomerate_module_assay` (which in my opinion is also clearer).

The code for `.merge_rows_or_columns` was using lapply for each of the 4 implemented stages of assay-wise operation (checking, transposing, merging and transposing back). For a clearer and more efficient syntax, I reduced the 4 lapply calls to a single streamlined (m)apply call, which in addition also checks and removes NAs for each assay (not only the first one).

DESCRIPTION
antagomir: It's a tough choice. Could depend on how much that performance gain would be in (2) in practice.

antagomir:
Still on this point: 

**Integrate more with existing agglomerate and merge functions:** 
- agglomerateByVariable takes factor variables as groups
- agglomerateByModule requires one or more binary (0/1, F/T) variables

Your argument was that factors with multiple levels should be one-hot-encoded to use the same.

But on the other hand, isn't it possible to use agglomerateByVariable for binary variables, and then it will essentially reduce to the same than agglomerateByModule? If so, agglomerateByModule would not have added value because same could be obtained with agglomerateByVariable applied on binary variables/factors/logicals/numerics.

But am I missing sometihing?

DESCRIPTION
RiboRings: For a sparse Matrix of 1000 features by 1000 modules with 80% zeros, the time is about 0.1 s, or 1/10 compared to using base matrix and crossprod. The memory usage is 4 Mb instead of 12 Mb. Efficiency further improves for higher percentages of zeros, whereas it stays the same for base matrix.

Reference: https://alfurka.github.io/2022-08-12-faster-matrix-computation-with-r-using-sparsematrix/

RiboRings:
Several binary modules can be reduced to a single factor column as long as each feature belongs to only one module. Otherwise, the module matrix is necessary, in which case using agglomerateByVariable for each module would return one agglomerated SE with two rows (module = TRUE and module = FALSE) for each and every module, whereas agglomerateByModule would return one agglomerated SE with all modules as rows.

antagomir:
Yes indeed, unless we consider the case where the variable in agglomerateByVariable contains NAs, which will not be agglomerated. Binary module encoded with NA + TRUE could result in a single row per module. This is more unclear way to implement something like that.

Another point is that agglomerateByVariable has been written to collapse rows (or cols) based on a single variable (e.g. factor with levels). It could be extended to cover case with multiple variables but this would more intuitively be a set of collapsed matrices with varying dimensionality (which would be determined based on the number of factor levels per each variable). Matrices with varying dimensionality could go into altExps. This could make sense in principle but it would indeed be a different method than agglomerateByModule.

DESCRIPTION
RiboRings: Storing the modules in a separate place (e.g., metadata) might allow us to get rid of Matrix dependency, since the user would be able to convert modules to sparse array by themselves, and the agglomerateByModule function could be adjusted to support both Matrix and base matrix without importing it. But I would need to test if this can actually work.

TuomasBorman: If cross product is the only issue, you could try this:

```
library(mia)
data("GlobalPatterns")
tse <- GlobalPatterns

N_module <- 30L
modules <- sample(
    c(TRUE, FALSE),
    size = nrow(tse) * N_module,
    prob = c(0.2, 0.8),
    replace = TRUE
)
modules <- modules |> matrix(nrow = nrow(tse))

assay <- assay(tse)

ref1 <- as.matrix(crossprod(modules, assay))
test <- t(modules) %*% assay

all(ref1 == test)
```

RiboRings: Thanks! I think my code resembles your solution.

For some reason, matrix is faster than Sparse Matrix:

```
> time_for_matrix
   user  system elapsed 
  4.194   0.393   4.587 
> time_for_sparse_matrix
   user  system elapsed 
  7.253   1.396   8.744
```

It might be that the object checks take longer for sparse matrix, but I don't really know. Anyway, now it's possible to use both of them (default to basic matrix).

RiboRings:
This PR is ready for final check/merging. Now the modules are retrieved from a metadata slot specified in `groups` argument.

tests/testthat/test-3agglomerate.R
```diff
@@ -153,13 +153,53 @@ test_that("agglomerate", {
...
+    )
+    # Add modules to rowData
+    modules <- modules |>
+        matrix(nrow = nrow(tse))
+    # Add row and column names
+    rownames(modules) <- rownames(tse)
+    colnames(modules) <- paste0("module_", seq_len(ncol(modules)))
+    # Store modules in metadata slot
+    metadata(tse)$modules <- modules
+    # Add pseudocount assay
```
antagomir: Would be great if we had a demo data set that is already in correct format so that it would be possible to have a minimal example that only focuses on the feature being tested?

R/agglomerate.R
```diff
@@ -212,6 +218,30 @@
...
+#'
+#' # Convert modules to matrix
+#' modules <- modules |>
+#'     matrix(nrow = nrow(tse))
+#' # Add feature names as rownames
+#' rownames(modules) <- rownames(tse)
+#' # Add module names as colnames
+#' colnames(modules) <- paste0("module_", seq_len(ncol(modules)))
+#' # Store modules table in metadata slot
+#' metadata(tse)$modules <- modules
```
antagomir: ready made demo data would be super helpful to simplify the example

R/agglomerate.R
```diff
@@ -212,6 +218,30 @@
...
+#' 
+#' # Agglomerate based on modules
+#' tse_module <- agglomerateByModule(tse, by = 1, group = "modules")
```
antagomir: should we also show in the example how one could add this to altExp directly (as an alternative to creating a completely new tse)?

antagomir (commented):
Seems good, just some suggestions.

Checks still fail?

DESCRIPTION
TuomasBorman: Sparse matrix stored non-zero values and zero values separately so there is overhead in it. The normal matrix is simpler format, but takes more space in memory.

So this is not a bug or something

tests/testthat/test-3agglomerate.R
RiboRings: Yes but it might be easier after we've developed some method to fetch modules from a database.

antagomir: We do have butyrate producing taxa for Metaphlan (3 or 4?) and for Greengenes2. We could use that as an example (this is just one module, though) if we have a gut microbiome example data set with one of these methods. But OK to wait then if not.

antagomir:
I let it for you guys to see when this should be merged.

R/merge.R
```diff
@@ -102,34 +100,14 @@
     # can control this behavior; it can specify the preserved rows for every
     # group or index.
     archetype <- .norm_archetype(f, archetype)
-    
     # Get assays
     assays <- assays(x)
-    # We check whether the assays include values that cannot be summed. For
-    # instance, summing negative values do not make sense.
-    if( check.assays ){
```
TuomasBorman, line 110: check.assays do not have effect anymore. There is this internal flag because the assays are already checked in some functions. For instance, agglomerateByPrevalence checks the assays before calling this function. If we do not have this flag, the assays are checked second time.

This should be handled somehow. (open for suggestions)

TuomasBorman (changes requested):
Thanks!

The code seems ready. Only minor things (see comment). I will still check Zulip discussion.

There has been discussion in many channels. Why metadata slot was chosen over row/colData? Adding module information to rowData would be more intuitive in my opinion as there should be module information for every row(?) and allow tracking the rows.

This would also simplify the method as we can expect that the module information is in specific format.

tests/testthat/test-3agglomerate.R
RiboRings: I think we could add an example with a mia dataset and signatures from BugSigDB.

RiboRings:
> The code seems ready. Only minor things (see comment). I will still check Zulip discussion.

Ok. I'll open a separate PR for the function that adds modules from (BugSigDB) signatures. After that I think we can add a more concrete example to the vignettes.

> There has been discussion in many channels. Why metadata slot was chosen over row/colData? Adding module information to rowData would be more intuitive in my opinion as there should be module information for every row(?) and allow tracking the rows.
>
> This would also simplify the method as we can expect that the module information is in specific format.

I agree that rowData could be a more intuitive location. After our Zulip discussion, we concluded that putting modules in a separate slot could be more convenient/efficient as the number of modules can easily get very big.

If things go in the direction I think, users won't really need to manually add modules to either metadata or rowData, because the modules table will be internally derived from the signatures output by a database (like in the case of BugSigDB). That way, we are free to decide where is best to store modules.

R/merge.R
RiboRings: I see! Is it feasible to check them only once here in the .merge_rows_or_cols function, and avoid checking them elsewhere?

RiboRings: I added check.assays arg back for now.

RiboRings:
Ready to be checked and merged.

antagomir (approved):

antagomir:
Seems good for now.

In  [this paper](https://www.sciencedirect.com/science/article/pii/S266652472400079X?via%3Dihub#appsec1) (appendix, p. 12), we have a list of butyrate producing bacteria:

image

This is a realistic use case for the module method.

Two observations:

1. The module is often in practice tied to a particular preprocessing method (e.g. Greengenes2, MetaPhlAn 2 vs 3 vs 4...).
2. Real modules may contain features from different levels of the row hierarchy (in the above case, mostly genus but also one species, because the genus would be too generic)
 
Suggestion:
- Let us think whether a "module" could include alternative versions / mappings for different processing methods. For instance, a module could be a named list, each element corresponding to one processing method. If not provided then the method could assume that there is just one version that directly applies. Or the versions could be in the original module definition table if there is then a handy R importer that can also pick the selected version. Note that also BugSigDB supports this kind of thing (ncbi, metaphlan..)
- consider how to support cases including multiple feature levels, and show that in the examples
- include the above case as a tiny demo data set in the package and use that in the examples? This would be directly useful in subsequent publications using this feature.

TuomasBorman:
In agglomeration-point-of-view, storing these modules to rowData would be a simpler approach. As Leo mentioned, modules include usually taxonomy information --> rowData would be ideal place to store the module

Moreover, if modules are stored to metadata(), they cannot be used for instance in visualization.

Instead of making complex agglomeration function for different module-formats, it could rely that the module information is in rowData. Specific importers could make sure that the module data is wrangled into correct format.

antagomir:
If it is too difficult, we can first provide a simpler method that does not consider multiple (e.g. taxonomic) ranks. 

But this would be relevant immediate extension as the example shows.

RiboRings:
Ok now it's back to row/colData.

RiboRings:
> If it is too difficult, we can first provide a simpler method that does not consider multiple (e.g. taxonomic) ranks.
> 
> But this would be relevant immediate extension as the example shows.

Thanks for rising this point!

I prefer addressing this issue in the next PR because the solution might also depend on the function to convert BugSigDB metaphlan signature to modules table. Because we probably want to consider mixed ranks already when making the modules table.

For modules with mixed ranks, we can check how the bugsigdbr function [getSignatures](https://rdrr.io/github/waldronlab/bugsigdbr/man/getSignatures.html) is doing it (and maybe other useful functionality).

antagomir:
Perfect. Make sure we don't reinvent the feel w.r.t BugSigDB if they already developed similar solutions.

Can you open a new issue on the multirank thing, so we keep track on it?

TuomasBorman (approved):
Looks good, thanks!

Issue #746, opened 2025-07-07
Bioconductor extended failure - at risk of deprecation/removal


lshep:
Hello Package Maintainer,

We reached out via email but have not heard a response or seen any updates. 

We would like to bring to your attention that your package is failing in both release and devel Bioconductor on all platforms. This is very problematic. Please fix the package in the next few weeks to avoid deprecation.  

https://bioconductor.org/checkResults/release/bioc-LATEST/
https://bioconductor.org/checkResults/devel/bioc-LATEST/

If you have further questions or concerns please reach out on the [bioc-devel@r-project.org](mailto:bioc-devel@r-project.org)

We appreciate your quick attention to this matter especially that there are many packages that are in the reverse dependency list that rely on your package. 

Cheers,
On behalf of the Bioconductor Core Team

TuomasBorman:
Thank you for this notice. This will be fixed as soon as possible.

-Tuomas

TuomasBorman:
This is caused by changes in robust CLR calculation in vegan when it was pushed to CRAN. (related issue: https://github.com/microbiome/mia/issues/742)

-->

Update tests. The new version of vegan calculates the rclr correctly.

@eotamm can you fix?

I will fix the release version also.

It seems that your email and notifications have gone to spam folder. I will make sure that this does not happen again.

eotamm:
Hi, I’m on vacation this week, but I can fix it next week.

antagomir:
In this case i suggest that @raivo-otus could have a look?

TuomasBorman:
That's ok. It is quick thing, I will be able to do it.

TuomasBorman:
> In this case i suggest that @raivo-otus could have a look?

Did not see that before replying. I will check if I have today night time. If not, then I would appreciate if @raivo-otus can have a look.

TuomasBorman:
I made the necessary changes and updated both release and devel as my late-night task because this failure thing was little bit embarrassing from me. I will close this issue when the tests run without errors in Bioconductor.

Related to this I created an issue: https://github.com/microbiome/mia/pull/747. From historical reasons, there are unit tests that are not relevant anymore. This failure was an example of a challenge that they can cause. The unit tests should only test the functionality in mia and not other packages.

TuomasBorman:
Everything is now good.

PR #747, opened 2025-07-07
Fix unit tests

### LINKED ISSUE #742: Calculation of clr & rclr has changed
This PR in vegan changed the calculation of CLR and rCLR transformations: https://github.com/vegandevs/vegan/pull/667

Effectively, the default value of `na.rm` is changed from `TRUE` to `FALSE`, and missing values are replaced with `0`.

TODO:
Fix unit tests.

TuomasBorman:


TuomasBorman:
Keep only single rclr test. This functionality comes from vegan package nowadays, so there is no need to test it extensively.

TuomasBorman:
Fixes this: https://github.com/microbiome/mia/issues/746

Issue #748, opened 2025-07-07
Review necessary unit tests, functions, parameters and packages


TuomasBorman:
1. Build and check run takes long time

Some of the unit tests take long time to run. Moreover, some are unnecessary. For instance, most the transformations come from vegan package which means that it is not necessary for us to test the calculation in detail. Unit tests should only check that the calculation with `transformAssay()` equals to `vegan::descostand()`.

Identify unit tests that are not needed anymore. Modify also tests that take to long time to run.

2. Remove deprecated parameters and functions

3. mia dependencies

Check which packages are really needed in mia. For instance, some clustering functionality was already some time ago moved to bluster package. Can we remove some of the dependencies? Can we use "soft-dependency" (see https://github.com/microbiome/mia/blob/95ee8b1c13c7699f405b20569c80d4df6a23a5dd/R/addLDA.R#L71)

Currently, there is little bit too many dependencies which makes the package harder to install.

@raivo-otus can you check?

raivo-otus:
Sure! I'll see what I can accomplish. 

It will take me some time to get comfortable with the codebase, but I take it these are not time sensitive issues?

TuomasBorman:
Yes, thanks! Take your time

raivo-otus:
Started a draft PR to track progress on review of  transformation tests.

Regarding others have identified that `test-5addNMF.R` causes extreme slowdown when running whole test suite, but executes well if run individually. Not sure what might be the cause.

antagomir:
Would it help to empty the memory on the first line, like `rm(ls())`

raivo-otus:
> Would it help to empty the memory on the first line, like `rm(ls())`

Have tried implementing this. Also tried triggering garbage collection, but both attempts show no improvement.
Remains very slow when running the whole suite.

antagomir:
Does the order matter? if you run that file first (do they run in alphabetical order..)?

PR #749, opened 2025-07-10
DRAFT: Test/simplify transformation tests

raivo-otus:
Reviewing mia unit tests. Related issue #748 

Opening a draft PR to receive ongoing feedback to confirm correct direction/intent.


antagomir (approved):
seems good so far

tests/testthat/test-5transformCounts.R
```diff
@@ -188,6 +156,15 @@ test_that("transformAssay", {
...
         assay(tse, "na_values")[4, 5] <- NA
+        
+        # Tests that clr robust gives values that are approximately same if only
+        # one value per sample is changed to zero
+        expect_equal(assays(mia::transformAssay(tse, assay.type = "test",
+                                                method = "rclr")),
+                     assays(mia::transformAssay(tse, assay.type = "test2",
+                                                method = "rclr")),
+                     check.attributes = FALSE)
+
```
TuomasBorman: This can also be removed as this should be checked in vegan package

TuomasBorman: Basically, what these tests should test is

1) mia::transformAssay() equals to vegan::decostand(). Only couple tests are needed to get certainty.
2) Check that transformations implemented in mia work correctly. These require more extensive testing.

TuomasBorman (changes requested):
Great, thanks! See comments

TuomasBorman (commented):
This does right things but only for small part of tests.  For those transformations that come from vegan, only one test is needed, i.e., test that checks that `transformAssay` equals to `vegan::decostand` (test that mia do not introduce errors when calling the function)

PR #753, opened 2025-07-17
use release in gha runners

0xMuluh:


TuomasBorman:
Thanks! This should fix the problem: https://github.com/markrobinsonuzh/TreeSummarizedExperiment/issues/88

TuomasBorman:
I though that the problem would resolve by itself, but it has not. But the version bump should fix it. If not, then let's come back to this. 

The development should be done against devel version so changing the version in GHA is not optimal.

Issue #733, opened 2025-05-16
pairwise test wrapper


0xMuluh:
**Is your feature request related to a problem? Please describe.**
It is common to perform group-wise/feature-wise Wilcoxon tests which we have seen in many experiments, un/paired

**Describe the solution you'd like**
A wrapper function can take as parameters the tse object, the variable, the index: shannon, b_f_ratio etc and the comparison: treatment, time etc. This should work for within and across group comparisons. The function should also support paired test. 

```
getWilcoxon <- function(tse, col.var, index.var, comp.var, paired = FALSE) {
  df <- as.data.frame(colData(tse))
  
  df %>%
    group_by(.data[[col.var]]) %>%
    wilcox_test(
      formula = as.formula(paste(index.var, "~", com.var)),
      paired = paired
    ) %>%
    adjust_pvalue()
}
```
An extended wrapper could support tests for features as this is also commonly used in genus/species abundance analysis. hence the function could have a mode for features or sample variables. i.e

```
getWilcoxon <- function(tse,
       mode = c("variable", "feature"),
       assay.type = counts,
       col.var = NULL,
       index.var = NULL,
       comp.var,
       paired = FALSE,
       adjust_method = "BH") {

  mode <- match.arg(mode)

  if (mode == "variable") {
    if (is.null(col.var) || is.null(index.var)) {
      stop("For mode = 'variable', both 'col.var' and 'index.var' must be provided.")
    }

    df <- as.data.frame(colData(tse))

    result <- df %>%
      group_by(.data[[col.var]]) %>%
      wilcox_test(
        formula = as.formula(paste(index.var, "~", comp.var)),
        paired = paired
      ) %>%
      adjust_pvalue(method = adjust_method)

  } else if (mode == "feature") {
    df <- meltSE(tse, add.col = TRUE)

    result <- df %>%
      group_by(FeatureID) %>%
      wilcox_test(
        formula = as.formula(paste(assay.type,"~", comp.var)),
        paired = paired
      ) %>%
      adjust_pvalue(method = adjust_method)
  }

  return(result)
}

```

**Additional context**
We can also test if this applicable to cases with repeated measurements across different times, groups and samples.

@antagomir 
@TuomasBorman 
@artur-sannikov

antagomir:
I like the idea of supporting both colData variables as well as features.

This could be done more simply by just providing a numeric vector to test. However, if paired tests or other (e.g. mixed effects models) would be supported then supporting the full TreeSE can be justified.

[...]

PR #757, opened 2025-08-01
statistical tests - pairwise - omnibus

0xMuluh:
#733


antagomir (commented):
Is the name `getPairwiseTest`  correct with Kruskal test? Kruskal is not a pairwise test.

In examples:
```
tse <- tse[1:100, ]
tse <- agglomerateByRank(tse, rank = "Genus") 
```

-> could we just agglomerate to higher rank (e.g. Phylum) for a quick example? Then no need to take subset of 100.

man/getPairwiseTest.Rd
```diff
@@ -0,0 +1,155 @@
...
+    method = "kruskal"
+)
+
+# Add results to metadata
+tse <- addPairwiseTest(
+    tse,
+    assay.type = "relabundance", 
+    group = "SampleType",
+    name = "kruskal_test"
+)
```
antagomir: Would be good to also show how those results can be then investigated / tabulated / visualized further?

tests/testthat/test-getPairwiseTest.R
```diff
@@ -4,9 +4,9 @@ test_that("PairwiseTest", {
     library(dplyr)
     # Data setup - reused throughout all tests
     data(GlobalPatterns, package="mia")
-    tse <- GlobalPatterns[1:100, ]  # Subset for faster testing
+    tse <- GlobalPatterns
     tse <- transformAssay(tse, method = "relabundance")
-    tse <- agglomerateByRank(tse, "Genus")
+    tse <- agglomerateByRank(tse, "Phylum")
```
antagomir: if this is just for testing purposes then subsetting (even to [1:10, ]...) would be faster and agglomeration could be entirely skipped?

0xMuluh:
> Is the name `getPairwiseTest` correct with Kruskal test? Kruskal is not a pairwise test.

True. However, kruskal is folloed by dunn's test.  For the test to be valid, user has to check that kruskal test is useful.

What other name is suitable getSignificance? getTest?

antagomir:
> > Is the name `getPairwiseTest` correct with Kruskal test? Kruskal is not a pairwise test.
> 
> True. However, kruskal is folloed by dunn's test. For the test to be valid, user has to check that kruskal test is useful.
> 
> What other name is suitable getSignificance? getTest?

We might like to distinguish between these two cases:

1) Pairwise test between two groups (e.g. Wilcoxon test)

2) Multi-group test (e.g. Kruskal-Wallis test), followed by a post-hoc test (e.g. Dunn). Potentially we would even like to split these into two different functions that could be used subsequently.

The (2) could reduce to (1) if there are only 2 groups but it is still conceptually different.

These are very basic methods and one could question whether it makes sense to implement new wrappers for them.

The motivation would be that this helps to standardize the way this is done with TreeSE objects, and make it faster to deploy standard summary tables and figures with minimal extra effort. Then also new improved methods can be proposed using the same logic (e.g. probabilistic variants).

I would start by implementing (1) as `getPairwiseTest`.

Once that works, we could tackle (2). That could go in its own issue? You can try to implement these in parallel.

0xMuluh:
Thanks for raising these points. Yes, the distinction between pairwise and multi-group testing should be made clear.  I agree that `getPairwiseTest` is not the most accurate name when Kruskal-Wallis is involved.

>     1. Pairwise test between two groups (e.g. Wilcoxon test)

I'll proceed with refactoring (1) as `getPairwiseTest` . This case is currently covered for un/paired.  Additionally there is also support for t -test.  It could be dropped since the interests is towards non parametric tests ( or maybe issue warning if data violates assumptions?)

>     2. Multi-group test (e.g. Kruskal-Wallis test), followed by a post-hoc test (e.g. Dunn). Potentially we would even like to split these into two different functions that could be used subsequently.

1. Right, for the omnibus tests, this can be easily separated though I would question here if there is use case where e.g  Kruskal-Wallis test is not followed by e.g Dunn's test? Hence the need why it is automated.

> The (2) could reduce to (1) if there are only 2 groups but it is still conceptually different.

True, a simple check can deduce if the number of groups and assign the potential test, but i wonder if we should decide this for user as it would be too opinionated.

Currently the supported methods(with effect sizes, means and log2FCs) include;
- Wilcoxon → un/paired
- t-test  → un/paired
- Kruskal → followed by Dunn
- Friedman → followed by Wilcoxon

The tibble returned by the current fxn is the same for all except for the global tests where the `attrib` contains omnibus tests.

The current fxn is biased as it runs all these tests without check/warning user if test is suitable(non-normality, small n, number of groups... ) but again I would expect user to know their design/experiment. 

The image shows an example return.

Screenshot from 2025-08-07 09-33-06

2. Should fxn determine test based on user data? or perhaps issue warnings or stop completely ?
3. Should post-hoc be optional?

As concerns the naming, this can be addressed by splitting omnibus/post-hoc logic from non/pairwise comparisons:

- `getOmnibusTest()` → Kruskal/Friedman.

- `getPosthocTest()` → Dunn/Wilcoxon-signed-rank.

- allow `getTest()` to chain them (default: post-hoc=TRUE). 

Enough examples can demonstrate use cases.
If these address the aforementioned concerns, then I can refactor the fxn.

antagomir:
1. Separate the pairwise and omnibus tests, the user can decide which one they want to run and we don't decide for the usre.
2. Let's not check model assumptions as part of this, users are supposed to know/check their data separately. 
3. Use case for omnibus without posthoc is for instance when omnibus does not reveal any significant differences -> no use to run posthoc. And in some cases it might be just interesting to make a quick test if differences exist regardless of where.
4. Consider if we could include results from multiple tests; like with alpha diversity the user can choose to get multiple indices, perhaps with this the user could choose to see p-values from multiple methods (e.g. Wilcoxon test and t-test p-values provided in the same table). If this gets too complicated then one test is enough.
5. Open a separate issue about omnibus tests to keep this more manageable.
6. Perhaps getPairwiseTest is a good name but these are DAA methods, intuitive name could also be something like getPairwiseDA
7. Internally we could have separate functions getWilcoxon, getttest etc.?
8. Hide from output table: "statistic", "magnitude"; put effsize and log2FC next to each other

0xMuluh:
ping #759

0xMuluh:
method is now updated to handle pairwise, omnibus and posthoc
getDA_annot

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+        test_fun <- pairwise_t_test
+        effect_fun <- cohens_d
+    }
+    
+    pw <- df |>
+        as.data.frame() |>
+        # Create grouping to test these groups separately
+        dplyr::group_split(across(all_of(grouping_vars))) |>
+        purrr::map_df(function(df_group) {
+            tryCatch({
```
TuomasBorman, line 518: Avoid try-catch

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+        effect_fun <- wilcox_effsize
+    } else {
+        test_fun <- pairwise_t_test
+        effect_fun <- cohens_d
+    }
+    
+    pw <- df |>
+        as.data.frame() |>
+        # Create grouping to test these groups separately
+        dplyr::group_split(across(all_of(grouping_vars))) |>
```
TuomasBorman, line 516: Use importFrom instead

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+    # comparison_vars: the variable(s) we are testing across
+    if( is.null(comparison_vars) ){
+        comparison_vars <- group
+    }
+    grouping_vars <- grouping_vars[ !grouping_vars %in% comparison_vars ]
+    
+    formula <- as.formula(paste(y, "~", paste(comparison_vars, collapse = "+")))
+    
+    res <-  global <- effect <- NULL
+    method_funs <- list(
```
TuomasBorman, line 463: Use same code-styling as elsewhere

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+setMethod("addDA", signature(x = "SummarizedExperiment"),
+    function(x, da.type = c("pairwise", "omnibus", "posthoc"),
+           name = "DAtest", ...) {
+        da.type <- match.arg(da.type)
+        res <- getDA(x, da.type, ...)
+        x <- .add_values_to_metadata(x, name, res)
+        return(x)
+    }
+)
+
```
TuomasBorman, line 202: I think we can remove these wrappers. They do not add much value

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+#' @rdname getDA
+#' @export
+setMethod("getPairwiseDA", signature(x = "SummarizedExperiment"), 
+    function(x, assay.type = NULL, features = NULL, row.var = NULL, 
+            col.var = NULL, group, facet.by = NULL, comp.by = NULL,
```
TuomasBorman, line 207: What is facet.by?

R/getDA.R
```diff
@@ -0,0 +1,812 @@
...
+        
+        if (length(unique(sample_counts$n_samples)) > 1) {
+            stop(
+                "When 'pair.by' is specified, all subjects must have the ",
+                "same number of samples.\nRemove 'pair.by' or filter your ",
+                "data to include only subjects with balanced repeats.",
+                call. = FALSE
+            )
+        }
+    }
```
TuomasBorman, line 344: Is this required for the function to work? I guess user should already be aware of this, and we should not restrict user to run the method if some samples are missing from certain time points

TuomasBorman (changes requested):
I think the function structure can still be simplified and made more generic. I come back to this later.

This functionality is related to little bit bigger project. Let's discuss this in the office.  This PR is good foundation for it.

TuomasBorman:
These will be implemented in https://github.com/microbiome/daa/issues/8

PR #758, opened 2025-08-04
:memo: docs: add vegan 2.7-1 requirement for rCLR and rAitchison

### LINKED ISSUE #737: documentation on rCLR / rAitchison
The documentation could mention that vegan 2.7-1 (2026-06-05) is required for rCLR and rAitchison methods.

mm-hak:
This PR adds a note to the README specifying that the `rCLR` and `rAitchison` methods require the `vegan` R package version 2.7-1 (2026-06-05) or later. Fixes #737.


TuomasBorman:
Thanks, check the date.

Should it be 2025?

TuomasBorman:
Also, started to wonder if OMA is right place for this information as this is not problem in mia but rather in vegan (and OMA has more generic info).

antagomir:
I would say that this is a special case because we made the PR to vegan because there was a need for these in microbiome analyses. And the previous version had shortcomings. This would be relevant information for many prospective users of this functionality.

TuomasBorman (approved):
Thanks!

Issue #767, opened 2025-08-25
Ambiguous error message in internal agglomeration


TuomasBorman:
Certain functions can give bad error message when column is not found from rowData. This is affecting more than the following function as this comes from internal agglomeration function (check all). User cannot do anything based on this information --> improve the message (catch the cause in internal agglomeration function)

```
library(mia)
data("GlobalPatterns")
summarizeDominance(GlobalPatterns, rank = "test")
```

> Error: 'f' must have the same number of rows as `'x'`

TuomasBorman:
This also gives the same error message:

`agglomerateByVariable(tse, by = 1, group = "test")`

PR #770, opened 2025-08-29
update agglomerate messages

0xMuluh:
ping #767 #771


R/agglomerate.R
```diff
@@ -431,6 +435,8 @@ setMethod("agglomerateByModule", signature = c(x = "SummarizedExperiment"),
     function(x, by, group, na.rm = FALSE){
         # Check margin
         by <- .check_MARGIN(by)
+        # Check group
+        .check_group_present(group, x)
```
TuomasBorman, line 439: Is this already checked in line 437 (old code)

R/utils.R
```diff
@@ -195,6 +195,35 @@
...
+             "rows or columns of x.",
+             call. = FALSE)
+    }
+    if (!.is_a_string(group) && !(is.vector(group) || is.factor(group))) {
+        stop("'", name,
+             "' must be either a column name from rowData(x)/colData(x) or a ",
+             "vector/factor of appropriate length.",
+             call. = FALSE)
+    }
+}
```
TuomasBorman, line 225: I think this is too complex as everything works. The only problem is minor message

TuomasBorman (changes requested):
I think this is little bit too complex. Something like this could solve the issue with minimum changes:

To this line:

https://github.com/microbiome/mia/blob/85f5c2fc6b70fd91b63a7b420dc38c4afec0e5b1/R/merge.R#L89

```
# User has defined a a column. Give informative error message if it cannot
# be found. agglomerateByRank already catches this error but other
# agglomeration function do not, so the argument name is group.
if( .is_a_string(f) && !f %in% colnames(rowData_FUN(x)) &&  nrow_FUN(x) != length(f) ){
    stop("'group' must be one of the following options: '", paste0(colnames(rowData_FUN(x)), collapse = "', '"), "'", call. = FALSE)
}
```

TuomasBorman:
https://github.com/microbiome/mia/pull/793 solves the issue.

Issue #771, opened 2025-09-01
add miaTime to DESCRIPTION


0xMuluh:
add miaTime to DESCRIPTION and remove all `require` calls in code

PR #772, opened 2025-09-01
update desc

0xMuluh:
ping #771


TuomasBorman (approved):

TuomasBorman:
@Daenarys8 Can you check why the check in test-5prevalence.R fails? It works in my local computer.

0xMuluh:
@TuomasBorman 
The grouping factor f was incorrectly constructed using `rownames(x)[match(...)]` which indexed back into rownames(x) instead of `rownames(res)`. This caused sequences to be mapped to wrong groups when building consensus reference sequences.

Fix: Changed to `rownames(res)[match(rownames(x), rownames(res))] `to correctly map each taxon to its corresponding prevalence group.

Test update: The test expectation for the Alistipes consensus sequence was updated because:

- The test was previously skipped (miaTime wasn't in Bioc yet)
- The expected value SCRAG... was stale
- The actual consensus from all 16 Alistipes sequences (including 4 unusual ones in SilvermanAGutData) is BCNMK...

TuomasBorman:
Thanks for detailed comments. I made couple of changes, there was logical issue that I found based on your comments.

Issue #775, opened 2025-09-01
Pseudocount message


TuomasBorman:
The following message is printed even though user do not want to apply pseudo count. If `FALSE` (default) this message should not be printed. Even though the message is printed, the function works correctly (pseudo count is not added if it is `FALSE`)

`all(mat > 0, na.rm = TRUE)` --> `pseudocount && all(mat > 0, na.rm = TRUE)`

https://github.com/microbiome/mia/blob/c7dcb2f0dc2b902e2e084c3f6b50b9c6bb2331e4/R/transformCounts.R#L657

PR #776, opened 2025-09-01
pseudocount message

0xMuluh:
ping #775


TuomasBorman (changes requested):
The point of this issue was that the method gave the warning even when user did not want to add pseudocount. This is misleading and causes confusion.

The following should work:

```
if( pseudocount && all(mat > 0, na.rm = TRUE)  ){
     pseudocount <- 0
     message("Message")
}
```
Now the message is shown only if pseudocount is TRUE,  i.e., user wants to add pseudocount.

TuomasBorman (approved):
Thanks!

PR #794, opened 2025-12-08
https://github.com/microbiome/OMA/issues/824

0xMuluh:


R/getMDS.R
```diff
@@ -17,6 +17,13 @@
 #' difference is that these functions apply microbiome-specific options such
 #' as \code{getDissimilarity} function by default.
 #'
+#' By default \code{scater::calculateMDS} keeps only the \code{ntop = 500}
+#' most variable features. When rarefaction is enabled (\code{niter > 0}) the
```
antagomir: does the R code throw a warning of this if it does not use all features? (if data has more than 500)

antagomir (approved):

R/getMDS.R
TuomasBorman: As we have `mia::getMDS` and `mia::addMDS`, we could set the default to _no filtering_. At least, I have not heard that variance filtering is commonly used in MDS in microbiome context (or then authors of papers have not explicitly stated that)

antagomir: Yes, I would also set the default to no filtering.

We can show examples on how to add filtering.

TuomasBorman:
@0xMuluh Can you update the PR based on discussion?

Issue #755, opened 2025-07-24
Missing tree.file and col.data args for most importers


RiboRings:
Only `importMetaPhlAn` has `tree.file` argument? Is there a reason why the others don't have it (`importBIOM`, `importQIIME2`, `importMothur`, `importHUMAnN`)? If not, would it be possible to provide the `tree.file` arg for those as well?

I made a wrapper for miaDash where I needed it. By the way, `importBIOM` doesn't have `col.data` either:

```
.importBIOM <- function(file, col.data = NULL, tree.file = NULL, ...){
  
    tse <- importBIOM(file, ...)
  
    if( !is.null(col.data) ){
        coldata <- read.table(file = col.data, header = TRUE, sep = "\t")
        rownames(coldata) <- colnames(tse)
        colData(tse) <- DataFrame(coldata)
    }

    if( !is.null(tree.file) ){
        rowTree(tse) <- read.tree(tree.file)
    }
  
    return(tse)
}
```

antagomir:
It's a good point. I think the tree import should be added to all methods where possible.

TuomasBorman:
I agree, however, it should be clearly documented. For instance, BIOM file can contain phylogeny and sample metadata (as far as I know. I guess that is why importBIOM does not have  these arguments) so these arguments might be confusing if not sufficiently documented.

RiboRings:
Yes we could return original phylogeny/metadata when the respective args are set to NULL, and replace them otherwise.

TuomasBorman:
Sounds good. Can you prepare PR?

(Check other importers in mia, there might be things to take into account. For instance, if tree do not match, we could give more informative error)

PR #798, opened 2026-01-05
Add tree and metadata options to importers

0xMuluh:
ping #755


R/convertFromBIOM.R
```diff
@@ -68,6 +68,13 @@ NULL
 #'
 #' @param file BIOM file location
 #'
+#' @param col.data a DataFrame-like object that includes sample names in
+#'   rownames, or a single \code{character} value defining the file
+#'   path of the sample metadata file (tsv). (Default: \code{NULL}).
+#'
+#' @param tree.file \code{Character scalar}. Optional path to a phylogenetic
+#'   tree. If provided, replaces any tree stored in the BIOM metadata.
```
antagomir, line 76: Is it possible to provide multiple trees?

R/convertFromBIOM.R
```diff
@@ -111,10 +118,27 @@ NULL
...
+
+    # Load/replace tree if provided
+    if (!is.null(tree.file)) {
+        if (!.is_non_empty_string(tree.file)) {
+            stop("'tree.file' must be a single character value or NULL.",
+                 call. = FALSE)
+        }
+        tree <- ape::read.tree(tree.file)
+        rowTree(tse) <- tree
+    }
```
TuomasBorman, line 156: The node names in tree must match exactly the rownames. Otherwise, the rows cannot be linked with the tree and it leads to warning that some rows were dropped (it is hard for user to know what is the problem):

> Warning message:
> 18371 row(s) couldn't be matched to the tree and are/is removed. 

One could provide links between rows and nodes of tree, but it becomes too complicated.

Proposal:
- Pack this tree adding script into function
- Check if rownames can be found from tree. If not, give error:

> "Rownames do not match with tree labels. Construct TreeSE without tree (tree.file=NULL) and then add the tree manually with changeTree(tse, tree = tree_object, rowNodeLab = link_vector)"

R/convertFromBIOM.R
```diff
@@ -111,10 +118,27 @@ NULL
...
+
+    tse
```
TuomasBorman: return(tse)

R/convertFromBIOM.R
TuomasBorman: In theory uses, but there is no any practical use-cases

TuomasBorman (changes requested):

R/convertFromBIOM.R
```diff
@@ -111,10 +118,27 @@ NULL
...
+        tse <- .add_coldata(tse, col.data)
+    }
+
+    # Load/replace tree if provided
+    if (!is.null(tree.file)) {
+        if (!.is_non_empty_string(tree.file)) {
+            stop("'tree.file' must be a single character value or NULL.",
+                 call. = FALSE)
+        }
+        tree <- ape::read.tree(tree.file)
```
TuomasBorman: Add importFrom and use without specifying the package

Issue #768, opened 2025-08-27
getTop(rank = "genus", ...)


TuomasBorman:
There is no `rank`/`group` argument in `getTop`. Consider adding as this would allow quick check of most abundant features in different levels. It could be supported internally.

The following could be useful:

`getTop(tse, rank = "genus", ...)`

PR #799, opened 2026-01-05
Add rank support to getTop

0xMuluh:
ping #768


R/summaries.R
```diff
@@ -106,32 +111,39 @@ NULL
...
         # check assay
         .check_assay_present(assay.type, x)
         #
         if(method == "prevalence"){
-            taxs <- getPrevalence(
-                assay(x, assay.type), sort = TRUE, include.lowest = TRUE, ...)
+            taxs <- do.call(
+                getPrevalence,
+                c(list(assay(x, assay.type), sort = TRUE, include.lowest = TRUE),
+                  dots))
```
TuomasBorman, line 132: As `dots` are just `...` without any modification, shouldn't the old code do the exact same?

R/summaries.R
```diff
@@ -106,32 +111,39 @@ NULL
...
-        assay.type = assay_name, assay_name = "counts", 
+        assay.type = assay_name, assay_name = "counts", rank = NULL,
         na.rm = TRUE, ...){
+        dots <- list(...)
         # input check
         method <- match.arg(method, c("mean","sum","median","prevalence"))
+        if (!is.null(rank)) {
+            .check_taxonomic_rank(rank, x)
+            x <- agglomerateByRank(x, rank = rank)
+        }
```
TuomasBorman, line 122: You could use `.merge_features` and keep the functionality "optional" so that the `rank` argument is not visible in function definition (too many arguments make the usage more complex)

TuomasBorman (changes requested):

TuomasBorman:
Any updates on this @0xMuluh ?

TuomasBorman (approved):

Issue #800, opened 2026-01-13
Binning transformation


TuomasBorman:
Add "binning" transformation. The transformation is used for instance [here](https://www.biorxiv.org/content/10.64898/2026.01.05.697599v1.abstract)

- The binning should be able to done row or column-wise (easy as we use MARGIN)
- All zero values are assigned to bin 0.
- All non-zero values (N number of them) are assigned to B number of bins evenly based on abundance.
    - Sort taxa into descending order
    - Assign N/B most abundant species into bin B
    - Assign the next N/B most abundant species into B-1, and continue the process

@raivo-otus

raivo-otus:
Seems straightforward enough to implement. I'll add it to my tasks list!

TuomasBorman:
Add information on binning transformation to OMA

TuomasBorman:
@raivo-otus could you make sure that this is done?

PR #801, opened 2026-01-15
Feat/binning transformation

raivo-otus:
Adds a quantile based binning transformation to the mia::transformAssay() -function, as discussed in #800 
Default value of bin = 4, reflects roughly division to "rare, low, medium, high" which is easy to understand.
Unit tests include checks for both sample- and feature-wise transforms. 

Should the transformation default to using "relabundance" assay, or leave choice to user discretion?

Pending tasks:
- [x] Add documentation
- [ ] Add information on binning transformation to OMA

Potential optimizations;
- Adding a parallel version, with e.g. doFuture, can be beneficial for large datasets. 
- C++ implementation called with Rcpp would be even better, both for serial and parallel performance.


R/transformCounts.R
```diff
@@ -36,6 +36,8 @@
 #'   to fill reference sample's column in returned assay when calculating alr.
 #'   (Default: \code{NA})
 #'   \item \code{ref_vals} Deprecated. Use \code{reference} instead.
+#'   \item \code{bins}: \code{Numeric scalar}. For \code{"binning"}, specifies
```
TuomasBorman: Maybe nbins. We have already ncomponents, and ntop

R/transformCounts.R
```diff
@@ -907,6 +910,59 @@ NULL
...
+            # Calculate bin assignments
+            if( n_nonzero < bins ){
+                # For samples with fewer than B non-zero abundance species,
+                # species are distributed proportionally across bins 1 through B
+                bin_values <- round(seq(from = bins, to = 1, length.out = n_nonzero))
+            } else {
+                # Ranks 1 to n_nonzero
+                # Formula: bin = B - ceiling(rank * B / N) + 1
+                ranks <- seq_len(n_nonzero)
+                bin_values <- bins - ceiling(ranks * bins / n_nonzero) + 1
```
TuomasBorman: Can this approach be used also when `n_nonzero < bins`? It would make things simpler for user.

TuomasBorman (commented):
Looks good, couple comments

TuomasBorman:
> Should the transformation default to using "relabundance" assay, or leave choice to user discretion?

As the binning is based on ranks, shouldn't counts and relabundance lead to same result?

TuomasBorman:
Also discussed over lunch; you could update OMA's ML chapter if binning improves the accuracy

R/transformCounts.R
raivo-otus: In the [paper](https://www.biorxiv.org/content/10.64898/2026.01.05.697599v1.full)  linked in #800 they mention in the methods section;

> For samples with fewer than B non-zero abundance species, species were distributed proportionally across bins 1 through B to ensure that low-diversity samples still spanned the full binned range. This avoided overconcentration of values in the highest bins, which could otherwise lead to underutilization of the lower bins during training and limit the model’s ability to generalize across samples with varying species diversity. 

Hence the different behavior in the case of `n_nonzero < bins`

raivo-otus:
> > Should the transformation default to using "relabundance" assay, or leave choice to user discretion?
> 
> As the binning is based on ranks, shouldn't counts and relabundance lead to same result?

My concern is with using e.g. CLR transformed values for the binning, which causes unexpected binning. In that sense it feels like a 'safetynet' to add some sort of check to use relabundance or counts.

TuomasBorman:
Ahh, yes. Maybe you could check that the values are positive and give error if not as the result does not make any sense.

R/transformCounts.R
raivo-otus: Resolved, as no objections.

antagomir:
The standard binning in R is done with function "cut". This is widely used and has many useful arguments. Just thinking whether that should be supported, or in general should multiple binning options be supported and their difference explained in the documentation.

But that can be another PR.

R/transformCounts.R
```diff
@@ -126,6 +128,12 @@
 #' are taken into account. This method keeps only values greater than
 #' \code{threshold} and replaces all other values with \code{value}.
 #'
+#' \item 'binning': Binning of the abundance values into a specified number of
+#' bins. The non-zero values are sorted and divided into \code{nbins} groups of
+#' equal size (quantiles). The group with the highest abundances is assigned to
+#' bin \code{nbins}, while the group with the lowest non-zero abundances is
+#' assigned to bin 1. Zero values are assigned to bin 0.
+#'
```
antagomir, line 137: Add literature ref (if not in @details )

raivo-otus:
> The standard binning in R is done with function "cut". This is widely used and has many useful arguments. Just thinking whether that should be supported, or in general should multiple binning options be supported and their difference explained in the documentation.
> 
> But that can be another PR.

I'll look into using `cut`. It should be possible to implement quantile based binning with cut aswell. It most likely is faster to use built-in functions where possible. 

I think testing different binning methods would be beneficial. The quantile binning approach is supported by the BiomeGPT paper, but there are of course other ways like simple equal width bins etc. Seems logical to include in a separate PR to expand functionality if it seems useful to include other binning options.

TuomasBorman:
@raivo-otus any updates?

raivo-otus:
Added lit ref to the BiomeGPT paper describing the binning strategy, and changed the implementation to utilize `cut()`.

TuomasBorman (approved):
Awesome, thank you!

PR #802, opened 2026-01-26
Multimedia preprocessed demo data

YihanLiu4023:
This is the preprocessed demo data for multimedia analysis


antagomir:
GHA check seems to fail with warning

Error: Error: R CMD check found WARNINGs
  anchors:
Execution halted
    addCluster.Rd: BlusterParam-class
  Please provide package anchors for all Rd \link{} targets not in the
  package itself and the base packages.


I think you could fix it by adding something like this in the relevant roxygen files on relevant functions

` @param ... some parameter of class \link[bluster]{BlusterParam-class}
`

YihanLiu4023:
Hi team, I just created a new PR for updating the mia/man/addCluster.Rd file as I mentioned in email last time. Can you please try again if that change can fix this error? Thanks a lot!

antagomir:
The files in man/*.Rd are automatically created from the roxygen annotations in the corresponding R files. 

I can only see changes in the Rd file, and not in R file in this commit.

This makes me wonder whether the R file has been updated, and Rd file generated from that automatically with `devtools::document()`? If not, can you update the R file and re-generate the Rd file?

The commit was not synced with the devel branch, I just updated this and checks are running.

YihanLiu4023:
Thanks for the clarification! I updated the roxygen source in R/cluster.R and regenerated man/addCluster.Rd with devtools::document() from my local desktop and pulled it back.

antagomir:
Can you confirm when the PR is final (as I am not sure if something still on the way).

RiboRings:
Related to https://github.com/microbiome/OMA/pull/768#discussion_r3616663252.

I do not think this PR is needed, since the ibdhdb dataset is already part of the mia datasets.

PR #789, opened 2025-11-07
Add Joint-RPCA to mia: methods, tests, examples, vignette

aituar17:
### Summary
This PR adds a complete Joint-RPCA (OptSpace) module for multi-omic integration in `mia`.
It introduces the full backend implementation, preprocessing utilities, and example workflows for replicating iHMP/IBDMDB analyses.

The update includes:
- Core functions (`jointRPCA`, `jointRPCAuniversal`, and supporting OptSpace solvers) for single- and multi-omic RPCA ordination.
- Comprehensive preprocessing, projection, and masking utilities ensuring stable compositional handling.
- Unit tests validating structure, reproducibility, and projection consistency on synthetic data.
- Quarto example notebooks demonstrating two- and three-omic IBDMDB analyses, benchmarking, and replication of published results.
- Example IBDMDB data subsets for reproducible testing (with instructions for optional MTX data download).

Together, these additions enable full Joint-RPCA analysis and validation within R, reproducing the same behavior as the original Python implementation on IBD multi-omic datasets.

#### Core Joint-RPCA Implementation
- `R/jointRPCA.R` — Implements the core Joint Robust Principal Component Analysis (Joint-RPCA) using the OptSpace algorithm across one or more compositional tables. Handles preprocessing, filtering, shared sample alignment, and factorization for multi-omic inputs.
- `R/jointRPCAuniversal.R` — Provides a universal user-facing wrapper for `jointRPCA()`, automatically detecting input type (`matrix`, `list`, `SummarizedExperiment`, or `MultiAssayExperiment`) and dispatching accordingly.
- `R/jointRPCAutils.R` — Utility functions supporting the main pipeline, including data validation, helper wrappers, and structure checks used throughout the Joint-RPCA modules.
#### OptSpace Optimization Backend
- `R/jointOptspaceHelper.R` — Internal engine that orchestrates the Joint OptSpace factorization: splitting data into train/test subsets, calling the solver, constructing ordination results, and projecting held-out samples.
- `R/jointOptspaceSolve.R` — Low-level OptSpace solver implementation for joint factorization across multiple paired matrices, producing shared sample embeddings and per-view feature loadings.
- `R/optspaceHelper.R` — Helper for single-view (non-joint) OptSpace-based RPCA. Used for initialization and simple validation steps when only one table is provided.
#### Preprocessing and Data Handling
[...]

PR #804, opened 2026-02-20
Add Joint-RPCA to mia: methods, tests

antagomir:
We should proceed towards merging joint-rpca to devel.

Ideally we should get this included in the next Bioc release, there is still time.


antagomir:
This PR originates from #789 

Use examples [here](https://github.com/aituar17/Joint_RPCA_in_R) - to be moved in the vignettes when this is in mia.

What are the minimum changes to add before we can merge to devel?

The rest can be reported in new issues.

antagomir (commented):
Seems good to me as is.

R/getJointRPCA.R
```diff
@@ -0,0 +1,1191 @@
...
+        assay_names_used <- mae$assay_names_used
+    } else if (inherits(x, "SummarizedExperiment")) {
+        tables <- list(assay(x))
+        anm <- assayNames(x)
+        nm <- if (length(anm) && !is.na(anm[1])) anm[1] else "assay1"
+        names(tables) <- nm
+    } else if (is.list(x) && all(vapply(x, is.matrix, logical(1)))) {
+        tables <- x
+        if (is.null(names(tables))) {
+            names(tables) <- paste0("view", seq_along(tables))
```
TuomasBorman, line 208: It still lacks usage of generic functions, for instance.

Have @raivo-otus had time to check this? I could also dedicate time to this this week so that we get this done.

raivo-otus: I have not yet found time to dive into this, sorry.

In the interest of time, you might be able to work quicker on this matter.

TuomasBorman:
I will merge this PR now. There are several things that need to be taken into account as soon as possible. I will create an issue.

PR #810, opened 2026-03-12
Convert MAE to TreeSE and back

TuomasBorman:


TuomasBorman:
Use `SingleCellExperiment::mainExpName()` instead of `name = "main"`

TuomasBorman:
@RiboRings Could you finalize this PR? This is not urgent.

Issue #817, opened 2026-04-09
deprecation warnings from scuttle::sumCountsAcrossFeatures


lwaldron:
**Describe the bug**
When using `mia::splitByRanks` or other functions that trigger agglomeration under the hood, a flood of deprecation warnings is generated because `mia` is internally calling `scuttle::sumCountsAcrossFeatures`.

In the devel version of `scuttle`, `sumCountsAcrossFeatures` has been formally deprecated in favor of `aggregateAcrossFeatures`.

**To Reproduce**
``` r
suppressPackageStartupMessages(library(mia))
data("GlobalPatterns", package = "mia")
res <- splitByRanks(GlobalPatterns)
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Duplicated labels were made unique.
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
#> Duplicated labels were made unique.
#> Warning in .local(x, ...): 'sumCountsAcrossFeatures' is deprecated.
#> Use 'aggregateAcrossFeatures' instead.
#> See help("Deprecated")
```

<sup>Created on 2026-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

**Expected behavior**
`mia` should migrate from `scuttle::sumCountsAcrossFeatures` to `scuttle::aggregateAcrossFeatures` inside `R/merge.R` to prevent the cascading warnings for downstream users.

**Additional context**
Will start being seen by users with the Bioconductor 3.23 release, so best to fix before then.
```r
[...]

PR #820, opened 2026-04-11
Patch: Shift to mia version of scuttle function to aggregate rows by group

RiboRings:
Hi!

This PR addresses issue https://github.com/microbiome/mia/issues/817.

I initially planned to replace scuttle::sumCountsAcrossFeatures with scrapper::aggregateAcrossGenes. However, the latter exclusively takes a list of sets as input and returns a list of row sums for each set. This differs from the current preferred inputs (factor vector) and outputs (directly the aggregated matrix). This would downgrade any eventual DelayedArray (and possibly) sparse matrices to normal matrices, which is not ideal.

Instead, I decided to leverage the [mia in-house function](https://github.com/microbiome/mia/blob/4c0be600df3a8a38537fdf59a3779ba696a8f07c/R/merge.R#L147) for aggregation which initially handled specifically NA cases. I think this solution has some benefits: it simplifies the code (one aggregation function instead of two), it reduces dependencies, and avoids downgrading DelayedArrays to base matrices.

As a side note, assays are currently agglomerated sequentially, but if relevant we could parallelise using `bpmapply` instead of `mapply`.


R/calculateUnifrac.R
```diff
@@ -81,15 +81,14 @@
...
-#' @importFrom scuttle sumCountsAcrossFeatures
 .merge_assay_by_rows <- function(x, node.label, average = FALSE, ...){
     if( !.is_a_bool(average) ){
         stop("'average' must be TRUE or FALSE.", call. = FALSE)
     }
     # Merge assay based on nodeLabs
-    x <- sumCountsAcrossFeatures(
-        x, ids = node.label, subset.row = NULL, subset.col = NULL,
-        average = average)
+    x <- .sum_counts_accross_features(
```
antagomir: accross -> across ?

antagomir (approved):

antagomir:
Seems good. 

Parallelization might not hurt if it can be added. Maybe not critical here but can be useful if easy to add.

Summing is across features. If this can be easily implemented symmetrically - to enable calculation for either rows or features (`MARGIN` parameter, as in some other functions) then I think that should be done. There can be use cases where one wants to agglomerate either way.

R/calculateUnifrac.R
RiboRings: It was spelled like that in the old code, I thought maybe for some reason. Anyway I think we can safely change it.

RiboRings:
Thanks for the comments @antagomir!

It is already possible to control the margin with the argument `by`, which can be defined in the user-end functions `agglomerateByVariable` and `agglomerateByModule`.

I saw BiocParallel is already imported so this can be parallelised easily and without new dependencies.

R/merge.R
```diff
@@ -129,19 +141,17 @@
...
     is_not_na <- !is.na(x)
     type(is_not_na) <- "integer"
     # Aggregate data to certain groups
-    x <- rowsum(x, ids, na.rm = TRUE)
+    x <- rowsum(x, ids, na.rm = na.rm)
     # Calculate average if specified
-    if( average ){
-        x <- x/rowsum(is_not_na, ids)
-    }
+    if( average ) x <- x / rowsum(is_not_na, ids)
```
TuomasBorman: The functionality of this line is not changed, so I would keep it unchanged

R/merge.R
```diff
@@ -105,17 +104,30 @@
...
+    # Retrieve experiment assays
     assays <- assays(x)
+    # Remove rows with NA group from assays
+    to_remove <- which(is.na(f))
+    if( length(to_remove) != 0L ){
+        # Remove NA groups
+        f <- f[-to_remove]
+        # Remove assays rows corresponding to NA groups
+        assays <- lapply(assays, `[`, -to_remove, )
+    }
```
TuomasBorman: Seems good, but why this is added? Did `scuttle::sumCountsAcrossFeatures` handle this differently than `mia:::sum_counts_accross_features_na`?

R/calculateUnifrac.R
TuomasBorman: As the name is anyways changed, good to  fix the typo

TuomasBorman (commented):
Seems good to me

R/merge.R
RiboRings: Yes, the issue was that `agglomerateByVariable` produces some NA rows in the grouping vector `f` which is passed to `rowsum`. However, `rowsum` interprets these NA rows as a group on their own right, which causes an additional row at the end of the assay (sum of rows for NA group). This then causes a mismatch in the number of rows of the assays and that of the subsetted rowData.

So `scuttle::sumCountsAcrossFeatures` had the same output but without the final row for the NA rowsum.

RiboRings: This is the most efficient way I could come up with, but happy to discuss alternatives.

TuomasBorman: Ahh, I see. Thanks for clarification.

Would it make sense to add it to `.sum_counts_across_features` instead? I guess then our function is working the most similar to `scuttle::sumCountsAcrossFeatures`

RiboRings: Yes but then subsetting is repeated for each assay, which is less efficient.

TuomasBorman: Only subsetting of `f` which is a vector operation so it should not have much difference.

Also, this does looping now in two places: here when subsetting and again when agglomerating. By moving the the subsetting to agglomeration function, we would avoid the first loop.

I am pretty sure that none of these approaches is quicker than others. The choice is more about making the code clearer (I tend to think subsetting in agglomeration function is tidier as it aligns with the old function).

TuomasBorman (approved):

RiboRings:
Thanks @TuomasBorman for wrapping up the PR! Sorry I couldn't get to it before.

PR #823, opened 2026-04-22
Initialise exporters to raw, qiime and mothur

RiboRings:
Hey!

Just opening the PR earlier to share the progress. Will use this for the benchmark and refine it later on, but feel free to check it already if you like.


R/exporters.R
```diff
@@ -0,0 +1,296 @@
...
+        col_data, function(col) if( is.factor(col) ) as.character(col) else col
+    )
+    
+    col_types <- apply(
+        col_data, 2L, function(col) switch(
+        type(col), character = "categorical", integer = , double = "numeric")
+    )
+    
+    col_data <- rbind(col_types, col_data)
+    col_data <- cbind(`sample-id` = c("#q2:types", colnames(x)), col_data)
```
TuomasBorman, line 200: this has "variable_name" formatting, but line 88 has `variable.name`. I think rest of the package has variable names with underscore

TuomasBorman:
Seems very nice! Is this ready?

TuomasBorman:
This is somewhat related to this: https://github.com/ArtifactDB/alabaster/issues/1

RiboRings:
Hi! Almost ready, just need some more time.

R/exporters.R
RiboRings: I just thought it might be confusing to have both assay.dir and assay_dir

antagomir: is it possible to harmonize (is there a need to have both?)

RiboRings:
Do you have any suggestions how to test these exporters?

TuomasBorman:
> Do you have any suggestions how to test these exporters?

You could test that:

1. Correct files are created
2. The formatting matches with desired
3. They contain right data

PR #824, opened 2026-04-30
Generalise agglomerateByModule to non-binary numerical modules

RiboRings:
Hi!

Recently ariadne started supporting numeric matrices describing how much each feature coverage each module (module coverage). So I updated `agglomerateByModule` to also accept this kind of numeric non-binary matrices.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mia
 Type: Package
-Version: 1.21.1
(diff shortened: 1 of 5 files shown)


RiboRings:
Would be great to have this merged to facilitate ariadne submission to Bioconductor

TuomasBorman (approved):
Looks good!

PR #825, opened 2026-05-11
Initialise applyByModule

RiboRings:
Hi!

This PR introduces the `applyByModule` function which provides a way to perform module-wise aka group-wise operations and analyses. Check `?applyByModule` for a few example uses.

It's still a draft, but we can discuss here:
- if current implementation looks good
- if mia vs ariadne is the right place to have it
- if name sounds good (applyByGroup and agglomerateByGroup are also nice though)

Cheers,
Giulio


TuomasBorman (commented):
I did some reformatting, but kept the functionality as it was. Tests need to be added, but after that I think this is ready.

Issue #756, opened 2025-07-29
rbiom::unifrac deprecated


RiboRings:
Hi!

I noticed that `rbiom::unifrac` was deprecated. Currently, it's being used for unifrac calculation according to vignettes

TuomasBorman:
Thanks!

With quick look, I guess they started to support only BIOM objects: https://www.rdocumentation.org/packages/rbiom/versions/2.2.1/topics/bdiv_table

We could use `bdiv_table( as_rbiom(assay(tse)), bdiv = "Unifrac")` but that feels little bit overhead. We could also consider switching to some other implementation / implement it by ourself.

We had it implemented in mia but it was in R (not C++): https://github.com/microbiome/mia/commit/ba42574a62539d99f092c93480b31576e6676a44 That could be used as reference.

antagomir:
There are differences between Unifrac implementation in R, see our earlier discussion in issue #336 

Notably, there are two other implementations that give similar results but they are slower.

phyloseq implementation may have some issues as its results somehow differ from the other options.

-> Check which implementation would work best for now. Or does the license allow using the earlier rbiom code, even if it is not maintained in rbiom any more? C implementation would be faster if available (or implementable) but more work.

antagomir:
@jepasan would you have a chance to look at this, what do you think?

jepasan:
The implementation I adapted the Faith's PD code from has code for unifrac as well, and based on a quick look-over I think it should be possible to adapt that for our code, as well. The code is a little more complex than in Faith's case, though, so it might take a while to get it working. In the short term, we could probably copy the old rbiom code, since it seems to be under the MIT license.

antagomir:
Thanks @jepasan !

If we can use the old rbiom code and it is also good in previous benchmarks then what would be the added value of adapting the code from the Faith source? Could we expect substantial speed gains (as it is in C)?

jepasan:
It seems like the rbiom package already uses C for its unifrac implementation. However, I believe rbiom uses the Fast Unifrac algorithm, while the reference implementation uses Strided State UniFrac, which at least on paper should be significantly faster and use less memory.

antagomir:
Ok - so we could expect to do better by implementing the method you proposed?

Is there a link to that paper?

It could be worth doing because beta diversity calculation times are real bottleneck especially on larger data collections.

Should be validated by comparisons with the rbiom version.

jepasan:
[...]

PR #829, opened 2026-05-25
Striped Unifrac C++ implementation

jepasan:
As discussed in #756, this code adds a C++ implementation of the Striped Unifrac algorithm (https://www.nature.com/articles/s41592-018-0187-8, adapted from the supplementary code in the article), extending the code of the previously added C++ Faith index implementation (#522).
Results should be identical to those provided by ecodive for both weighted and unweighted cases, and at least based on my own testing on a normal laptop somewhat faster and less memory intensive. Further improvements could be likely be made by adding support for multithreading and more optimization. It would also be fairly easy to adapt the code for other forms of unifrac such as generalized unifrac if desired.


antagomir:
Excellent. 

Do you have the code available to demonstrate the ecodive comparisons (or could we include it in the unit tests)? This might be helpful for later.

jepasan:
The current tests in test-5Unifrac.R already compare the results against ecodive, so that part should already be covered. The datasets used there aren't big enough for there to be any large differences in runtime, but benchmarking should also be possible by taking those comparisons and timing them.

RiboRings:
Cool! @TuomasBorman If this can be checked and merged, I'll try to include it in the OMA benchmark.

R/addAlpha.R
```diff
@@ -39,7 +39,7 @@
 #'   whether to remove internal nodes when Faith's index is calculated.
 #'   When \code{only.tips=TRUE}, those rows that are not tips of tree are
 #'   removed. (Default: \code{FALSE})
-#'
+#'
```
TuomasBorman: Remove modification

R/calculateUnifrac.R
```diff
@@ -1,8 +1,6 @@
 #' @importFrom ape drop.tip
-#' @importFrom ecodive weighted_unifrac
-#' @importFrom ecodive unweighted_unifrac
 .get_unifrac <- function(
-        x, tree, weighted = FALSE, node.label = nodeLab, nodeLab = NULL, ...){
+        x, tree, new = FALSE, weighted = FALSE, node.label = nodeLab, nodeLab = NULL, ...){
```
TuomasBorman: Where this "new" goes? What is the purpose?

src/tree.h
TuomasBorman, line 1: Remove modifications from this file as they do not modify the functionality

src/tree.cpp
TuomasBorman, line 1: Same here. Try to avoid making changes that are not strictly needed

TuomasBorman (changes requested):
Thanks, looks good! I will check this this week in more detail

R/calculateUnifrac.R
antagomir: @jepasan can you clarify briefly?

antagomir:
@jepasan can you add the updates proposed by TB then we can hopefully merge

R/calculateUnifrac.R
jepasan: Sorry, that was left there by accident, I've removed it now.

antagomir:
The Ubuntu check is still failing - any solutions possible @jepasan ?

jepasan:
@antagomir, as far as I can tell the error is
```
 remote: Permission to microbiome/mia.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/microbiome/mia.git/': The requested URL returned error: 403
Running post deployment cleanup jobs… 🗑️
/usr/bin/git checkout -B github-pages-deploy-action-dp7khhn7u
Reset branch 'github-pages-deploy-action-dp7khhn7u'
/usr/bin/git worktree remove github-pages-deploy-action-temp-deployment-folder --force
Error: The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌
```
which I don't think has anything to do with my code, at least. Some kind of workflow permissions issue, I guess?

antagomir:
Right, thanks -  @0xMuluh  would you have a chance to check how to resolve?

TuomasBorman (approved):
Great!

The GHA fail is not related to the code.

PR #835, opened 2026-07-22
Disable deploying documentation from PR

sabujcb:
work setup update


antagomir:
Hi @sabujcb can you add a short explanation here what this PR does?

TuomasBorman:
The current GHA builds the documentation and then tries to deploy it. This leads to an error because it does not have permission to deploy to `gh-pages`.

Disabling this also disables building the full documentation. This means that the GHA no longer checks whether the full documentation website can be generated. However, `run_rcmdcheck: ${{ true }}` is still enabled, which checks the documentation, examples, and vignettes as part of `R CMD check`.

I think this is an acceptable trade-off, especially as it lets us use the standard `rworkflows` workflow.

TuomasBorman (approved):
Thanks!

Issue #735, opened 2025-05-25
subsetByTop/agglomerateByTop


TuomasBorman:
Currently, we have functions `agglomerateByPrevalence()` and `subsetByPrevalence()`.  There could be a function that allows user to select the top-X most prevalent taxa.

Currently, user must do it like so:

```
top <- getTop()
tse <- tse[top, ]
```
but this could be streamlined to `subsetByTop()` function. There should be also a way to retain taxa that do not belong to this top-X group, and merge them to "other" (useful for instance before running `plotAbundance()`). Currently, this is done like this:

```
top <- getTop()
grouping <- rownames(tse)
grouping[ !grouping %in% top ] <- "other"
tse <- agglomerateByVariable(tse, grouping)
```

this could be streamlined with `agglomerateByTop()`.

These should be rather fast and easy to implement.

antagomir:
This makes me think that there are numerous possible approaches to define the "features of interest", not only "top" (which can have variable definitions as well.

Why not add support for something like:

```
top <- getTop()
tse <- agglomerateByVariable(tse, top)
```

This would remove the need to have many different agglomerateBy functions for different purposes, I see the use case but I am concerned that it might be getting too divergent.

TuomasBorman:
The logic of `agglomerateByVariable()` is different than in your description. The input should define groups for each feature (in which group each feature will be merged).

Your suggestions (if I read it correctly) just subsets the data based on input (and additionally retain the rest of the data by merging them into "other").

Your approach could be implemented but then I would create new function for that (maybe just `subsetCols/rows()`)

```
top <- getTop()
tse <- subsetRows(tse, top)
```
The difference to `tse[ , ]` is that rest of the data is merged into "other" group.

antagomir:
Yes, I didn't write clearly. I should have used another function name. The idea was that that kind of support could be developed, as you summarized for subsetRows.

If might be possible to load such functionality like `tse[top, , other=TRUE]` (consider `tse[top, , drop=FALSE]` which also exists). 

But maybe that would be less clear.

Ideally, subsetting would not need to be row or col specific (could we guess it in most cases from the input, and use margin argument for the rest).

In any case, I would be in favor of something like `subset()` in order to avoid having too many divergent agglomeration functions.

PR #837, opened 2026-07-23
Filter RPCAInput 

sabujcb:


antagomir:
Seems good! Some checks still fail?

antagomir (approved):

antagomir:
@TuomasBorman can we accept this now?

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+  }
+  
+  if (is.null(rownames(mat))) {
+    stop("Feature IDs are required as row names.", call. = FALSE)
+  }
+  
+  if (is.null(colnames(mat))) {
+    stop("Sample IDs are required as column names.", call. = FALSE)
+  }
+
```
TuomasBorman: Many of these tests are not needed for the function to work

TuomasBorman (commented):
Hi, thanks!

1.
This is not only realted to jointRPCA, but can be more generic. Could the function be subsetBy* or filterBy* instead of filterRPCAInput (related to (related to https://github.com/microbiome/mia/issues/735)

2.
Can you check that the style of code and documentation aligns with the rest of the package?

Use generic methods.

3.

Also, not sure if `MultiAssayExperiment` method is needed. That complicates stuff unnecessarily. For instance, now these same thresholds are used for every omics layer.

4.

I also think this is unnecessary complex. The operation is very simple and can be achieved with less code


```
library(mia)

data(GlobalPatterns)

tse <- GlobalPatterns

subsetByAbundance <- function(
        tse,
        assay.type = "counts",
        min.sample.count = 0,
        min.feature.count = 0,
        min.feature.frequency = 0,
        ...
){
    mia:::.check_assay_present(assay.type, tse)
    if( !mia:::.is_an_integer(min.sample.count) ){
        stop(.)
    }
    if( !mia:::.is_an_integer(min.feature.count) ){
        stop(.)
    }
    if( !mia:::.is_a_numeric(min.feature.frequency) ){
        stop(".")
    }
    #
    tse <- .filter_based_on_abundance(
        tse,
        assay.type = assay.type,
        min.sample.count = min.sample.count,
        min.feature.count = min.feature.count,
        min.feature.frequency = min.feature.frequency,
        ...
    )
    return(tse)
}

.filter_based_on_abundance <- function(
        tse,
        assay.type = "counts",
        min.sample.count = 0,
        min.feature.count = 0,
        min.feature.frequency = 0,
        na.rm = FALSE,
        ...
){
    if( !mia:::.is_a_bool(na.rm) ){
        stop(".")
    }
    #
    mat <- assay(tse, assay.type)
    col_sums <- mat |> colSums(na.rm = na.rm)
    row_sums <- mat |> rowSums(na.rm = na.rm)
    row_frequency <- rowSums(mat > 0, na.rm = na.rm) / ncol(tse)
    
    col_index <- col_sums >= min.sample.count
    row_index <- row_sums >= min.feature.count & row_frequency > min.feature.frequency
    
    tse <- tse[row_index, col_index]
    
    return(tse)
}

subsetByAbundance(tse)
```

R/filterRPCAInput.R
antagomir: We can remove those. This increases the risk of misuse a bit but I think it is acceptable?

antagomir:
1. OK but be clear on the manpage which Python/Gemelli methods this corresponds to if the name is different.
2. OK if there is still something to do (good to have a look).
3. I am not sure if it is needed but at least this clearly is a multi-omic method, i.e. the type of case where MAE is usually promoted for. We could/should discuss whether there are general guidelines on when to use or not use MAE with multiomic datasets. I would hope that we can merge this ASAP and switching from MAE to something else might delay this further.
4.  OK to me at least.

Let us try to finalize this before the weekend if possible.

vignettes/filterRPCAInput.Rmd
TuomasBorman, line 1: Because this vignette is showing the usage of single function, I would remove this. The standard way to show the usage is the "examples" section in the function documentation.

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+    x,
+    assay.type = "counts",
+    experiments = NULL,
+    assay.types = NULL,
+    min.sample.count = 0,
+    min.feature.count = 0,
+    min.feature.frequency = 0) {
+  UseMethod("filterRPCAInput")
+}
+
```
TuomasBorman: Good use of generic methods. However, this is now S3 syntax. We use S4 because SummarizedExperiment is S4 object.

S4 would look like this...

setGeneric is added to AllGenerics.R file

```r
setGeneric(
    "filterRPCAInput",
    function(
        x,
        assay.type = "counts",
        experiments = NULL,
        assay.types = NULL,
        min.sample.count = 0,
        min.feature.count = 0,
        min.feature.frequency = 0,
        ...
    ) {
        standardGeneric("filterRPCAInput")
    }
)
```
Method goes to the file where functionality is implemented (in this case filterRPCAInput.R):

```r
setMethod(
    "filterRPCAInput",
    signature(x = "SummarizedExperiment"),
    function(
        x,
        assay.type = "counts",
        experiments = NULL,
        assay.types = NULL,
        min.sample.count = 0,
        min.feature.count = 0,
        min.feature.frequency = 0,
        ...
    ) {
        ...
    }
)
```

This is the pattern we generally follow throughout the package.

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+    min.feature.frequency = 0) {
+  .validate_rpca_filter_args(
+    min.sample.count = min.sample.count,
+    min.feature.count = min.feature.count,
+    min.feature.frequency = min.feature.frequency
+  )
+  
+  if (!assay.type %in% SummarizedExperiment::assayNames(x)) {
+    stop("'", assay.type, "' is not an assay in 'x'.", call. = FALSE)
+  }
```
TuomasBorman: For these commonly used input checks, we have created own functions.

In this case, `.check_assay_present`. By doing that, we harmonize the error messages and avoid bugs.

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+  
+  mat <- SummarizedExperiment::assay(x, assay.type)
```
TuomasBorman: No need to use package definition as SummarizedExperiment is loaded when mia is loaded. assay() is enough.

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+    experiments = NULL,
+    assay.types = NULL,
+    min.sample.count = 0,
+    min.feature.count = 0,
+    min.feature.frequency = 0) {
+  .validate_rpca_filter_args(
+    min.sample.count = min.sample.count,
+    min.feature.count = min.feature.count,
+    min.feature.frequency = min.feature.frequency
+  )
```
TuomasBorman: I think, MAE method makes the usage more complex without adding much value. The filtering is omic-specific operation. When using this MAE method, user cannot specify different thresholds to different omics, for instance.

I think this is very similar to `transformAssay()`; we do not have MAE method as the function is manipulating single omic.

R/filterRPCAInput.R
```diff
@@ -0,0 +1,368 @@
...
+  if (anyDuplicated(rownames(mat))) {
+    stop("Data table contains duplicate feature IDs.", call. = FALSE)
+  }
+  
+  if (anyDuplicated(colnames(mat))) {
+    stop("Data table contains duplicate sample IDs.", call. = FALSE)
+  }
+  
+  original_features <- rownames(mat)
+  original_samples <- colnames(mat)
```
TuomasBorman: SummarizedExperiment do not need to have row or column names. It is also usually safer to use indices instead of names.

TuomasBorman (commented):
Hi @sabujcb ! I made some changes, sorry about that, I heard that this PR was urgent.

I added some comments.

The functionality was already there and working, but I would try to keep the functions as simple as possible. If the same functionality can be achieved with less code, that is often better: less is more. There are some functions in mia that are a bit over-engineered, but for the sake of readability and maintainability, we should try to avoid adding more of those. Simpler code is usually easier to understand, test, and maintain.

About the naming: we could think about more generic name. Now it refers to RPCA, but I see this as a more generic utility function of filtering, thus the name could reflect that. But we can change it later.

 Thanks for this PR!

vignettes/filterRPCAInput.Rmd
sabujcb: Hi,
Sorry for noticing this late. I did not see the notification earlier. I am ok with these changes; less is always good. 
Thanks for reviewing.

TuomasBorman (approved):

PR #797, opened 2026-01-05
Add Rdpack references

0xMuluh:
ping #707

PR #842, opened 2026-08-10
Document Joint-RPCA filtering examples 

sabujcb:


antagomir (approved):
Seems good!

R/getRPCA.R
```diff
@@ -90,35 +96,67 @@
...
+#'     assay.type = "mtx"
+#' )
 #'
-#' # Apply data transformations. With impute=FALSE, missing values are preserved
-#' # and not imputed.
-#' mae[[1]] <- transformAssay(
-#'     mae[[1]], assay.type = "mgx", method = "rclr", impute = FALSE)
-#' mae[[2]] <- transformAssay(
-#'     mae[[2]], assay.type = "mtx", method = "rclr", impute = FALSE)
+#' # Apply the robust centered log-ratio transformation.
```
TuomasBorman: This is missing very important point: explanation of impute=FALSE

R/getRPCA.R
```diff
@@ -130,10 +168,21 @@
...
+#' \emph{mSystems}, 4(1), e00016-19.
+#' \doi{10.1128/mSystems.00016-19}
+#'
+#' Martino, C. and Shenhav, L. et al. (2020).
 #' Context-aware dimensionality reduction deconvolutes gut microbial community
 #' dynamics.
-#' _Nat. Biotechnol._ doi:10.1038/s41587-020-0660-7
+#' \emph{Nature Biotechnology}.
+#' \doi{10.1038/s41587-020-0660-7}
+#'
```
TuomasBorman: Could you add references like this: https://github.com/microbiome/mia/pull/797

R/getRPCA.R
```diff
@@ -16,6 +16,12 @@
 #' low-rank matrix approximation followed by principal component analysis.
 #' Missing values are handled using matrix completion based on the OptSpace
 #' algorithm.
+#'#' Before running RPCA or Joint-RPCA, untransformed assays can be filtered
```
TuomasBorman: Remoe extra "#'"

R/getRPCA.R
```diff
@@ -90,35 +96,67 @@
 #' }
 #'
 #' @examples
-#' data("ibdmdb")
+#' data("ibdmdb", package = "mia")
```
TuomasBorman: No need to specify package (I think, we should harmonize this and I _guess_ in most of the examples, the package is not explicitly defined)

TuomasBorman (changes requested):
Thanks, see few comments

R/getRPCA.R
antagomir: Harmonization is desirable. There are cases (like this one, I believe) where data set with the identical name is available in two packages (in e.g. phyloseq vs. TreeSE formats). If the package is not specified, the code may load the unintended one and cause confusion because it may not be obvious for the user why the subsequent examples fail. That has happened to me at least.

